PR #49, opened 2017-07-16
Normalize whitespace

DanRStevens:
Project was using mostly tabs. Converted a few random lines that had used spaces to tabs. Shader.cpp used almost all spaces, and was converted to tabs.

Removed whitespace from otherwise empty lines.

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -181,7 +181,7 @@ std::string NAS2D::toUppercase(const std::string& str)
  */
 int NAS2D::clamp(int x, int min, int max)
 {
-    return x < min ? min : (x > max ? max : x);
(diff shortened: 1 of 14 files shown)


ldicker83:
The files with lines like that come from back when Trevor was more active in development -- xcode apparently likes/liked to use spaces instead of tabs.

DanRStevens:
Would you prefer converting the other direction? Tabs to spaces?

Could also convert line endings too if you're interested. The project currently uses CRLF line endings, though with the .gitattributes file, Git often bothers me about trying to convert line endings to LF.

ldicker83:
No, definitely not. I much prefer tabs. I don't know why the spaces bother me so much.

PR #55, opened 2017-07-21
Remove glee package from dependency installation

DanRStevens:
Legacy GLee package is no longer needed now that src/Resources/Shader.cpp has been commented out.


ldicker83:
Much needed, thank you!

PR #77, opened 2018-08-14
Add initial CircleCI config

DanRStevens:
This includes a config file for CircleCI, and a custom Dockerfile with
tools installed for use as a primary CircleCI container.

Can be run locally using the circleci command line interface:
```
circleci build
```

Config file (default: `.circleci/config.yml`)  can be validated with:
```
circleci config validate
```


DanRStevens:
Would be nice to setup, test, and time CircleCI on this branch. I suspect it will produce faster Linux builds than the current Travis setup.

Might need special permission to setup necessary hooks to get CircleCI auto building.

PR #79, opened 2018-08-27
Ternary updates

DanRStevens:
Use the result of a ternary expression in an assignment statement (direct effect), rather then discard results of ternary expression and performing assignment as a side effect.


ldicker83:
Looks good, thank you!

PR #81, opened 2018-08-28
Add AppVeyor configuration file

DanRStevens:
This adds an [AppVeyor](https://www.appveyor.com/) config file to enable Continuous Integration builds in a Windows environment.

I'm still testing with this one. In particular, I have not resolved how dependencies are installed. I'm thinking they could be downloaded with [Nuget](https://www.nuget.org/), though this still needs some investigation.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,3 +5,5 @@ platform:
   - x64
 build:
   project: proj/vc14/NAS2D.sln
+init:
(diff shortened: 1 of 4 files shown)


DanRStevens:
I would recommend reviewing the Nuget package selection, and make sure they are suitable. I suspect some of them are not from the original project authors. There's a bit of inconsistency in the naming of packages.

I'd also be curious to know if the Nuget packages negatively affects the build for someone who already has a local install of the necessary dependencies.

Additionally, if packages are installed with Nuget, are those packages properly ignored with the .gitignore file. The .gitignore file is quite large, and mentions Nuget packages already, so that probably is the case.

PR #82, opened 2018-09-24
Cleanup filesystem code

DanRStevens:
I made some simplifications to the filesystem initialization code. You may want to double check the Windows portion. In particular, make sure that multiple trailing directory separators really are safely ignored on Windows.

Hopefully this cleanup will make it easier to address upgrading to the newer simpler methods in the more recent versions of Physfs.

### DIFF
--- a/src/Filesystem.cpp
+++ b/src/Filesystem.cpp
@@ -8,6 +8,10 @@
 // = Acknowledgement of your use of NAS2D is appriciated but is not required.
 // ==================================================================================
 
+#if !(defined(WINDOWS) || defined(__APPLE__) || defined(__linux__))
+#error Filesystem support for this platform has not been developed.
+#endif
+
 #include "NAS2D/Filesystem.h"
 #include "NAS2D/Exception.h"
 
@@ -25,7 +29,6 @@
 
 using namespace NAS2D;
 using namespace NAS2D::Exception;
-
 
 bool FILESYSTEM_INITIALIZED = false;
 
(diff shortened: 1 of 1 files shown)


ldicker83:
Reviewed the commits, they look good to me for the most part though I did note a couple of questions.

I'm reevaluating the init code for the FS and ... it could really use an overhaul especially with PhysicsFS having deprecated a number of functions that `FileSystem` still uses. Not sure if I should slate for a 1.5.0 release or later.

DanRStevens:
I'm afraid the review approvals on GitHub don't say which commit(s) you were viewing. I've encountered that limitation myself. As such, I'm a little uncertain what the comments pertain to.

Sometimes while reviewing, I have used a workflow where I examine each commit on it's own in a separate tab, occasionally making annotated code comments. The annotated code comment ties the comment to a commit, and a particular block of code (I aim for the bottom of the block since the preview includes lines above). One annoyance about the GitHub interface, is the first comment for each commit (in a new tab) has the button labelled "Start a review", rather than "Add review comment". It doesn't matter though, you can keep "starting a review", making comments, closing pages, opening new pages, starting new reviews, adding more comments, and then finally collect them all into a single review at the end. GitHub will remember all the comments you were making. For the final review, I like to open a review window that shows all commits. I find if I don't do that, GitHub doesn't accurately track that you've viewed everything, and can list your review as only pertaining to earlier commits. It affects the changes since you last viewed button. It may also impact where the comments appear in the PR timeline. In the final overall review window, I'll add any summary I need to, and either comment or approve all changes together.

PR #86, opened 2018-09-28
Add Docker support

DanRStevens:
This adds support for building an Ubuntu 16.04 based Docker image, with all needed dependencies, and compiling NAS2D.

I've updated the top level `makefile` to support easily running Docker commands. I've tested them on Linux, though if you have GNU Make and Docker installed, I believe they could also be run from Windows. Would appreciate some feedback and testing there.

Download (and cache) Ubuntu 16.04 base image, download and install all needed system packages, download, compile, and install all needed source packages, produce a tagged Docker image:
```make build-image-ubuntu-16.04```

Compile code using Ubuntu 16.04 base image:
```make compile-on-ubuntu-16.04```

----

Additionally, the `Makefile` in the `docker/` folder can likely be used to do download, compile, and install source packages on arbitrary versions of Linux. It has been generalized greatly from the initial version, and gives tag names to various build stages, allowing for the build and install steps to be easily customized.

----

Note: It took about 25 minutes to build the Docker image on my low end laptop. That didn't include downloading of the base Ubuntu image, which I already had cached, but rather just download and install of specific dependencies. The resulting image is cached, so the build only needs to be run once, unless of course the build environment itself is updated.

Using the image to compile the code takes about 45 seconds on my low end laptop.

----

Currently the Physfs package is installed as a stock system package. A small tweak to the `Dockerfile` would allow an updated version of Physfs to be installed from source. The provided dependency `Makefile` already supports this. If/when code is updated to require a newer version of Physfs, this switch over can be easily made. For now though, I wanted to keep the stock system package for testing purposes.


DanRStevens:
Heh, I opened this branch months ago with that initial commit for the dependency `Makefile`. I had never rebased or merged changes since then, as this was fairly isolated work. Meanwhile we updated and tweaked and CI build configs. Those were never included in this branch, hence the CI build failures.

PR #88, opened 2018-10-09
Normalize copyright symbol

DanRStevens:
Previously some files used an ASCII encoded copyright symbol (hex code A9). As this is beyond code 127 (hex code 7F), this means the byte encoding differs between ASCII and UTF-8. Most modern code editors assume UTF-8 encoding. As such, the symbol was being interpreted and displayed incorrectly in some editors.

Side note, those characters were showing up oddly in Git diffs due to the encoding problem.

### DIFF
--- a/include/NAS2D/Delegate.h
+++ b/include/NAS2D/Delegate.h
@@ -1,6 +1,6 @@
 // ==================================================================================
 // = NAS2D
-// = Copyright © 2008 - 2017 New Age Software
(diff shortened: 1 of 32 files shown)


ldicker83:
That came from when I was using Visual Studio 2008 which really didn't handle UTF encoding of any sort well. Looks good to me, should be a transparent change for everybody.

PR #91, opened 2019-03-26
Update Utility

DanRStevens:
I took a bit of inspiration from [`std::make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) on this one.

The `Utility::instantiateDerived(T* t)` method didn't actually instantiate the derived class. It just took an already instantiated object as a parameter. As such, the name wasn't entirely accurate, though I figure that was due to implementation difficulties, rather than lack of spirit. I saw two difficulties with being able to actually instantiate derived objects. One was knowing what derived class the caller wanted. The other was forwarding of constructor arguments to be able to construct arbitrary non-trivial types.

Both of the above problems were nicely solved by `std::make_unique` method for creating `unique_ptr` instances. I thought I'd borrow the method.

### DIFF
--- a/include/NAS2D/Utility.h
+++ b/include/NAS2D/Utility.h
@@ -122,7 +122,7 @@ public:
 	 *			created object or deallocating the object will yield
 	 *			undefined behavior.
 	 */
-	[[deprecated]]
+	[[deprecated("Consider using Utility::init instead")]]
 	static void instantiateDerived(T* t)
 	{
 		if (mInstance == t)
--- a/include/NAS2D/Utility.h
+++ b/include/NAS2D/Utility.h
@@ -90,10 +90,14 @@ public:
 	 *
 	 * \param	args	A list of arguments to be forwarded to the \c Type objects's constructor
 	 *
+	 * \return Reference to the newly created object as \c Type.
+	 * This allows further method calls to complete initialization or setup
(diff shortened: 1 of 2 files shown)


DanRStevens:
I realized something felt a little incomplete, so I added a few extra finishing touches.

@ldicker83: You may want to check the code comments for line wrap widths, as yours may be set shorter than mine. Would also appreciate some feedback on the method name I chose. I was wondering if maybe I should have named it `create`, or something similar.

include/NAS2D/Utility.h
```diff
@@ -64,6 +64,54 @@ class Utility
...
+	 *
+	 * \note	This method should be called before <tt>Utility::get()</tt>.
+	 */
+	template<typename Type = T, typename... Args>
+	static Type& init(Args&&... args)
+	{
+		// Instantiate a new object with forwarded constructor arguments
+		auto newInstance = new Type(std::forward<Args>(args)...);
+
+		if (mInstance)
```
ldicker83: According to static code analysis, this check is unnecessary.

ldicker83: Static code analysis indicates this check is superfluous -- in the interest of brevity this check should be removed and replaced simply with a call to delete.

PR #95, opened 2019-03-26
Make PHYSFS_mount calls more readable

DanRStevens:
Pass argument values that are more self documenting.


src/Filesystem.cpp
```diff
@@ -28,6 +28,12 @@ using namespace NAS2D::Exception;
 
 bool FILESYSTEM_INITIALIZED = false;
 
+enum MountPosition
+{
+	Prepend = 0,
```
ldicker83: Code styling should be all-caps for enumerated values. Generally something like MOUNT_PREPEND

PR #100, opened 2019-03-27
Rename `addToSearchPath` to `mount`

DanRStevens:
Both the higher level component (`Game`) and the lower level component (PhysFS) use `mount`. It makes sense to be consistent here.

This is a breaking change. Granted a simple one, and easy to fix for library users. I think it's worth it.


ldicker83:
Agreed. More appropriate naming than what I had before.

PR #104, opened 2019-03-29
Delete default copy constructor and copy assignment operator

DanRStevens:
Explicitly deleting the methods, rather than simply declaring them private and leaving them undefined, is perhaps more self documenting. Additionally, if a method within the class (with access to private members) tries to call either of them, it will be reported earlier as a compiler error rather than a linker error.


ldicker83:
This was originally built in 2008/2009 before C++11 was ratified and explicitly deleting implicit operators wasn't yet a thing.

PR #115, opened 2019-04-02
Utility constexpr-if

DanRStevens:
This expands the scope of the `Utility` class so it's able to handle non-default-constructible types too.

For non-default-constructible types (and only for those types), the code throws an exception at runtime if `get` is called before `init`. As the type is non-default-constructible, the method can't simply create an object on demand. It would need to call a non-default constructor and pass actual arguments which the method doesn't have.

Upon `nullptr`, the decision to either construct a new value (default constructible) or throw an exception (non-default-constructible) is made at compile time. There is no run-time check for this. The proper code is compiled in as part of template method selection.

The [`constexpr-if`](https://en.cppreference.com/w/cpp/language/if) feature used here is a C++17 feature. Compiler/project settings may need to be adjusted to enable it.

### DIFF
--- a/include/NAS2D/Utility.h
+++ b/include/NAS2D/Utility.h
@@ -11,6 +11,7 @@
 
 #include <utility>
 #include <type_traits>
+#include <stdexcept>
 
 
 namespace NAS2D {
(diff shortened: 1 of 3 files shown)


DanRStevens:
Looks like the version of GCC used on TravisCI is too old to use `constexpr-if`.
```
include/NAS2D/Utility.h: In static member function ‘static T& NAS2D::Utility<T>::get()’:
include/NAS2D/Utility.h:67:7: error: expected ‘(’ before ‘constexpr’
    if constexpr(std::is_default_constructible<T>::value)
       ^
```

@ldicker83: I'm curious what the state of support for `constexpr-if` is on Windows.

----

If `constexpr-if` is too new of a feature to use, it may be possible to rewrite using `enable_if` guards. There would be two versions of the method then, selected based on the type of the template argument.

PR #120, opened 2019-04-04
Vc17

cugone:
**Summary:**
- Added VC17 project files.
- Removed PhysicsFS dependency in favor of std::filesystem. 
- Fixed all warnings and errors in all configurations and builds.

**Added VC17 project files:**
Moving to VC17 allows for more modern features of the C++ language that were not implemented in VC14/15 and was a *requirement* for using std::filesystem outside of the experimental namespace. In addition it is a stepping stone for also creating a VC19 project which was released April 2, 2019.

**Removed PhyscsFS dependency**
Nas2D::Filesystem now delegates to std::filesystem calls and std::filesystem::path objects.
With the supported and implemented features in VS17's implementation of std::filesystem PhysicsFS is now redundant. In addition it's support for older file types is really not needed in this day-and-age.

**Fixed all warnings and errors in all configurations and builds**
This is so the project compiles under MSCV's Warning Level 4 and Treat Warnings as Errors. A nice side effect is it encourages always having clean builds. If the compiler is warning about something it probably shouldn't be ignored. Also, it just feels good seeing 0 errors, 0 warnings, 0 messages when compiling.


src/Filesystem.cpp
```diff
@@ -417,29 +322,178 @@ std::string Filesystem::workingPath(const std::string& filename) const
...
+        return false;
+    }
+
+    std::ifstream ifs{ p };
+    //Dump ifstream buffer directly into stringstream and convert to string
+    out_buffer = std::string(static_cast<const std::stringstream&>(std::stringstream() << ifs.rdbuf()).str());
+    return true;
+}
+
+//TODO: Needs Linux/MacOSX implementations!
```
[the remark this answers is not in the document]
cugone: The Linux and MacOSX implementations of GetExePath? Yeah, I did not know the OS-specific APIs and intended to update them later.

In plain C++, Generally the working path where the executable is run from is the first argument from the command line (argv[0]), the default std::filesystem::current_path() (not calling it with arguments first) returns the working directory which is the same as argv[0] but both of these can be changed via $PATH env_vars and calls to std::filesystem::current_path(std::filesystem::path new_path).

A quick google search netted an informative StackOverflow answer:
https://stackoverflow.com/a/1024933/421178
and another posted links to the required OS documentation:
https://stackoverflow.com/a/1024937/421178

It looks like the MacOSX implementation is as simple as the Windows one:
 - uint32_t size = 0;
 - std::string path{};
 - Call _NSGetExecutablePath(path.data(), &size) to get the required size
 - Resize the buffer: path.resize(size)
 - Call _NSGetExecutablePath(path.data(), &size) again and cram it into a buffer.
 - Construct a std::filesystem::path from the buffer and convert it to a preferred canonical path.
 - std::filesystem::path p{path};
 - p = p.canonical();
 - p.make_preferred();

Linux...not so much. Create a path to /proc/self/exe which is a symbolic link then convert it to an absolute canonical preferred path...

cugone:
Out of curiosity, what parts struck you as "I really like this"?

Yeah I have to apologize for breaking the build systems. Building my code consists of "perform a Buddy Build. Pull latest from Git or Perforce. Hit F7. Do both compile? Push to master or branch." I've never used build systems let alone *remote* build systems.

Also, I screwed up and accidentally added the changes to Common and Exception in the "fixed warnings and errors" commit.

The changes to Common were adding several string utility functions and the implementations to toUppercase and toLowercase were streamlined. (Why take a string by const-ref only to immediately copy it...just pass it by value and let the compiler do the copying. :P )

Exception was adding a simple exception type for "file not found"

ldicker83:
> The changes to Common were adding several string utility functions and the implementations to toUppercase and toLowercase were streamlined. (Why take a string by const-ref only to immediately copy it...just pass it by value and let the compiler do the copying. :P )

Those functions were originally written in 2008 long before C++ had any notion of 'move semantics'. It's not incorrect either way, good optimizing compilers will deal with it well enough. Anyway, tiny nitpick. I'm not sure when I'll have the chance to review everything, will hopefully be this weekend.

src/Filesystem.cpp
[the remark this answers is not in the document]
cugone: One thing that is definitely a problem is the build script is trying to build the vc14 project instead of the vc17 project. std::filesystem did not exist in vc14. As usual I forgot to update the AppVeyor.yaml script. :/

cugone:
The documentation is vary sparse on Travis (I'm starting to like AppVeyor more and more...) on how to actually set the -std=c++17 flag for gcc and clang so the filesystem errors will go away. Luckily they do document that they are using clang and llvm 7 which its c++17 implementation supports std::filesystem. GCC however is only 5.4.0 which is way behind. Time for more preprocessor defines!

src/Filesystem.cpp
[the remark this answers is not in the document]
cugone: @DanRStevens GetExePath now calls correct OS-dependent functions. I cannot locally test if any of it is logically correct, only syntactically through the TravisCI, as I do not have a local Linux or Mac OSX available.

DanRStevens:
Ok, in terms of what I like:
1. VS project file update
2. Making use of new move semantics (pass strings by-value)
3. Type alias with `using` rather than `typedef`

Those can each be separate branches/PRs.

In terms of what I think have blocking difficulties:
1. Using C++17 features:
 1a. Using `<filesystem>`
 1b. Using constexpr-if

As such, I don't think we're currently in a position to replace PhysFS. The main problem is cross platform support. We may be able to manage that soon, but we'll need updates to the CI builds first.

----

No worries about breaking the CI builds on a branch. That's what branches are there for. Problems get caught before they are merged to master. (At least in theory. We had some CI service problems recently, and managed to merge some problem commits to master during the outage. Cleanup is ongoing).

----

In terms of reviewing, approving, and merging code, I'd like to see that done on smaller more targeted branches. With that said, you are of course free to continue adding commits here for testing purposes and to work out the details. It's perfectly valid to use a PR for feedback and CI support, even for code that is broken and won't be merged in. Changes can always be cleaned up in a new branch/PR and the old one closed.

----

As for CI, my experiences with TravisCI have been a little on the awful side. I'd like to move away from TravisCI. In particular, I've had a pretty good experience with Linux builds on CircleCI. It's much easier to maintain builds there, better customization support, and the builds end up being way way faster.

Currently we only do MacOS builds on TravisCI. I think CircleCI has recently added MacOS support, so perhaps it's worth looking into. I'm not sure if it's available on the free tier yet.

AppVeyor has been a pretty good experience for Windows builds.

----

I would prefer fewer preprocessor defines whenever possible. :wink:

cugone:
@DanRStevens What's really killing me is the upwards of 20 minute wait for TravisCI to re-download and rebuild all the dependent packages before it gets close to touching the code. The AppVeyor windows build takes 30 seconds.

src/Filesystem.cpp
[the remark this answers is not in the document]
DanRStevens: Yes, the CI builds have only been doing straight builds so far. We are only just now starting to add post-build unit tests. We would need unit tests for this to check if the code runs correctly.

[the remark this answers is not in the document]
cugone: @DanRStevens  The linux builds pass. After several rounds of testing and googling, I've determined TravisCI's macOSX gcc, llvm and clang versions support neither the <experimental/filesystem> nor the <filesystem> headers AT ALL.

cugone:
@DanRStevens I have them set to Smart Indent, Tab Size 4, Indent Size 4, Insert Spaces. (side note, Tabs are evil. Spaces FTW) What's the deviation? It may have been VS auto-formatting and Visual Assist messing around too. I did a few bulk Create Implementation refactors and VS and Visual Assist both like to ignore my newline and format settings.

src/Filesystem.cpp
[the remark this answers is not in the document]
DanRStevens: That was my finding about a year ago on another project as well. That other project never ended up officially supporting MacOS because of that. A decision to be revisited when tools have a chance to catch up.

DanRStevens:
I just checked. The project uses hard tabs, not spaces. That would be the problem.

I think I have a slight preference for soft tabs these days, but only slight preference. Back when I used VS regularly, I used to prefer hard tabs. I've noticed most Windows projects I've worked with people on, the other developers seem to prefer hard tabs. For Linux work, the preference of other developers seems to be mostly soft tabs.

At any rate, it doesn't matter to me, as long as it's consistent. Though I prefer not using hard tabs to try and line up comments on the end of a line. That tends to get messy.

DanRStevens:
@cugone: Would you be interested in creating an account on CircleCI?
https://circleci.com/signup/

I noticed the CircleCI builds are not running on the branch in this PR. I suspect it's because CircleCI runs builds on user accounts rather than an organization account. I'm not sure if the runs work at the level of individual commits, or based on the user who opened the PR. I'm guessing if you created an account on CircleCI using GitHub as your authentication provider, the CircleCI builds would start running for this branch.

One benefit would be faster feedback for Linux builds. The CircleCI Linux builds only take about 24 seconds to run.

ldicker83:
Okay, I'm not reading most of the conversation here, it's just a wall of text. Dan is aware of what I'm struggling with in my own head, don't want this to sound like a "poor me, feel sorry for me" but suffice it to say I'm stuck in my head and it's been hard to be motivated to do anything.

That stated, I'm cloning your repository and the VC17 branch -- I noticed that you made a VC17 project directory -- Visual C++ isn't up to version 17, it's up to version 16 (aka, Visual Studio 2019). VC14 is Visual Studio 2015 and VC15 is Visual Studio 2017. So basically we've already got a Visual Studio 2017 solution set up. :)

I have it locally set to compile for C++17.

I'm going to look over everything here and see how it feels and move from there. My biggest concern is Linux and macOS support, though macOS support is of lower priority for me than Linux. Dan can weigh in on the Linux side of things... though I'm inclined to set myself up on a VM with some flavor of Linux (probably Ubuntu, I'm most familiar with that) and get more experience with it.

Anyway, will get back to you later tonight on this. I've seem a lot of interesting commits... so bear with me while I force myself to be alive again. :D

cugone:
@ldicker83 
That's an error on my part then, it was supposed to mean "VS17/C++17" and "VS19/C++20". Chalk it up to no sleep. :)

That's the one thing MSVC is a little confusing about is the numbers don't coincide with the IDE product naming numbers and they're *juuust* slightly off to be confusing.

If need be you can delete the createVc17projectfiles branch if it's already setup correctly in the vc15 folder. But last I looked it was missing the packages.config file and the .vcxproj files may need a few changes to the additional includes (relative path to the global include folder) and the Platform Toolset may need changing to "$(DefaultPlatformToolset)" because AppVeyor requests it that way.

The reason I made that newer branch is because @DanRStevens requested some changes that some portions of this PR could be split off into separate smaller PRs and since all the subsequent changes hinged on the features of VS2017/VC15 that was a requirement to be pushed first.

In addition I knew I could make the branch just for the project files and not break anything with the smaller requests like the string utilities and swapping the typedefs for using declarations.

ldicker83:
I'm actually not too worried about directory naming for the projects. I haven't committed my C++17 changes to the project so I'm willing to just move on with VC17 for the new project folder.

I'm more concerned about some copy/paste errors I'm seeing in Filesystem. I'll start marking issues as I see them. Wish I could combine all of your commits into one change so I could review the entire file... maybe that's what git squash is? I need to get more informed with how to use git well.

include/NAS2D/Filesystem.h
```diff
@@ -50,16 +57,64 @@ class Filesystem
 
 	void toggleVerbose() const;
```
ldicker83: The verbosity flag was removed in a prior commit in master -- I saw no need to keep verbose logging intact.

include/NAS2D/Filesystem.h
```diff
@@ -50,16 +57,64 @@ class Filesystem
...
+    void setWorkingDirectory(const std::filesystem::path& p);
+    std::filesystem::path getWorkingDirectory();
+
 private:
-	Filesystem(const Filesystem&) = delete;
-	Filesystem& operator= (const Filesystem&) = delete;
 
-	bool closeFile(void *file) const;
-	const char* getLastPhysfsError() const;
+    std::filesystem::path getExePath() const;
```
ldicker83: "EXE" is a Windows term for a binary -- I think most programmers/users would understand exactly what it means but perhaps there's a better name for this? Executable? Binary? App?

This is only a minor thing, really and is up for discussion, it can be changed in later versions if needed.

include/NAS2D/Filesystem.h
```diff
@@ -50,16 +57,64 @@ class Filesystem
...
+    std::string                        mOrganizationName{};    /**< The organization name. Only used for compatibility while transitioning from PhysFS */
+    std::string                        mAppName{};             /**< The application name. Only used for compatibility while transitioning from PhysFS */
+    std::filesystem::path              mWorkingDirectory{};  /**< The working directory. Typically 'mDataPath/mOrganizationName/mAppName/' */
+    std::filesystem::path              mExePath{};           /**< Path to the executable.*/
+    mutable bool                       mVerbose{ false };      /**< Displays lots of messages when true. Otherwise only critical messages are displayed. */
+    mutable bool                       mIsInit{ false };       /**< Has the file system been initialized? */
 };
 
+template<typename DirectoryIteratorType>
+void forEachFileInFolders(const std::filesystem::path& preferred_folderpath, const std::vector<std::string>& validExtensions, const std::function<void(const std::filesystem::path&)>& callback) {
```
ldicker83: Brace placement doesn't conform to the rest of the code. It's again a minor thing but I would argue that consistency is important regardless of placement choice. I usually go with brace placement on a new line.

src/Filesystem.cpp
```diff
@@ -33,48 +42,31 @@ enum MountPosition
...
 /**
- * Shuts down PhysFS and cleans up.
+ * Initializes PhysFS
  */
-void Filesystem::init(const std::string& argv_0, const std::string& appName, const std::string& organizationName, const std::string& dataPath)
+void Filesystem::init(const std::string& /*argv_0*/, const std::string& appName, const std::string& organizationName, const std::string& dataPath)
 {
-	if (PHYSFS_isInit()) { throw filesystem_already_initialized(); }
+    namespace FS = std::filesystem;
+    if(mIsInit) { throw filesystem_already_initialized(); }
```
ldicker83: Should be a call to function versus direct member access. This is a stylistic choice that I tend to use in all of my code, results in fewer changes should there be a different way to determine init status. Also, consistency.

include/NAS2D/Filesystem.h
```diff
@@ -50,16 +57,64 @@ class Filesystem
...
+    void forEachFileInFolder(const std::filesystem::path& folderpath, const std::string& validExtensionList = std::string{}, const std::function<void(const std::filesystem::path&)>& callback = [](const std::filesystem::path& p) { (void)p; }, bool recursive = false) const;
+
+    bool readBufferFromFile(std::vector<unsigned char>& out_buffer, const std::string& filePath) const;
+    bool readBufferFromFile(std::string& out_buffer, const std::string& filePath) const;
+
+    bool writeBufferToFile(void* buffer, std::size_t size, const std::string& filePath) const;
+    bool writeBufferToFile(const std::string& buffer, const std::string& filePath) const;
+
+
+    mutable std::vector<std::filesystem::path> mSearchPath{};
```
ldicker83: I'm unfamiliar with this style of in-place initialization -- was this C++11 or later? I'm not at all opposed, just curious. I usually use '='. Is there a difference?

src/Filesystem.cpp
```diff
@@ -110,59 +103,30 @@ bool Filesystem::mount(const std::string& path) const
...
-	auto searchPathList = PHYSFS_getSearchPath();
-	for (char **i = searchPathList; *i != nullptr; ++i)
-	{
-		searchPath.push_back(*i);
-	}
-	PHYSFS_freeList(searchPathList);
+    if(!isInit()) { throw filesystem_not_initialized(); }
 
+    StringList searchPath{};
+    for(const auto& p : this->mSearchPath) {
```
ldicker83: Brace placement. Condense into a single line:

'for(const auto& p : this->mSearchPath) { searchPath.push_back(p.string()); }

src/Filesystem.cpp
```diff
@@ -176,15 +140,23 @@ StringList Filesystem::directoryList(const std::string& dir, const std::string&
...
-	{
-		std::cout << "Unable to delete '" << filename << "':" << getLastPhysfsError() << std::endl;
-		return false;
-	}
-
-	return true;
+    if(!isInit()) { throw filesystem_not_initialized(); }
+
+    namespace FS = std::filesystem;
+    try {
```
ldicker83: Brace placement.

src/Filesystem.cpp
```diff
@@ -250,8 +196,10 @@ File Filesystem::open(const std::string& filename) const
  */
 bool Filesystem::makeDirectory(const std::string& path) const
 {
-	if (!PHYSFS_isInit()) { throw filesystem_not_initialized(); }
-	return PHYSFS_mkdir(path.c_str()) != 0;
+    namespace FS = std::filesystem;
+    if(this->isInit()) { throw filesystem_not_initialized(); }
```
ldicker83: This appears to be incorrect -- is the intention to check that the FS is initialized and if not to throw?

Additionally, why the explicit 'this' pointer?

src/Filesystem.cpp
```diff
@@ -262,10 +210,10 @@ bool Filesystem::makeDirectory(const std::string& path) const
  */
 bool Filesystem::isDirectory(const std::string& path) const
 {
-	if (!PHYSFS_isInit()) { throw filesystem_not_initialized(); }
+    namespace FS = std::filesystem;
+    if(this->isInit()) { throw filesystem_not_initialized(); }
```
ldicker83: This appears to be incorrect -- is the intention to check that the FS is initialized and if not to throw?

Additionally, why the explicit 'this' pointer?

src/Filesystem.cpp
```diff
@@ -417,29 +322,178 @@ std::string Filesystem::workingPath(const std::string& filename) const
...
+}
+
+void Filesystem::forEachFileInFolder(const std::filesystem::path& folderpath, const std::string& validExtensionList /*= std::string{}*/, const std::function<void(const std::filesystem::path&)>& callback /*= [](const std::filesystem::path& p) { (void)p; }*/, bool recursive /*= false*/) const {
+    namespace FS = std::filesystem;
+    auto preferred_folderpath = folderpath;
+    preferred_folderpath.make_preferred();
+    bool exists = FS::exists(preferred_folderpath);
+    bool is_directory = FS::is_directory(preferred_folderpath);
+    bool is_folder = exists && is_directory;
+    if(!is_folder) {
```
ldicker83: Condense into one line

DanRStevens:
Whoops, my bad about the version numbers. That is rather confusing:
VS2017/VC15/C++17

Perhaps we should consider renaming the project folder to "vs2017", rather than "vc15"? I think people would be more familiar with Visual Studio version numbers than Visual C++ version numbers. Could also be explicit with "vs2017.vc15".

Sorry for the trouble.

----

With that said, you're right there are a few project files settings that can be updated. I also noticed the packages file was not present in the "vc15" folder. We should probably fix that. Additionally, AppVeyor is still building with the "vc14" project folder. We'd probably need to fix the packages before changing that. The hardcoded include paths also need updating. I suspect they're being used by Leeor though. From what I understand, Leeor is still able to build locally, despite the Windows CI build being broken. I believe Leeor's local setup has an updated PhysFS in one of those directories, which is satisfying the dependency. The SDL files can and probably should come from the NuGet packages though, so we can likely trim down on the external includes. Would also seem cleaner to use relative includes to dependencies which are subfolders of the project (and .gitignored).

I was not aware of the "$(DefaultPlatformToolset)" setting. That's good to know about.

----

I think converting `typedef` to `using` is a fairly safe change without consequence. That may be worth separating into a branch.

Using move semantics for the string operations is worth separating out into another branch.

include/NAS2D/Filesystem.h
cugone: I like Binary, but concede to future discussion or consensus of the group. A preliminary googling of "What are executables called in Linux?" and "What are executables called in MacOSX?" returns ".app" for Mac and "Whatever the hell you want because Linux uses the file permissions to determine executability and with what program." in Linux.

include/NAS2D/Filesystem.h
cugone: This was a standards conformance change that was added to VS2017 late last year, I can't find changelog that mentions the exact version of 15.x it was. It's called default member initialization and prevents out-of-order initialization in c'tor initializer lists errors (most novice and beginner modern c++ programmers don't know about the rule that members initialize in order of declaration in the header not in order of initializer list in the implementation) and makes it so you only have to initialize the non-default members in the initializer list or are able to get rid of default c'tors entirely by either with the use of "= default" (good) or not declaring them at all and using the compiler generated c'tor (better) instead. A very useful change to the language that I've started to love.

src/Filesystem.cpp
cugone: The explicit 'this' pointer was me fighting with intellisense and I forgot to remove it. :P

src/Filesystem.cpp
cugone: This was a "readability over terse code" choice. Much easier to understand what the code is doing, but I'll change it.

src/Filesystem.cpp
cugone: Copy-Pasta bug from previous function. 'this' pointer was a fight with initellisense and VA, just forgot to delete it.

src/Filesystem.cpp
cugone: Changed if-statement to single line.

cugone:
> 
> I was not aware of the "$(DefaultPlatformToolset)" setting. That's good to know about.
> 
@DanRStevens  Neither was I until I read it in AppVeyor's documentation. Promptly went and looked it up in the Macro list of the property pages and I'll be darned, there it was!

src/Filesystem.cpp
```diff
@@ -33,48 +42,31 @@ enum MountPosition
...
-
-	mDataPath = dataPath;
-	if (PHYSFS_mount(mDataPath.c_str(), "/", MountPosition::MOUNT_PREPEND) == 0)
-	{
-		std::cout << std::endl << "(FSYS) Couldn't find data path '" << mDataPath << "'. " << getLastPhysfsError() << "." << std::endl;
-	}
+    mOrganizationName = organizationName;
+    mAppName = appName;
+    mDataPath = FS::path{ dataPath };
+    mWorkingDirectory = mDataPath / FS::path{ mOrganizationName } / FS::path{ mAppName };
```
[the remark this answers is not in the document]
cugone: @DanRStevens 
@ldicker83 

I'm a little confused on what you expect the implementation to return for mWorkingDirectory (and the up-coming mUserDirectory and mPrefDirectory) for windows and linux.

From what I can gather from @DanRStevens comment on pull #[122](https://github.com/lairworks/nas2d-core/pull/122#issuecomment-480472041)

For windows:

- mWorkingDirectory is project-dependent but generally:
  - Is the root path that is used when a relative path is hard-coded. e.g. "myfile.txt"; or "Data/myfile.txt".
  - Defaults to where the .exe resides and is *usually* argv[0] and now a call to `std::filesystem::current_directory();` returns the "current" working directory and a call to `std::filesystem::current_directory(path);` and can change it at run-time.
  - Windows has no concept of "home directory" and the term is ambiguous. Do you mean The user's profile? The User's Desktop? The All-Users Desktop? The All-Users Profile? The .exe directory? These are all different things and have different semantics.
- mPrefDirectory is project- and business-rule dependent:
  - %USERPROFILE%/AppData/Roaming/{mOrganizationName}/{mAppName}/
    - Used when the files are allowed to migrate between PCs and Users.
  - %USERPROFILE%/AppData/Local/{mOrganizationName}/{mAppName}/
    - Local is used when it is machine-dependent like GPU shaders/hardware-specific files.
  - %USERPROFILE%/AppData/LocalLow/{mOrganizationName}/{mAppName}/
    - LocalLow is used for reduced-permissions files.
- mUserDirectory is ambiguous and could mean either:
  - %USERPROFILE%
  - %USERPROFILE%/Desktop

DanRStevens:
@cugone, it looks like you're rather enjoying yourself. Also looks like you've learned quite a bit about the platform difference challenges.

I was speaking with Leeor, and we're both in agreement that we'd like some of your changes in sooner rather than later. I'm still concerned about platform breaks and testing, though that only applies to part of you changes. Leeor seems fine with breaking things and fixing them later. I'm a little more conservative, preferring to keep master working on all platforms. In particular, I develop mainly from Linux, so I don't want to see breaks there, as that could impact my ability to work on stuff.

So, how familiar are you with interactive Git rebase?

I think we could pull out some of your changes that don't impact platform concerns and merge them immediately. With Git interactive rebase, I believe we can pull out those changes without having to redevelop them. Rebase can be a powerful tool to help with refactoring. I can help a bit with it if needed.

I think it might also be good practice in general for working with big changes like this, and packaging them into PRs that are easy to review and accept. It occurs to me if the number of developers are growing on this project, we may need to retire some of our cowboy coding ways of the past, and practice methods that scale well to team development.

Looking forward to working with you.

cugone:
@DanRStevens 

> ...I develop mainly from Linux, so I don't want to see breaks there, as that could impact my ability to work on stuff. ...

No problem there. I've been trying to keep the Linux build from breaking but I can only really confirm if the code compiles after I push it because I do not have a local Linux distro/build chain (I wouldn't even know where to start getting one for Windows, but I hear Win10 added shell bash support).

Apple-based proprietary GCC/Clang implementations and library maintainers can have their toes lit on fire...Every other modern compiler at least *tries* to stay up-to-date with changes in the standard and library implementations within at most six months...Not *five years* behind.

 I develop exclusively for and from Windows and DirectX (bad experiences in the past has made me shy away from Linux, unfortunately. Plus, since learning DirectX I find it's more fun than using Allegro or SDL graphics libraries and allows more finer control than OpenGL :P ). So "platform differences" usually meant looking up the MSDN documentation and answering the question: "How do I do this in Windows vs plain C++"

> 
> So, how familiar are you with interactive Git rebase?
> 

Not at all. I've only recently started using the Visual Studio Developer Command Prompt to access features in the git command line that the built-in VS git source control plugin did not provide. Before that I used Perforce (has a scripting functionality, but I've never used it) and before that it was Tortoise/SVN and Git for Windows Desktop. All of which are geared towards GUI SCM.

>
> I think we could pull out some of your changes that don't impact platform concerns and merge them immediately. With Git interactive rebase, I believe we can pull out those changes without having to redevelop them. Rebase can be a powerful tool to help with refactoring. I can help a bit with it if needed.
> 

That's good, and I will definitely need help. It's become a habit to google for and pull up the --help file of a git command before using it just so I can familiarize myself and not horribly break anything. Rebase is one of those commands that any time the docs say "Oh, just rebase!" they litter the text with "CAUTION! You will totally mess everything up if you do this wrong!" and that terrifies me.

>
> I think it might also be good practice in general for working with big changes like this, and packaging them into PRs that are easy to review and accept. It occurs to me if the number of developers are growing on this project, we may need to retire some of our cowboy coding ways of the past, and practice methods that scale well to team development.
> 
> Looking forward to working with you.
>

Yeah that's one thing I keep having to force myself to do when working with Git is "lots of tiny commits, one big push". Which is totally contrary to working in Perforce, since the mantra there is "Don't Make Perforce Mad", you end up with "push every small change often" because you can't separate multiple changes in a file across changelists/commits like you can in Git. It's all or nothing. Hence my first few commits to this PR reflected that mentality.

I'm versed in Agile (with Scrum, but that's hard to do over text. :P ) development and use Jira and Trello all the time for team task management. The Git Projects board/Task trackers look interesting. It may also help to have everything written down. (Granted we had wikis and design docs at The Guildhall at SMU and guess how many times those were read/updated after their first draft?)

It's been fun so far! :D

cugone:
@DanRStevens 
@ldicker83 

I have half a mind to close this PR.  It has turned into a testbed that just confuses things, what with the move to the vs2017 folder and itself still being in vc17 and its build CIs being broken with old configs.

It only took about an hour to make those initial changes to Filesystem and it will likely go faster now that I know how to deal with the platform differences. If I need to reference the old code in any way I can just look up the closed PR.

DanRStevens:
No worries if you need to close a branch, or use a branch experimentally. As there are so many commits, it may be difficult to rebase this one. That's made worse by some updates having already been merged in with changes.

Mind you, I would still encourage you to give rebasing/refactoring it a try, even if you don't complete the task. It would be good practice for some very useful skills.

It could also cut down on the bulk you have to refer back to later when it's not so fresh in your mind.

cugone:
Backed up files/changes locally. Deleting PR.

PR #122, opened 2019-04-05
Remove `Filesystem::userPath`

### LINKED ISSUE #84: Update calls to deprecated PhysFS functions
A number of PhysFS functions have been deprecated, with suggested alternatives:
- [`getUserDir`](https://icculus.org/physfs/docs/html/physfs_8h.html#a9d86fca7a5cbbc53f5425ff7539b1426) ==> [`getPrefDir`](https://icculus.org/physfs/docs/html/physfs_8h.html#acd87392d234d070695303521bb8052a5) (or [`setSaneConfig`](https://icculus.org/physfs/docs/html/physfs_8h.html#afef8700f714b6800ff688372c540bfe2))
- [`addToSearchPath`](https://icculus.org/physfs/docs/html/physfs_8h.html#a0956d926822291a23ee540c2ffefa1af) ==> [`mount`](https://icculus.org/physfs/docs/html/physfs_8h.html#a8eb320e9af03dcdb4c05bbff3ea604d44)
- [`getLastError`](https://icculus.org/physfs/docs/html/physfs_8h.html#a9f78e4ab560885fd10202988d2fff1b6) ==> [`getLastErrorCode`](https://icculus.org/physfs/docs/html/physfs_8h.html#ae916a873aae2cf6b6e09d733cd1b3f4b) and [`getErrorByCode`](https://icculus.org/physfs/docs/html/physfs_8h.html#a545f6b5ba97faa19441932dffb65c3be)
- [`read`](https://icculus.org/physfs/docs/html/physfs_8h.html#a91234f576477f7de67d635761d01aa7b) ==> [`readBytes`](https://icculus.org/physfs/docs/html/physfs_8h.html#a63ac40ecdb3f9b91e6ee309bdbad0ce8)
- [`write`](https://icculus.org/physfs/docs/html/physfs_8h.html#af5dc83d5e4ef0b0c9bce1834ea1bb498) ==> [`writeBytes`](https://icculus.org/physfs/docs/html/physfs_8h.html#a192ae4f946656ae2aef5e16e438f94e1)
- [`isDirectory`](https://icculus.org/physfs/docs/html/physfs_8h.html#a5877fcafb72dca892c1c2465292eef47) ==> [`stat`](https://icculus.org/physfs/docs/html/physfs_8h.html#a7f5e23675c50ffe51ad00cceb5a65348)

Some of these calls could be updated using the current older version of PhysFS. This would reduce the warnings when using a newer library version, while still maintaining compatibility with the older version. Other updates may be part of an upgrade effort that breaks PhysFS version compatibility (such as `getPrefDir` and `setSaneConfig`).

DanRStevens:
I'm not sure if we need to support this method. It's not used in OPHD. It's implementation uses a deprecated PhysFS method. Do we need this?

Closes #84.

I don't mind providing a method such as this, but I think we need some clarity on what it is we want to support. In particular, do we want a `uesrPath` method, which might not be so useful for a game, but probably useful for a graphics editor and picture viewing program. Or do we want a `writePath` method, which returns the real filesystem path of the write folder? That might be useful in reporting paths to the end user, or letting them know where their files are stored. The previous code conflated these two ideas.


ldicker83:
I believe a userPath variable would be very useful to game developers. Many modern games are now using the User folder to store savegames, etc. versus writing those directly into their own working directory (or as in the case of OPHD, in root/data).

DanRStevens:
That's already supported without this method. The default write path is a location under the under the user folder (but not directly in it). That's part of the `setSaneConfig` setup.

From #84:

> > Deprecated:
> > As of PhysicsFS 2.1, you probably want PHYSFS_getPrefDir().
> 
> PHYSFS_getUserDir() : Get the path where user's home directory resides.
> PHYSFS_getPrefDir() : Get the user-and-app-specific path where files can be written.

By default, saved games will end up in `getPrefDir`. Should we have a method to return that?

DanRStevens:
I'm going to close this change as undesirable. Maybe there's another way we can silence the warning meanwhile.

DanRStevens:
Re-opened as per the discussion in #84. @ldicker83 if you want the method gone, along with the associated deprecation warning, just click the merge button. :wink:

PR #126, opened 2019-04-05
Create vc17 project files

cugone:
Separate PR for just the VC17 project files


cugone:
@DanRStevens I need sleep. I've forgotten how to tell VS to use the packages.config file. The entries "Project Master Settings" and "Referenced Packages" under the project settings do not appear for this branch but they do appear in the for-testing vc17 branch.

ldicker83:
So for whatever reason it's failing because it can't find the SDL and GLEW dependency headers. Probably just a config issue with nuget -- I'm still somewhat unfamiliar with it but I suspect that once that's fixed and the win32 build succeeds we can merge this into mainline.

cugone:
@ldicker83  Cool, cool. Here's some resources on how to do that:
 - https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package
 - https://docs.microsoft.com/en-us/nuget/quickstart/install-and-use-a-package-in-visual-studio
 - https://docs.microsoft.com/en-us/nuget/tools/package-manager-ui
 - https://docs.microsoft.com/en-us/nuget/consume-packages/ways-to-install-a-package

cugone:
@ldicker83 
There are some function definitions that can't be found, but that may be a due to version mismatching, the NuGet server only has 2.0.3 when the deprecated functions list mentions 2.1.0 and the official release site uses 3.0.2

Brett208:
@cugone & @ldicker83,

physfs is available via vcpkg as version 3.0.2. I believe vcpkg comes as prepackaged with appveyor. If we can get nas2d to build properly when using physfs from vcpkg, would you be willing to accept a split between using nuget and vcpkg for the different configurations to get appveyor automated builds working? I can test trying vcpkg this weekend, but do not have time tonight.

In my experience, vcpkg builds are kept closer up to date with newest release than nuget packages (although there are different pros and cons to both package managers)

https://github.com/Microsoft/vcpkg/blob/master/ports/physfs/CONTROL 

-Brett

cugone:
@Brett208 and @ldicker83 

I like that physfs is actually up-to-date.

I just tried this locally and getting the physfs headers recognized and found is a pain. Maybe I'm doing something wrong or it requires CMAKE to be fully integrated and appreciated in a project. Which I do not know how to use.

DanRStevens:
@Brett208, that's an excellent idea. I think that's exactly the sort of stop-gap solution we need to get the Windows builds working again. Might even become a permanent thing if the packages are more up-to-date, though ideally I'd like an automated package restore solution that works for developers.

From what I understand, Vcpkg must be integrated into the environment before the build will pick up Vcpkg installed packages. @cugone, it may be the integrate step you were missing.

I've managed to get a build working in a Windows virtual machine using Vcpkg. I'll try to push the changes into a new branch for testing on AppVeyor.

cugone:
@DanRStevens 

Fixed. I forgot to specify `:x64-windows`. Re-ran install as `vcpkg install:x64-windows` and all is good.

DanRStevens:
@cugone, I'm on IRC for a little while if you want to chat realtime about how we might integrate your changes:
https://www.outpost2.net/chat.html

cugone:
@DanRStevens 

Just hopped in to the channel.

DanRStevens:
The ideas from this PR have been repackaged into #130 and #131, which were instrumental in fixing the Windows CI build. I suggest closing this PR now.

Issue #123, opened 2019-04-05
MacOS builds on CircleCI


DanRStevens:
Currently the MacOS builds are being done on TravisCI. The experience on TravisCI has been kind of painful, with slow and often delayed builds. Additionally TravisCI doesn't seem to prioritize using updated compilers and standards. This may hold us back from using more modern compiler features.

We should check into doing MacOS builds on CircleCI. They may now offer MacOS as part of the free tier.

Reference:
https://circleci.com/docs/2.0/hello-world-macos/

cugone:
@DanRStevens 
@ldicker83 

MacOS builds on CircleCI under lairworks have been approved. It's equivalent to the "Seed" plan but perpetually free and 1x concurrency.

cugone:
@DanRStevens 
@ldicker83 

Good news:
 - MacOS builds are upwards of 30x faster on CircleCI.
 - Linux builds are 7x faster on CircleCI.

Bad news:
 - most recent stable xcode 10.2 still does not have full C++17 STL support as of its release 13 days ago.
   - They even *removed* the filesystem header and library support that existed in 10.2 beta 4!
   - https://forums.developer.apple.com/message/354637#354637
 - CircleCI currently only supports 10.0
 - It's entirely a waiting game for Apple to get their affairs in order then CircleCI to install the appropriate release.

The Internet is fed up with xcode being not-up-to-date and I can now see why so few cross-platform apps and games don't include MacOS in that list.

PR #129, opened 2019-04-07
Add MacOS builds to CircleCI

cugone:
Integrated macos into CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,48 @@
-version: 2.0
-jobs:
-  build:
+version: 2.1
+executors:
+  docker-executor:
     docker:
       - image: outpostuniverse/ubuntu-18.04-gcc-sdl2-physfs-circleci
+  macos-executor:
+    macos:
+      xcode: "10.0.0"
+jobs:
+  build-linux:
+    executor: docker-executor
     steps:
       - checkout
-      - run: make -k
+      - run: make --keep-going
+  test-linux:
+    executor: docker-executor
+    steps:
+     - run: echo "Running Fake Linux Tests..."
+  build-macos:
+    executor: macos-executor
+    steps:
+      - checkout
+      - restore_cache:
+          key: deps-v2-NAS2D-{{ arch }}
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
+      - run: brew link physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
+      - save_cache:
+          paths:
+            - /Users/distiller/Library/Caches/Homebrew
+            - /usr/local/Cellar
+          key: deps-v2-NAS2D-{{ arch }}
+          name: Caching brew install deps
+      - run: make --keep-going
+  test-macos:
+    executor: macos-executor
+    steps:
+     - run: echo "Running Fake MacOS Tests..."
+workflows:
+  build_and_test:
+    jobs:
+      - build-linux
+      - build-macos
+      - test-linux:
+          requires:
+            - build-linux
+      - test-macos:
+          requires:
+            - build-macos
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
   test-linux:
     executor: docker-executor
     steps:
-     - run: echo "Running Fake Linux Tests..."
+     - run: make --keep-going check
   build-macos:
     executor: macos-executor
     steps:
@@ -34,7 +34,7 @@ jobs:
   test-macos:
     executor: macos-executor
     steps:
-     - run: echo "Running Fake MacOS Tests..."
+     - run: make --keep-going check
 workflows:
   build_and_test:
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,10 +12,7 @@ jobs:
     steps:
       - checkout
       - run: make --keep-going
-  test-linux:
-    executor: docker-executor
-    steps:
-     - run: make --keep-going check
+      - run: make --keep-going check
   build-macos:
     executor: macos-executor
     steps:
@@ -31,18 +28,9 @@ jobs:
           key: deps-v2-NAS2D-{{ arch }}
           name: Caching brew install deps
       - run: make --keep-going
-  test-macos:
-    executor: macos-executor
-    steps:
-     - run: make --keep-going check
+      - run: make --keep-going check
 workflows:
   build_and_test:
     jobs:
       - build-linux
       - build-macos
-      - test-linux:
-          requires:
-            - build-linux
-      - test-macos:
-          requires:
-            - build-macos
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - checkout
       - run: make --keep-going
-      - run: make --keep-going check
+      - run: echo "Testing disabled" || make --keep-going check
   build-macos:
     executor: macos-executor
     steps:
@@ -28,7 +28,7 @@ jobs:
           key: deps-v2-NAS2D-{{ arch }}
           name: Caching brew install deps
       - run: make --keep-going
-      - run: make --keep-going check
+      - run: echo "Testing disabled" || make --keep-going check
 workflows:
   build_and_test:
     jobs:
--- a/proj/vs2017/NAS2D.vcxproj
+++ b/proj/vs2017/NAS2D.vcxproj
@@ -136,7 +136,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
       <AdditionalIncludeDirectories>../../include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
@@ -149,7 +149,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <ClCompile>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
       <Optimization>MaxSpeed</Optimization>
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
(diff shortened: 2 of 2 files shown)


Brett208:
It looks like the macOS docker image is not downloading the dependent projects

 - physfs
 - sdl
 - glew

I see the Macintosh xcode project files in the proj folder. I see circleci is targeting xcode 10.0.0. I don't understand how circleci knows where the xcode project files are located at. 

@ldicker83, does Trevor Allen still work on nas2d, or could he give an assist maybe to properly hook the xcode project to circleci?

Thanks,
Brett

cugone:
@DanRStevens 
@ldicker83 

All builds pass

cugone:
@DanRStevens 

This works, but macOS takes four minutes to rebuild the brew deps every time.

Alright if I create another branch/PR just to address testing caching?

If it doesn't work I can just delete it and we'll still have this to fall back on.

.circleci/config.yml
```diff
@@ -1,9 +1,26 @@
...
       - run: make -k
+  build-macos:
+    executor: macos-executor
+    steps:
+      - checkout
+      - run: brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
+      - run: make -k
+workflows:
+  version: 2
+  linux_macos_build:
```
DanRStevens: I think I might just name this `build` as a way to reduce repetition. The extra detail doesn't really add anything, though the name could become misleading or add to maintenance if the steps below change. For instance, we might someday add a Windows build, or remove one of the existing Linux or MacOS builds. Ideally we shouldn't have to update the name here to keep it in sync.

.circleci/config.yml
```diff
@@ -1,9 +1,26 @@
...
-jobs:
-  build:
+version: 2.1
+executors:
+  docker-executor:
     docker:
       - image: outpostuniverse/ubuntu-18.04-gcc-sdl2-physfs-circleci
+  macos-executor:
+    macos:
+      xcode: "10.0.0"
```
DanRStevens: I looked quickly into what `executors` are. I'm not sure we're actually really using this feature in any special way. If the `docker`/`macos` sections were moved below the specific jobs, I think it would function identically.

This leaves me feeling like I've missed something about how to effectively use `executors`.

.circleci/config.yml
```diff
@@ -1,9 +1,26 @@
...
+    executor: docker-executor
     steps:
-      - run: brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
       - checkout
       - run: make -k
+  build-macos:
+    executor: macos-executor
+    steps:
+      - checkout
+      - run: brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
```
DanRStevens: If the `brew install` commands were moved ahead of checkout, that would make the remaining build steps identical between platforms. That might enable use of `commands` to reduce repetition.

DanRStevens:
Tagging Issue #123.

.circleci/config.yml
```diff
@@ -1,9 +1,26 @@
...
+      - run: make -k
+workflows:
+  version: 2
```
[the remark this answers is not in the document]
cugone: I was going off the official docs https://circleci.com/docs/2.0/configuration-reference/#workflows
that say "currently must be 2." I guess they haven't updated it. :/

.circleci/config.yml
cugone: https://circleci.com/docs/2.0/configuration-reference/#workflows
https://circleci.com/docs/2.0/configuration-reference/#jobs

 - Workflow names must be unique.
 - Job names must be unique.
 - The key "build" has a special meaning.

Since this workflow encompassed building linux and then macos, that's what it was named. It's trivial to change it when needed, especially if you are already changing the file for something else.

.circleci/config.yml
cugone: It's also useful for organization and readability. By having the environment definition separated out from the build job it reduces noise and makes it easier to read and follow what the job is supposed to be doing.

.circleci/config.yml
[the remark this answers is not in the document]
DanRStevens: I think you're right. It probably is old documentation that hasn't been updated. I've seen many examples in their documentation that specify `version` both at the top level and within the `workflows` section. I always thought that was really strange, especially when the version numbers didn't match exactly.

.circleci/config.yml
cugone: - `checkout` must occur before a cache restore.
 - `brew install` and `brew link` are part of the restore/save cache block and cannot be moved outside of it.
 - See #135

.circleci/config.yml
DanRStevens: From what I read, executors are useful when you re-use an environment. Here we are only using each environment once, so we're not really benefiting from that.

I'm not sure it does anything for readability here.

I think readability might be better improved by pulling out the build and test commands and re-using that. That's where the duplication really is.

.circleci/config.yml
DanRStevens: It's still an extra detail that has to be remembered. We should strive to reduce extra details like that.

The job name "build" is only special when not using workflows. Names must be unique (within a config.yml), but I think that's only with a section (workflow or job, not unique among both at the same time).

----

What about naming the workflow `build-and-test`? That's more descriptive of what it does (build and test), rather than how it does it (on linux and macos). It's the building and testing that's important, more so than what systems that's done on.

.circleci/config.yml
DanRStevens: Can you provide a reference for the `checkout` restriction? That seems surprising.

.circleci/config.yml
```diff
@@ -16,8 +16,17 @@ jobs:
     executor: macos-executor
     steps:
       - checkout
-      - run: brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
+      - restore_cache:
+          key: deps-v2-NAS2D-{{ arch }}
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
+      - run: brew link physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
       - run: make -k
+      - save_cache:
```
DanRStevens: I think the `save_cache` should be moved directly after the `brew` commands. The `make -k` should be done after the cache save.

.circleci/config.yml
cugone: https://circleci.com/docs/2.0/configuration-reference/#section=reference
 - Shows checkout as first step in a job

https://www.youtube.com/watch?v=0-ePDGv1qLc
 - Shows checkout as first step in a job

.circleci/config.yml
cugone: I don't remember why I put it in the restore/save block in the first place. It probably can be moved since it doesn't touch or modify the brew download folder or the brew install folder.

.circleci/config.yml
cugone: Adding test jobs now makes using executors "useful".

.circleci/config.yml
cugone: Changed as requested. :)

.circleci/config.yml
DanRStevens: Huh, I wouldn't have thought to use executors like that. I would have just appended the test step after the build step.

I see using multiple executors results in more result output in the GitHub interface. It now reports build and test pass/fail results separately.

Actually, I think it spins up a new Docker instance for the tests when you do this. That likely means it would need to compile the code again in the new instance before running tests. It also doubles the container startup time.

We may need to do this for branch versus PR builds, but I don't think we should be doing this for build and test. Build and test should both be done in the same container.

.circleci/config.yml
DanRStevens: I didn't see any restriction mentioned in those links. Just a few examples.

In one example I saw, they were caching the source code to speed up the Git checkout. That meant putting a cache restore before the `checkout` step, and a cache save after it. In other examples they are caching dependencies installed by tooling, where the source code contains a file which lists the dependencies. In those cases the `checkout` step was (arbitrarily) placed before the cache restore. After the cache restore they then install dependencies, and save them to the cache.

In our case, we have no list of dependency packages in version control, and no tool running on MacOS to install those dependencies from a list stored in a file. Instead, we expect the dependencies to be pre-installed as part of the environment. This is much like when a developer does a one-time install of dependencies to their local machine, and then just builds repeatedly from then on. That would be matched by doing a cache restore, global environment install of dependencies, a cache save, and then finally all the needed build steps, which might include checking out or updating code, building, and testing.

----

I think we can use the [`commands`](https://circleci.com/docs/2.0/configuration-reference/#commands-requires-version-21) feature here, with two commands. A generic "build-and-test" command for all platforms, and some kind of install dependencies (with caching) command for MacOS. The `jobs` section would then just invoke one or both commands, depending on the platform.

cugone: Found an example where caching the source is a good idea.
https://circleci.com/docs/2.0/caching/#source-caching
---
I expected there to be multiple tests performed that may be OS-dependent (Filesystem comes to mind immediately. Was also going to say DirectX but CircleCI doesn't do windows....yet) The workflow I imagined for each OS was: build on each OS, fan in to run common tests, fan out for OS-dependent tests, fan in for deploy on OS.

https://circleci.com/docs/2.0/workflows/#fan-outfan-in-workflow-example

But would you consider "any OS build breaking means the entire project is broken"? If so, then putting all the builds into one job makes sense. I don't think this is the case though.
It just means there may be one small thing wrong with that implementation for that OS and when it's fixed the one CircleCI job can be re-run.

DanRStevens:
Ok, I've gotten a bit behind on code updates. Sorry about that.

I'm afraid enough time has passed so that I've lost a bit of context on this one. I think I wanted some of the build steps pulled out into `commands`. We probably do need the `executors` so they can be re-used between branch and PR builds.

Looks like things here were working, and we were really close to getting it into the form we want. We've created a rather long wall of text though, that will be difficult for Leeor to review. Maybe we can do another rebase to clean things up a bit, and give this another look over. Once that is done perhaps we can port the changes to a new PR so we leave the wall of text behind.

@cugone, how do you feel about another rebase? I can potentially help if needed, though I'm afraid my availability may be a bit spotty for the next little while.

cugone:
@ldicker83 
@DanRStevens 

Request review.

.circleci/config.yml
```diff
@@ -1,8 +1,48 @@
...
+          paths:
+            - /Users/distiller/Library/Caches/Homebrew
+            - /usr/local/Cellar
+          key: deps-v2-NAS2D-{{ arch }}
+          name: Caching brew install deps
+      - run: make --keep-going
+  test-macos:
+    executor: macos-executor
+    steps:
+     - run: echo "Running Fake MacOS Tests..."
```
DanRStevens: I don't think testing should be a separate job.

.circleci/config.yml
```diff
@@ -1,8 +1,48 @@
...
+  build_and_test:
+    jobs:
+      - build-linux
+      - build-macos
+      - test-linux:
+          requires:
+            - build-linux
+      - test-macos:
+          requires:
+            - build-macos
```
DanRStevens: As I don't think testing should be a separate job, there should be no need for dependency here.

DanRStevens:
Hah, yeah, the test containers didn't even have a checked out copy of the code, never mind a compiled version to run through the unit tests.

I'll make some changes to address.

DanRStevens:
Ok, so that's what it should look like. It now tries to continue building the test code, which the original build will be linked into. Unfortunately there is no Google Test dependency installed in the CI environments, so it fails at the `#include <gtest/gtest.h>` line in the test code.

I'll disable actually running the tests for now.

PR #130, opened 2019-04-12
Use vcpkg for PhysFS

DanRStevens:
Install PhysFS 3.0.2 on AppVeyor using `vcpkg`.

----

Note: Any project specific NuGet package installs are likely to override the general environment `vcpkg` installs. As such, you'll want to remove the PhysFS NuGet package from a project to make use of the environment install. That makes this branch dependent on project file updates.

----

I tried removing the PhysFS package from the project in the `vc14/` folder. The AppVeyor build was having trouble finding the `vcpkg` installed version of PhysFS. Locally I was able to build, though was required to also update the targeted platform and toolset. That didn't seem appropriate for this branch, so I left it out, though may explain the difference.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,6 +6,7 @@ platform:
 cache:
   - C:\tools\vcpkg\installed\
   - C:\projects\nas2d-core\proj\vc14\packages -> proj\vc14\packages.config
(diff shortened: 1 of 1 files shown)


DanRStevens:
Merging as part of a change set to fix the Windows CI build.

PR #131, opened 2019-04-12
Create vs2017 project

DanRStevens:
This copies the `vc15/` project files to a `vs2017/` folder, and then applies changes from https://github.com/lairworks/nas2d-core/pull/126 on top. This gives an easier way to see the recent changes and improvements to the Visual Studio 2017 project file.

I then did some additional cleanup to remove absolute paths from the project file.

### DIFF
--- a/proj/vs2017/NAS2D.vcxproj
+++ b/proj/vs2017/NAS2D.vcxproj
@@ -1,0 +1,320 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release Gold|Win32">
+      <Configuration>Release Gold</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release Gold|x64">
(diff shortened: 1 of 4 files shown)


proj/vs2017/NAS2D.vcxproj
```diff
@@ -137,7 +137,7 @@
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
-      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>../../include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
cugone: This needs to be relative to a known macro path otherwise it will break if the project is ever moved.

I usually set it to $(SolutionDir): `$(SolutionDir)../../include;%(AdditionalIncludeDirectories)`

Changing while having "All Configurations" and "All Platforms" selected will also update all down-stream AID entries in this file.

DanRStevens: Relative directories should be fine. They will remain unchanged if the whole project is ever moved. I've used such paths in other projects quite extensively.

I'm a little uncertain if relative paths are relative to the project file or the solution file, though both are in the same folder, so would be the same.

----

You're right that we should probably look into removing some of the duplication in the project file at some point. Many settings are the same between all project configurations.

cugone: They are relative to whatever you set it to:
 - Solution: `$(SolutionDir)`: "Append the directory path where the solution lives here"
 - Project: `$ProjectDir)`: "Append the directory path where the project lives here"

It just happens that they both live in the same place for this project. Usually the solution is at a higher directory level than a project, or a project is in an entirely different location.

DanRStevens:
Merging as part of change set to fix the Windows CI build.

PR #140, opened 2019-04-15
Hook unit tests

Brett208:
Had to put in a lot of changes to the test project file to make work. 

The test project is building with toolset v141 now which matches the VS2017 solution file. One of the downsides to Visual Studio is that we would have to add another test project if we want to pair the tests with the VS2015 solution. Yuck.

I noticed there is a gtest nuget package update for the Microsoft gtest nuget package. I didn't incorporate yet, but we should explore. May improve our situation some sense we have some problems with the current package.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,5 @@
 image: Visual Studio 2017
-configuration: 
-  - Release
-  - Debug
+configuration: Release
 platform:
   - x86
   - x64
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,3 +13,7 @@ install:
   - nuget restore proj/vs2017
 build:
   project: proj/vs2017/NAS2D.sln
+test_script:
+  - cd %APPVEYOR_BUILD_FOLDER%\test
+  - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe'
+  - cd %APPVEYOR_BUILD_FOLDER%
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,6 +14,8 @@ install:
 build:
   project: proj/vs2017/NAS2D.sln
 test_script:
+  - echo Current working directory = '%CD%'
+  - echo Test exe path = '%APPVEYOR_BUILD_FOLDER%\proj\vs2017\%PLATFORM%\%CONFIGURATION%\test.exe'
   - cd %APPVEYOR_BUILD_FOLDER%\proj\vs2017
   - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe'
   - cd %APPVEYOR_BUILD_FOLDER%
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,5 @@ build:
   project: proj/vs2017/NAS2D.sln
 test_script:
   - echo Current working directory = '%CD%'
-  - echo Test exe path = '%APPVEYOR_BUILD_FOLDER%\proj\vs2017\%PLATFORM%\%CONFIGURATION%\test.exe'
-  - cd %APPVEYOR_BUILD_FOLDER%\proj\vs2017
-  - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe'
-  - cd %APPVEYOR_BUILD_FOLDER%
+  - echo Test exe path = '%APPVEYOR_BUILD_FOLDER%\test\%PLATFORM%\%CONFIGURATION%\test.exe'
+  - '%APPVEYOR_BUILD_FOLDER%\test\%PLATFORM%\%CONFIGURATION%\test.exe'
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,7 @@
 image: Visual Studio 2017
-configuration: Release
+configuration: 
+  - Release
+  - Debug
 platform:
   - x86
   - x64
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,6 +14,6 @@ install:
 build:
   project: proj/vs2017/NAS2D.sln
 test_script:
-  - cd %APPVEYOR_BUILD_FOLDER%\test
+  - cd %APPVEYOR_BUILD_FOLDER%\proj\vs2017
   - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe'
   - cd %APPVEYOR_BUILD_FOLDER%
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -16,6 +16,4 @@ install:
 build:
   project: proj/vs2017/NAS2D.sln
 test_script:
-  - echo Current working directory = '%CD%'
-  - echo Test exe path = '%APPVEYOR_BUILD_FOLDER%\test\%PLATFORM%\%CONFIGURATION%\test.exe'
   - '%APPVEYOR_BUILD_FOLDER%\test\%PLATFORM%\%CONFIGURATION%\test.exe'
--- a/test/packages.config
+++ b/test/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn" version="1.8.0" targetFramework="native" />
+  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn" version="1.8.1" targetFramework="native" />
 </packages>
--- a/test/packages.config
+++ b/test/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-static" version="1.8.1" targetFramework="native" />
+  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn" version="1.8.0" targetFramework="native" />
 </packages>
--- a/test/test.vcxproj
+++ b/test/test.vcxproj
@@ -33,12 +33,18 @@
   <ImportGroup Label="PropertySheets" />
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir>$(Platform)\$(Configuration)\</IntDir>
+    <OutDir>$(ProjectDir)x86\$(Configuration)\</OutDir>
+    <IntDir>x86\$(Configuration)\</IntDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
-    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir>$(Platform)\$(Configuration)\</IntDir>
+    <OutDir>$(ProjectDir)x86\$(Configuration)\</OutDir>
+    <IntDir>x86\$(Configuration)\</IntDir>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <OutDir>$(ProjectDir)$(Platform)\$(Configuration)\</OutDir>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <OutDir>$(ProjectDir)$(Platform)\$(Configuration)\</OutDir>
   </PropertyGroup>
   <ItemGroup>
     <ClCompile Include="Filesystem.test.cpp" />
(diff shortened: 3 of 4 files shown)


.appveyor.yml
```diff
@@ -13,3 +13,7 @@ install:
   - nuget restore proj/vs2017
 build:
   project: proj/vs2017/NAS2D.sln
+test_script:
+  - cd %APPVEYOR_BUILD_FOLDER%\test
+  - '%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe'
```
Brett208: I was using Windows specific directory separators here since it builds for Windows. Just realized the rest of the file had Linux style directory separators. Could be cleaned up. Looks like either separator works.

Brett208:
Calling it a day. Need to figure out where appveyor is building test.exe. Maybe different from my local compilation or I'm doing something wrong.

.appveyor.yml
DanRStevens: In this case it makes sense to use "\" since these are shell commands, and so need to use the Windows specific "\". The other paths are not shell commands, nor are they used to run shell commands. They are likely fed directly to the kernel through the filesystem API, and so will accept either directory separator.

test/test.vcxproj
```diff
@@ -9,13 +9,21 @@
...
+    </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
-    <ProjectGuid>{78f02848-bacb-4ad3-985e-0eb0d5b3dd53}</ProjectGuid>
+    <ProjectGuid>{127f1d28-73f4-4710-baa8-8e57a8c07dd4}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
     <ConfigurationType>Application</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v141</PlatformToolset>
```
DanRStevens: Consider using `$(DefaultPlatformToolset)` here. This was discovered by @cugone.

test/test.vcxproj
```diff
@@ -32,6 +32,14 @@
...
   <ImportGroup Label="PropertySheets" />
   <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
+    <IntDir>$(Platform)\$(Configuration)\</IntDir>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
+    <OutDir>$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
+    <IntDir>$(Platform)\$(Configuration)\</IntDir>
+  </PropertyGroup>
```
DanRStevens: Any idea why this was added? These look like they might be default values.

I think the NAS2D project file may have inconsistent settings here between configurations. May be worth checking the NAS2D project file for configuration problems here, and if it relates to any configuration problems in the test project.

test/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-static" version="1.8.1" targetFramework="native" />
+  <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn" version="1.8.0" targetFramework="native" />
```
DanRStevens: For reference, the `rt-static`/`rt-dyn` NuGet package to use should (I believe) match the `RuntimeLibrary` setting in the project file. The `RuntimeLibrary` setting in the test project should match the `RuntimeLibrary` setting in the NAS2D project.

Basically, all compiled modules which are linked together (gtest, test project, NAS2D) should either all link to the static version of the runtime, or they should all link to the dynamic version of the runtime.

If you link to the dynamic version, then the library will have a dependency on the MSVCRT runtime DLL. If you link to the static version, the compiled output may be slightly larger, but will be standalone.

I have a preference for static linking. When it comes time for deployment, end users are never forced to install MSVCRT redistributable packages. In the larger picture, it also avoids the Windows DLL Hell problem. I've heard that Google uses static linking for everything, which I believe stems from that.

We may want to check with @ldicker83 on this one, as it has implications for downstream projects and distribution. Namely OPHD.

test/test.vcxproj
```diff
@@ -29,27 +37,44 @@
...
+    <Import Project="..\proj\vs2017\packages\Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn.1.8.0\build\native\Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn.targets" Condition="Exists('..\proj\vs2017\packages\Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn.1.8.0\build\native\Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn.targets')" />
   </ImportGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>../include/NAS2D;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
-      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
```
DanRStevens: Tagging this for reference with the above discussion.

test/test.vcxproj
```diff
@@ -58,12 +83,28 @@
...
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <ClCompile>
-      <AdditionalIncludeDirectories>../include/NAS2D;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>WIN32;NDEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <RuntimeLibrary>MultiThreadedDLL</RuntimeLibrary>
+      <WarningLevel>Level3</WarningLevel>
+      <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <DisableSpecificWarnings>4996</DisableSpecificWarnings>
+      <AdditionalIncludeDirectories>$(MSBuildThisFileDirectory)include;..\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
DanRStevens: I'm curious about this addition here. Looks like we may need to clean up the include folders. There should be a single include folder to the NAS2D include directory.

test/test.vcxproj
DanRStevens: For running the test executable, the path to the executable will be impacted by this setting.

In particular, the line:
`%APPVEYOR_BUILD_FOLDER%\%PLATFORM%\%CONFIGURATION%\test.exe`

The above settings are trying to place the executable in a subfolder from the solution directory. I assume that's the folder with the solution file, and not the repository root. The repository root will be `%APPVEYOR_BUILD_FOLDER%\`.

Brett208: Before clearing these settings, compare OP2Utility's test project settings. By default x86 builds will use directories `Debug` and `Release`. x64 builds will use `x64\Debug` and `x64\Release` directories. 

We had to standardize by telling the project to use a hardcoded `x86` directory when building. Otherwise it was difficult to line up the location of the test executable between the x86 and x64 builds. 

I was trying to use $(Platform). For x86 builds it produces `win32` for the directory. Unfortunately, appveyor produces `x86`, hence our hardcode of x86 in OP2Utility. 

From Outpost2UtilityTest project settings:
OP2UtilityTestProjectSettings

For the output directory, instead of using $(SolutionDir) we could use $(ProjectDir), which might look a bit cleaner.

test/packages.config
Brett208: Agreed on static linking being a little easier to integrate against outside of Windows environment. I vaguely remember we have run into trouble on missions not being compatible on Wine when not statically linked.

I would be careful about saying slightly larger builds for static. I noticed op2ext was reset from dynamic to static linking since the last release (no sure exactly why or when). Dynamically, it compilied around 65 KB with an associated PDB file of about 1,620KB. Statically, it now compiles at 577KB with an associated PDB file of about 9,748kb. That is closing on a factor of 10 increase. Fortunately both files zip reasonably well.

For NAS2D, I switched from static to dynamic for the test project to match the 2017 project being dynamic already in addition to being Microsoft's recommendation. Sorry I wasn't clear about that in my PR comments. Happy to switch to whatever the project file is updated to.

test/test.vcxproj
Brett208: My bust, will resolve tonight. Forgot to remove the 'include' directory.

test/test.vcxproj
Brett208: Hmm, using this property sets the test project to build using Visual Studio 2010's toolset. Not sure how it chooses what the default should be. I like the idea of it choosing for you based on the other projects in the solution though.

.appveyor.yml
Brett208: Guess I'm good with whatever, just seemed strange.

Brett208:
Took about 11 minutes to run all appveyor tests. I enabled both release and debug configurations to check them. May want to disable the debug configuration just to reduce compilation time. I noticed a large number of warnings appear in the x86 debug build, but didn't investigate at all.

Also, there appears to be a gmock/gmock.h include available in nuget gtest 1.8.1. We should check that out, maybe in OP2Utility to see if we have gmock now. Version 1.8.1 was released about 21 days ago. I would like it if there was some way to subscribe to nuget package releases that we care about.

@DanRStevens, feel free to make whatever changes you want. (I think you have permission to do so). If you want me to tease anything out on Windows let me know. 

-Brett

test/test.vcxproj
DanRStevens: Ok, yes, I remember that now. Thank you for the reminder.

In terms of the "Win32", I think that's because the `Platform` setting in the project file is literally "Win32". We can probably change that to "x86", as it's just an identifying text string.

I suspect newly created projects will use "x86", but because the project was upgraded from an older version of Visual Studio, it's still using the old default of "Win32".

test/packages.config
DanRStevens: Good to point out the inconsistency between the main project file and the test project file.

You're right, the inflation in code size can seem dramatic when you count it proportionally. I suppose to be fair, you'd need to compare the size increase with the size of the CRT itself. (Though it may actually still be out of proportion with the size of the CRT). Still, the benefits may outweigh the size increase, which in overall numbers, doesn't seem to be a huge concern for users these days. In particular, users of older versions of Windows, or Wine, shouldn't need to install a CRT redistributable package.

Not sure about the Microsoft recommendation part. Is that in an error/warning message, or posted online somewhere? I'm guessing from their perspective, they're probably assuming an up-to-date version of Windows.

----

I suppose we could offer multiple configurations. Not sure if we want to go down that path though.

test/test.vcxproj
cugone: @DanRStevens 
@Brett208

I found this thread on the AppVeyor support pages while looking into testing a personal project:
https://help.appveyor.com/discussions/problems/5282-tests-not-being-run#comment_40766291 

Apparently the `Win32` vs `x86` definition of the `Platform` variable is related to native vs managed projects. To fix it you can re-define the platform variable to the correct `Win32` value in the `.appveyor.yml` file:

> `before_test:`
> `  - cmd: if %Platform%=x86 set Platform=Win32`

and still keep the actual project property page as `$(Platform)`

Brett208: @cugone, Looks like a nice alternative.

test/packages.config
Brett208: @DanRStevens,

See image below for where recommendation to use dynamic comes from. I'm happy with static or dynamic.

google test setup dialog

Brett208:
Thanks for the comments. If anyone prefers, we can switch to setting the platform to win32 in appveyor instead of the project. I'm happy either way.

cugone:
11 minutes to run 4 tests...or am I missing what tests are actually being run?

Brett208:
@cugone,

Actually it took 13+ minutes to compile 4 times and run unit tests 4 times.

-Brett

cugone:
I took a look at the logs and have narrowed down where the time is coming from:
 - Each platform compiles the code *twice*: x86 and x64 regardless of platform actually chosen. (1 min each)
 - Re-installing physfs (1 min)
 - Each configuration is run twice, Debug x86/x64 and Release x86/x64 (3 minutes each)
 - TOTAL: ~12 minutes

The tests themselves only take 3 ms. Not bad. (I have several hundred tests running in my engine and it takes only 300 ms)

But since this is just a PR to get tests integrated, changing the appveyor config file should probably be a separate thing after this is pushed.

Brett208:
1 min 47 seconds total with the debug removed.

DanRStevens:
I'm very glad to see this working. Thanks for the efforts.

Now that things have been figured out (CI config is usually a pain), it may be good to clean things up a bit to make it easy for Leeor to review and merge. If you want to spare Leeor from reading the wall of text we've generated with the back and forth in this PR, it might be helpful to move things to a fresh PR. Be sure to request a review from Leeor once it's ready.

Brett208:
Trying to recreate this in a new branch with cleaner commits. It turning into a slog at best.

Brett208:
Closing. Replaced by PR #151 which just improves commit ordering.

PR #141, opened 2019-04-21
Add BuildConfig

cugone:
Added a build config file.

---

This allows for commonly used preproc defines to live in one place.
Examples include:

```
UNUSED:
 - Unused variable does not throw a warning.
 - More explicit and verbose than commenting out
   formal argument names in function implementations.
PLATFORM_LINUX: Linux-specific code
PLATFORM_WINDOWS: Windows-specific code
PLATFORM_APPLE: macOS specific code
TOKEN_STRINGIZE and TOKEN_STRINGIZE_SIMPLE:
 - Easy stringization of nested macros.
 - Useful for __FUNCTION__ and __LINE__ stringize and type names.
TOKEN_PASTE and TOKEN_PASTE_SIMPLE:
 - Concatenation of names and __LINE__ or __FUNCTION__ macros.
SUBSYSTEM_DEBUG:
 - Compiling out subsystem debug code in release builds
   such as AUDIO_DEBUG, RENDER_DEBUG, etc.
 - Compile-time choosing of memory tracking. e.g.
   TRACK_MEMORY/TRACK_MEMORY_VERBOSE/TRACK_MEMORY_BASIC.
   Useful for finding memory leaks in Debug/Release builds
   prior to pushing to Gold or Master build.
```

Right now only the platform-specific defines have been implemented.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -118,7 +118,7 @@ void someFunction(int foo)
 
 ```
 
-Please note that in all cases braces are used including for simple single line if/for/while blocks. While somewhat verbose this is intended to avoid common mistakes mistakes that are easy to overlook. For example:
+Please note that in all cases braces are used including for simple single line if/for/while blocks. While somewhat verbose this is intended to avoid common mistakes that are easy to overlook. For example:
 
 ```cpp
 
--- a/include/NAS2D/BuildConfig.h
+++ b/include/NAS2D/BuildConfig.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#if defined(__clang__)
+#define PLATFORM_CLANG
+#define PLATFORM_LINUX
+#elif defined(__GNUC__) || defined(__GNUG__)
+#define PLATFORM_GNUC
+#define PLATFORM_LINUX
+#endif
+
+#if defined(__APPLE__)
+#undef PLATFORM_LINUX
+#define PLATFORM_APPLE
+#elif defined(_MSC_VER) || defined(_WIN32) || defined(_WIN64)
+#define PLATFORM_WINDOWS
+#endif
--- a/proj/vs2017/NAS2D.vcxproj
+++ b/proj/vs2017/NAS2D.vcxproj
@@ -136,7 +136,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
       <AdditionalIncludeDirectories>../../include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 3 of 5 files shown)


cugone:
@DanRStevens 
@ldicker83 

Requesting review.

cugone:
@DanRStevens 
> 
> Ehh, I don't really get the point. I'm not much of a fan of globally included header files if you can avoid them. Besides, I think we should try to minimize and eliminate the use of platform detects and macros. This just seems to encourage their use.
>

[sarcasm]
> Do you want code duplication? Because that's how you get code duplication.
> --Archer

[/sarcasm]

Seriously though, it's not a "global" header, just like any header you should only include it where you need access to the macros.

The notion of "globals are bad" doesn't really apply in game development. There are so many systems that the game needs access to that trying to pass them around to all the functions that need them just adds more noise to the code. That said, that only applies to the actual game, not the engine and macros are already global by their very nature.

When developing OS-specific systems, say, native handles to threads. The platform defines are helpful in separating out the calls that are different for each OS. Changing the name of a thread is different for each OS: Windows is `::SetThreadDescription`; while linux and macOS is `pthread_setname_np`.

The other macros have important usefulness across the code. `RENDER_DEBUG` can be used to compile out DirectX debug flags when creating a device; `AUDIO_DEBUG` same thing for audio. `UNUSED` is obvious. `TRACK_MEMORY` is helpful to compile out `new/delete` operator overloads when not in the Gold build.

ldicker83:
I'm pretty much against the idea of preprocessor definitions used in this manner (in fact I don't care much for them in most cases). Each platform already offers defines based on compiler and really its only Linux where I see this being a potential issue. In some cases, such as visual, studio, the project file provides the windows defines (e. G., _WINDOWS and WIN32) by adding them to the environment namespace. Apple does something similar via xcode.

I don't really see the benefit of setting up platform defines in this manner. Consistency perhaps?

cugone:
@ldicker83 

True, I hate the preprocessor (and macros, oh god, macros) as well, but there are cases where it is a necessary evil:
 - `if-constexpr` does not work at file- or class-scope (probably never will) for conditionally compiling out entire implementations. The preprocessor does.
 - Profiling a block of code using RAII requires generating a unique class name. Appending `__LINE__` (and `__FUNCTION__` to any messages) to the class name serves this purpose but can only be done using token pasting and stringization with preprocessor defines and macros.
 - Providing 32-bit and 64-bit implementations of a struct then defining a macro that chooses one based on the compiled version. Similar to Windows' unicode vs ASCII functions: `MessageBoxA` and  `MessageBoxW`, then a defined `MessageBox` that is picked based on if the `UNICODE` preprocessor define exists.

DanRStevens:
Correct that `if constexpr` can't be used at the top-level.

----

We're not currently using much in the way of platform specific code. All that stuff has been pushed down into the libraries. A quick grep of `#ifdef` and `#ifndef` shows all uses are in third party code. (Some of it was copy/pasted into the repo, such as the XML parser and the Delegate code, most of it is in the NuGet dependencies).

Perhaps there may be a use for platform detects in the future. In particular, if we drop some of the dependencies which are currently handling platform specific concerns (such as PhysFS) or implement new code such as a DirectX renderer. However, at present, there are no such uses, and I'm hesitant to add such code before it would actually be used.

Additionally, if we do use platform detects, I'd rather use the compiler provided macros directly, rather than translating them to new custom names.

I vote close on this one.

ldicker83:
True, but there will be more when dumping PhysFS and GLEW. I think the only dependencies that will remain are the SDL2 libraries. At least that's the ultimate goal.

DanRStevens:
Yes, we may need some platform specific detects after dropping PhysFS. For that I think we should use the compiler provided macros, rather than redefine our own based on the compiler provided macros. The ones we should need are:
```cpp
_WIN32  # Windows 32-bit and 64-bit (yes, this is defined for Win64)
__APPLE__
__linux__
```

I don't recommend using either `__clang__` nor `__GNUC__` as those are compiler detects rather than platform detects. They won't serve the need we have with filesystem code. In particular, we use both GCC and Clang on both Linux and MacOS, so it would create confusion there.

I believe our needs are well met by the built in macros.

PR #142, opened 2019-04-22
Container updates

DanRStevens:
This includes some general Dockerfile cleanup. Much of this is to resolve Codacy issues from recently changed lint rules.


DanRStevens:
Was curious if using an explicit ":latest" tag would eliminate a Codacy warning about not using an explicit tag. Nor surprisingly it still generates a warning, just a different more detailed one.

I'm reverting that one change so it's back to the old code. Curious if Codacy will still mark it as new, or will classify it as pre-existing.

PR #157, opened 2019-04-28
Cache less on MacOS

DanRStevens:
Don't cache the package download directory. Only cache the built package directory. This should help reduce cache size, which may speed up the build due to smaller caches and faster cache restore.

One downside is that adjusting the installed package list means all packages have to be re-downloaded and compiled again. This however should be a relatively rare occurrence.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,6 @@ commands:
           name: Caching brew dependencies
           key: deps-v2-NAS2D-{{ arch }}
           paths:
-            - /Users/distiller/Library/Caches/Homebrew
             - /usr/local/Cellar
       - run: brew link physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew
   build-googletest:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,11 +12,11 @@ commands:
     steps:
       - restore_cache:
           name: Restoring brew dependencies
-          key: deps-v2-NAS2D-{{ arch }}
+          key: deps-v3-NAS2D-{{ arch }}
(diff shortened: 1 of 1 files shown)


DanRStevens:
This is a bit of an experimental branch. I've opened the PR to share the results of my testing.

Originally the following two folders were cached:
```
/Users/distiller/Library/Caches/Homebrew
/usr/local/Cellar
```

I did a bit of reading and found the first one is the cache for downloaded archives, while the second one is the install location for built binary packages. Technically we only need to cache the installed binary packages, not the original files used to generate them.

In this branch I tried removing the download cache from the CircleCI caching.

A few discoveries:
 - All downloaded packages are "bottles" (pre-built binary packages), none are source builds
 - Uncached total download and install time for bottles was about 38 seconds.
 - The cache version bump causes cmake to be added to cache (33MB)
 - The cache ended up being slightly larger (155MB to 161MB) (+cmake, -packageCache)
 - Cache restore and install time is about 21 seconds.
 - MacOS takes about 17 seconds to restore a cache of ~160MB
 - Caches are stored on Amazon's S3 (Simple Storage Service)
 - Linux builds run on AWS, MacOS runs in a separate data center
 - Linux is able to restore caches much faster than MacOS due to data center cache proximity

Comparing build times and cache size, the results appear to be about the same either way. Caching the extra folder doesn't seem to have any noticeable effect on build time. Mind you, the addition of cmake, which at 33MB is by far the largest package, may have hidden improvements from removing the brew package cache from CircleCI caching.

For the above reason, and perhaps code clarity, I'd recommend merging this. Additionally, the cache version bump in this PR caused cmake to be added to the cache, which fixes that earlier omission. Despite other packages being restored from cache, CMake was being continually re-downloading and re-installing by brew since it was added as a dependency until this cache version bump.

PR #158, opened 2019-04-30
Glew to vcpkg

ldicker83:
Branch name is slightly inaccurate, this not only moves GLEW dependencies from nuget to vcpkg but also changes the nuget dependencies from what were set up in the vs2017 project files to more up to date versions of the SDL2 dependencies (SDL2 core, SDL_Image, SDL_Mixer, SDL_ttf).


DanRStevens:
Hey, sweet. Updated SDL2 packages, and it looks like they're all from a consistent packager now too.

Issue #164, opened 2019-05-05
Convert `Mixer` to interface


DanRStevens:
I think we should convert the `Mixer` class to an interface with pure virtual methods. We can then have a Null Mixer derived class which implements those methods with simple do nothing stubs. (We can call it `MixerNull` to be more consistent with `Mixer_SDL`, which I really wish didn't have an underscore in the name).

Using a virtual base class would provide better error checking for forgotten methods in derived types. It would also eliminate the warnings currently being produced on AppVeyor. (I noticed these just after the `.clang-format` file was added):
```
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(49): warning C4100: 'sound': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(98): warning C4100: 'time': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(98): warning C4100: 'loops': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(98): warning C4100: 'music': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(105): warning C4100: 'time': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
C:\projects\nas2d-core\include\NAS2D\Mixer\Mixer.h(142): warning C4100: 'level': unreferenced formal parameter (compiling source file ..\..\src\Mixer\Mixer_SDL.cpp) [C:\projects\nas2d-core\proj\vs2019\NAS2D.vcxproj]
[...]

PR #166, opened 2019-05-05
Convert Mixer to Abstract Data Type

cugone:
Converted Mixer to an Abstract Data Type.
It contains data and non-virtual member functions
so calling it an "interface" is a misnomer.

Derived NullMixer from Mixer ADT

Updated Non-compiling usage of plain Mixer to
NullMixer when audio driver fails to load.

Added NullMixer header and implementation to project files.

### DIFF
--- a/include/NAS2D/Mixer/Mixer.h
+++ b/include/NAS2D/Mixer/Mixer.h
@@ -34,34 +34,34 @@ public:
 	/**
 	 * C'Tor
 	 */
-	Mixer() {}
+	Mixer();
 
 	/**
 	 * D'tor
 	 */
-	virtual ~Mixer() {}
+	virtual ~Mixer() = 0;
 
 	/**
 	 * Plays a sound on the first available sound channel.
 	 *
 	 * \param	sound	A reference to a Sound Resource.
 	 */
-	virtual void playSound(Sound& sound) {}
+	virtual void playSound(Sound& sound);
 
 	/**
 	 * Stops playing all sounds on all channels.
 	 */
-	virtual void stopSound() {}
+	virtual void stopSound();
 
 	/**
 	 * Pauses sound on all channels.
 	 */
-	virtual void pauseSound() {}
+	virtual void pauseSound();
 
 	/**
 	 * Resumes sound on all channels.
 	 */
-	virtual void resumeSound() {}
+	virtual void resumeSound();
 
 	/**
 	 * Starts playing a Music track.
@@ -69,24 +69,24 @@ public:
 	 * \param music	Reference to a Music Resource.
 	 * \param loops	Number of times to repeat the music.
 	 */
-	void playMusic(Music& music, int loops = Mixer::CONTINUOUS) { fadeInMusic(music, loops, 0); }
+	void playMusic(Music& music, int loops = Mixer::CONTINUOUS);
 
 	/**
 	 * Stops all playing music.
 	 */
-	virtual void stopMusic() {}
+	virtual void stopMusic();
 
 	/**
 	 * Pauses the currently playing Music.
 	 */
-	virtual void pauseMusic() {}
+	virtual void pauseMusic();
 
 	/**
 	 * Resumes the currently paused Music.
 	 *
 	 * \note	It is safe to call this function if the music is stopped or already playing.
 	 */
-	virtual void resumeMusic() {}
+	virtual void resumeMusic();
 
 	/**
 	 * Starts a Music track and fades it in to the current Music volume.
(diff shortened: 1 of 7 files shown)


cugone:
Tagging Issue #164

DanRStevens:
Ahh, it seems we were working on this in parallel. I have my solution in PR #167.

cugone:
@DanRStevens  Yeah that's why I brought up Issue #168 to help alleviate potentially overriding each others work.

include/NAS2D/Mixer/Mixer.h
```diff
@@ -95,92 +95,89 @@ class Mixer
...
 	 *
 	 * This c'tor is not public and can't be invoked externally.
 	 */
-	Mixer(const std::string& name) : mName(name)
-	{}
+	Mixer(const std::string& name);
 
-protected:
-	NAS2D::Signals::Signal0<void>	_music_complete; /**< Callback used when music finished playing. */
+    NAS2D::Signals::Signal0<void>	_music_complete; /**< Callback used when music finished playing. */
```
DanRStevens: Indentation was modified here.

include/NAS2D/Mixer/Mixer.h
```diff
@@ -95,92 +95,89 @@ class Mixer
...
 
 private:
 	// No default copy constructor or copy operator
 	// Calling these any any time should result in an error
 	Mixer(const Mixer&) = delete;
 	Mixer& operator=(const Mixer&) = delete;
 
-private:
-	std::string		mName;		/**< Internal name of the Renderer. */
+    std::string		mName;		/**< Internal name of the Renderer. */
```
DanRStevens: Indentation.

include/NAS2D/Mixer/Mixer.h
```diff
@@ -34,59 +34,59 @@ class Mixer
...
-	virtual ~Mixer() {}
+	virtual ~Mixer() = 0;
 
 	/**
 	 * Plays a sound on the first available sound channel.
 	 *
 	 * \param	sound	A reference to a Sound Resource.
 	 */
-	virtual void playSound(Sound& sound) {}
+	virtual void playSound(Sound& sound);
```
DanRStevens: I think many of these methods are better being pure virtual (`= 0`).

src/Mixer/Mixer.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+	Mixer::Mixer()
+	{
+	}
+
+	Mixer::~Mixer()
+	{
+	}
+
+	void Mixer::playSound(Sound& /*sound*/)
```
DanRStevens: I kind of like how you got around the Warning C4100 here.

src/Mixer/Mixer.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+
+	Mixer::Mixer(const std::string& name)
+	: mName(name)
+	{
+	}
+
+	Mixer::Mixer()
+	{
+	}
```
DanRStevens: For empty constructors, it may be better to use `= default` rather than `{}`. The empty `{}` makes the class non-trivial, 

References:
https://en.cppreference.com/w/cpp/types/is_trivial
https://en.cppreference.com/w/cpp/language/copy_constructor#Trivial_copy_constructor

src/Mixer/Mixer.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+	void Mixer::playSound(Sound& /*sound*/)
+	{
+	}
+
+	void Mixer::stopSound()
+	{
+	}
```
DanRStevens: If the function was declared pure virtual (`= 0`), no empty function definition would be needed to satisfy the compiler.

src/Mixer/Mixer.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+	void Mixer::fadeOutMusic(int /*time*/ /*= Mixer::DEFAULT_FADE_TIME*/)
+	{
+	}
+
+	bool Mixer::musicPlaying() const
+	{
+		return false;
+	}
```
DanRStevens: This method isn't completely empty, but mostly exists to satisfy the requirement to return a value. As the default isn't likely useful to most derived classes, this would also be better off as pure virtual.

src/Mixer/Mixer.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+	void Mixer::unmute()
+	{
+	}
+
+	void Mixer::stopAllAudio()
+	{
+		stopMusic();
+		stopSound();
+	}
```
DanRStevens: I think I do actually like how this was moved to an implementation file.

src/Mixer/NullMixer.cpp
```diff
@@ -0,0 +1,50 @@
...
+	void NullMixer::unmute()
+	{}
+
+	void NullMixer::soundVolume(int /*level*/)
+	{}
+
+	void NullMixer::musicVolume(int /*level*/)
+	{}
+
+} // namespace NAS2D
```
DanRStevens: No EOL (End Of Line) at EOF (End Of File).

Pretty minor. Mostly just the GitHub visual warning, and if you `cat` the file at the Linux terminal, your next prompt may appear at the end of the last line of the file, rather than on a newline by itself.

There's also some history to it:
https://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning
https://stackoverflow.com/questions/31574837/c-standards-newline-ending-of-source-files

Older C++ standards used to require that newline. In the case of C, it might still be required.

src/Mixer/Mixer.cpp
cugone: It's my go-to solution. It keeps the code self-documenting what with the header file still containing the argument names (and more importantly, by extension, Intellisense) and at the same time gets rid of the warning.

src/Mixer/Mixer.cpp
cugone: Whups, that was an oversight.

DanRStevens:
Ack, in my confused busy state today, I forgot to post here earlier. I'd meant to do that before attempting some work using this branch's code.

I was looking at the changes here, and trying to decide what to do. There are two PRs, both basically doing the same thing, and I wanted a few parts of both.

As the biggest part I liked here was the split of the .h into .h and .cpp files, which is mostly the addition of the .cpp files, I thought it might be easiest to apply that addition over top of the other branch. This resulted in PR #174.

PR #167, opened 2019-05-05
Add null mixer

DanRStevens:
This closes Issue #164.

PR #174, opened 2019-05-06
Cherry pick Mixer changes from cugone's work

DanRStevens:
This copies the separation of Mixer into .h and .cpp files. This also brings back the parameter names in the .h file.

Point taken that the Mixer base type contains some non-virtual member functions, so isn't really a true "interface".

I've edited the work to account for my naming difference, plus some style updates so the code more closely matches existing style (for better or worse). It's possible I've introduced some unintended errors, though I did try to check carefully, and local tests were passing.

Original work is from PR #166, and has been based on top of #167. This is an attempt to (subjectively) merge the best of both.

----

The new .cpp files don't currently have a copyright header. I'm not sure it's appropriate for me to edit that, so I leave it to @cugone or @ldicker83 to decide what to do there.

### DIFF
--- a/include/NAS2D/Mixer/Mixer.h
+++ b/include/NAS2D/Mixer/Mixer.h
@@ -69,7 +69,7 @@ public:
 	 * \param music	Reference to a Music Resource.
 	 * \param loops	Number of times to repeat the music.
 	 */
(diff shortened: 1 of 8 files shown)


cugone:
I am okay with these changes.

PR #176, opened 2019-05-06
Add filesystem unit tests

### LINKED ISSUE #175: Add Unit Tests to cover the Filesystem interface
Unit tests should be added to cover the Filesystem interface. This will help verify planned changes to this component.

Additionally, it may help us to clarify what we really need from the interface to this component. Some methods may be rather specific to the underlying implementation, rather than a truly useful high level abstraction.

After tests are written, we may want a discussion as to what the tests should be.

DanRStevens:
This closes #175.

Note that in the last test added for `mount`, I believe the behavior is buggy. I added the test to describe the current behavior. We should consider updating it to reflect intended behavior. I believe it should allow mounting of any arbitrary additional folder from the filesystem, rather than being restricted to a subfolder of the initial data folder (which is already a subfolder of the executable's folder).

In general, we should start a discussion on what we want from the Filesystem layer.

I'm actually wondering if perhaps it should be split to separate out real filesystem concerns from virtual filesystem concerns. For example, the executable's path/name, or the application write folder for the current platform would be real filesystem concerns. They affect the choice of mount points. Whereas reading and writing withing the data folders is a virtual filesystem concern.

----

Another thought I had, is the verbose logging should probably be removed, and instead replaced by unit tests or exceptions. Some seem to be debugging related, and just provide textual confirmation of what is happening. These should become unit tests. Some are for feedback to the caller, and are often associated with an error code. These should be converted to exceptions.

### DIFF
--- a/test/Filesystem.test.cpp
+++ b/test/Filesystem.test.cpp
@@ -85,3 +85,19 @@ TEST_F(FilesystemTest, writeReadDeleteExists) {
 	fs.del(testFilename);
 	EXPECT_FALSE(fs.exists(testFilename));
 }
+
+TEST_F(FilesystemTest, isDirectoryMakeDirectory) {
+	const std::string fileName = "file.txt";
+	const std::string folderName = "subfolder/";
+
+	EXPECT_TRUE(fs.exists(fileName));
+	EXPECT_FALSE(fs.isDirectory(fileName));
+
+	fs.makeDirectory(folderName);
+	EXPECT_TRUE(fs.exists(folderName));
+	EXPECT_TRUE(fs.isDirectory(folderName));
+
+	fs.del(folderName);
+	EXPECT_FALSE(fs.exists(folderName));
+	EXPECT_FALSE(fs.isDirectory(folderName));
+}
--- a/test/Filesystem.test.cpp
+++ b/test/Filesystem.test.cpp
@@ -48,7 +48,31 @@ TEST_F(FilesystemTest, directoryList) {
 	EXPECT_THAT(pathList, Contains(testing::StrEq("file.txt")));
 }
 
+TEST_F(FilesystemTest, exists) {
+	EXPECT_TRUE(fs.exists("file.txt"));
+}
+
 TEST_F(FilesystemTest, open) {
 	const auto file = fs.open("file.txt");
 	EXPECT_EQ("Test data\n", file.bytes());
+}
+
+// Test a few related methods. Some don't test well standalone.
+TEST_F(FilesystemTest, writeReadDeleteExists) {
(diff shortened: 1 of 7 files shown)


DanRStevens:
The Windows builds failed because the NuGet package for Google Test don't include Google Mock. I used GMock Matchers in some of the tests.

Perhaps a more complete GTest package can be installed with `vcpkg` instead of NuGet.

I'll try to address the Windows CI environment issue in another branch, and get back to this branch once that is fixed. Meanwhile, review of the test code is always welcome.

test/Filesystem.test.cpp
```diff
@@ -28,6 +28,15 @@ TEST_F(FilesystemTest, dataPath) {
 	EXPECT_EQ("data/", fs.dataPath());
 }
 
+TEST_F(FilesystemTest, extension) {
+	EXPECT_EQ("txt", fs.extension("subdir/file.txt"));
+	EXPECT_EQ("txt", fs.extension("file.txt"));
+	EXPECT_EQ("reallyLongExtensionName", fs.extension("file.reallyLongExtensionName"));
+	EXPECT_EQ("a", fs.extension("file.a"));
+	EXPECT_EQ("", fs.extension("file."));
+	EXPECT_EQ("file", fs.extension(".file"));
```
cugone: You aren't testing for paths that do not contain a dot or have no extension.

The extension function should really include the dot in the return value to distinguish between files that end in a dot (`"file."` returns `"."`) and files that do not have an extension at all (`"file"` returns `""`.)

DanRStevens: Including the "." as part of the extension is exactly the kind of thing we need to discuss about what we want going forward.

Good point about testing a path with no ".".

I skipped over testing of directory names with an extension and a trailing "/", because they don't work. As in, the following would fail:
```cpp
EXPECT_EQ("ext", fs.extension("DirName.ext/"));  // Actually returns "ext/"
```

cugone: > 
> I skipped over testing of directory names with an extension and a trailing "/", because they don't work. As in, the following would fail:
> 
> ```c++
> EXPECT_EQ("ext", fs.extension("DirName.ext/"));  // Actually returns "ext/"
> ```

This goes back to my suggestion of parroting the standard. The first thing the standard does is check against the special paths `dot` and `dot dot` (`.`, `..`). If the paths are equal to either of those two, it returns and empty string. Otherwise is calls `filename()` which has its own rules to extract the filename, including if there is a trailing slash, and that is what `extension()` operates on. `filename()` may very well return an empty string and if it does, `extension()` is done and it just returns that.

Issue #184, opened 2019-05-13
gmock/gtest not found


cugone:
Pulled latest on master and when building gmock and gtest headers can't be found.

DanRStevens:
I switched to install Google Test (and Google Mock) with `vcpkg` instead of NuGet. You'd have to run the `vcpkg` install commands locally yourself. It's not as integrated as NuGet which will auto restore/update packages.

The `vcpkg` install commands from .appveyor.yml are:
```
vcpkg install gtest:x86-windows
vcpkg install gtest:x64-windows
```

cugone:
That fixed it.

cugone:
@DanRStevens 
@ldicker83 

Recently I've noticed that when I try to build the Test project it fails with the error `gtest.lib not found`.

I've found a fix for this but I'm not sure if it will affect AppVeyor.

In the properties page of the test project: `Linker > Input > Additional Dependencies` instead of listing `gtest.lib;gtest_main.lib` as local to the project adding the relative paths as `$(vcpkgroot)/lib/gtest.lib;$(vcpkgroot)/lib/manual-link/gtest_main.lib` allows the project to build.

However when run in Debug no tests are run, giving the following output (See below).

It's slightly better in Release. Although, the main project doesn't build the tests at least run but half the `FilesystemTest` tests fail. For the main project there is a conflict with the Code Analysis Rule set in use. It is a hard-coded path that does not exist locally. Disabling Code Analysis fixes the issue but I don't know if you want to do this.

For the tests that fail it seems to be because the initialization step of the `FilesystemTest` object sets the working directory to `%APPDATA%\Lairworks\NAS2DUnitTests` (it creates those folders) but the tests themselves expect the working directory to be local to the project so they never find the `data/` folder they are looking for.

Any thoughts?

This is the output when trying to run the tests in Debug:

> [5/23/2019 4:49:24 PM Informational] ------ Discover test started ------
[5/23/2019 4:49:25 PM Informational] ========== Discover test finished: 0 found (0:00:01.3580879) ==========
[5/23/2019 4:49:40 PM Informational] ------ Discover test started ------
[5/23/2019 4:49:40 PM Informational] ========== Discover test finished: 0 found (0:00:00.4495447) ==========
[5/23/2019 4:56:28 PM Informational] ------ Discover test started ------
[5/23/2019 4:56:28 PM Informational] ========== Discover test finished: 0 found (0:00:00.5338749) ==========
[...]

PR #185, opened 2019-05-13
Apply string util fixes

### LINKED ISSUE #182: Add Dan's fixes to string functions
See title. Tagging #180 .

cugone:
Applies changes recommended by @DanRStevens in PR #180 . I pulled latest on master before starting and for some reason gtest and gmock were not found when trying to build (Issue #184 ). Dunno how that happened.

Closes #182

### DIFF
--- a/include/NAS2D/Common.h
+++ b/include/NAS2D/Common.h
@@ -46,10 +46,10 @@ std::pair<std::string, std::string> splitOnLast(const std::string& str, char del
 std::string join(std::vector<std::string> strs, char delim, bool skip_empty = true);
 std::string join(std::vector<std::string> strs, bool skip_empty = true);
 std::string trimWhitespace(std::string string);
-bool startsWith(const std::string_view& string, const std::string_view& start) noexcept;
-bool endsWith(const std::string_view& string, const std::string_view& end) noexcept;
-bool startsWith(const std::string_view& string, char start) noexcept;
-bool endsWith(const std::string_view& string, char end) noexcept;
+bool startsWith(std::string_view string, std::string_view start) noexcept;
+bool endsWith(std::string_view string, std::string_view end) noexcept;
+bool startsWith(std::string_view string, char start) noexcept;
+bool endsWith(std::string_view string, char end) noexcept;
 
 /**
  * Simple helper function to provide a printf like function.
--- a/include/NAS2D/Common.h
+++ b/include/NAS2D/Common.h
@@ -13,7 +13,6 @@
 
 #include <memory>
 #include <string>
-#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -46,8 +45,8 @@ std::pair<std::string, std::string> splitOnLast(const std::string& str, char del
 std::string join(std::vector<std::string> strs, char delim, bool skip_empty = true);
 std::string join(std::vector<std::string> strs, bool skip_empty = true);
 std::string trimWhitespace(std::string string);
-bool startsWith(const std::string_view& string, const std::string& start) noexcept;
-bool endsWith(const std::string_view& string, const std::string& end) noexcept;
+bool startsWith(const std::string& string, const std::string& start);
+bool endsWith(const std::string& string, const std::string& end);
 bool startsWith(const std::string& string, char start);
 bool endsWith(const std::string& string, char end);
 
--- a/include/NAS2D/Common.h
+++ b/include/NAS2D/Common.h
@@ -13,6 +13,7 @@
 
 #include <memory>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


src/Common.cpp
```diff
@@ -177,13 +177,27 @@ std::vector<std::string> NAS2D::split(std::string str, char delim /*= ','*/, boo
 std::pair<std::string, std::string> NAS2D::splitOnFirst(const std::string& str, char delim)
 {
 	const auto delim_loc = str.find_first_of(delim);
-	return std::make_pair(str.substr(0, delim_loc), str.substr(delim_loc + 1));
+	if (delim_loc == std::string::npos)
+	{
+		return std::make_pair(std::string{}, str);
```
DanRStevens: I feel like this may be backwards. If there is no delimiter, and so nothing to split on, I would expect the full string in the first return value, while the second value was empty.

src/Common.cpp
```diff
@@ -177,13 +177,27 @@ std::vector<std::string> NAS2D::split(std::string str, char delim /*= ','*/, boo
...
+	}
 }
 
 std::pair<std::string, std::string> NAS2D::splitOnLast(const std::string& str, char delim)
 {
 	const auto delim_loc = str.find_last_of(delim);
-	return std::make_pair(str.substr(0, delim_loc), str.substr(delim_loc + 1));
+	if (delim_loc == std::string::npos)
+	{
+		return std::make_pair(str, std::string{});
```
DanRStevens: This one also feels reversed to my expectations.

include/NAS2D/Common.h
```diff
@@ -45,8 +46,8 @@ std::pair<std::string, std::string> splitOnLast(const std::string& str, char del
 std::string join(std::vector<std::string> strs, char delim, bool skip_empty = true);
 std::string join(std::vector<std::string> strs, bool skip_empty = true);
 std::string trimWhitespace(std::string string);
-bool startsWith(const std::string& string, const std::string& start) noexcept;
-bool endsWith(const std::string& string, const std::string& end) noexcept;
+bool startsWith(const std::string_view& string, const std::string& start) noexcept;
+bool endsWith(const std::string_view& string, const std::string& end) noexcept;
```
DanRStevens: Ahh, excellent, I like this. That's true, this was added in C++17, which we're now using.

cugone: Funny thing, when looking this up, I found that in C++20 there's going to be a `starts_with` and `ends_with` added to the standard library (that *only* operates on `string_view`s). So this is just a temporary work around until that gets implemented by at least MSVC. 

The string versions internally convert to a string_view:
https://en.cppreference.com/w/cpp/string/basic_string/starts_with
https://en.cppreference.com/w/cpp/string/basic_string/ends_with
https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with
https://en.cppreference.com/w/cpp/string/basic_string_view/ends_with

src/Common.cpp
cugone: I intended it to be "wherever the rest of the string is". In this case "the rest of the string" is the whole string. Maybe a comment or documentation?

src/Common.cpp
cugone: Now I'm confused. You said the previous one felt backwards, yet this is the reverse of the previous and you're saying it's backwards too.

cugone:
@DanRStevens 
@ldicker83 

Re-added string_view versions, both VS2017 and VS2019 support them.

include/NAS2D/Common.h
DanRStevens: A `string_view` should encompass read only access to all of the common string types. If an algorithm is designed to work with a `string_view`, it should work for any of the string types, be they `std::string`, or null terminated `char*` arrays. It's basically a `const char*` and a `size` field tied together. Very cheap to construct or pass by-value.

src/Common.cpp
DanRStevens:
```cpp
splitOnFirst(",abc", ',');  // { "", "abc" }
splitOnFirst("a,bc", ',');  // { "a", "bc" }
splitOnFirst("ab,c", ',');  // { "ab", "c" }
splitOnFirst("abc,", ',');  // { "abc", "" }
splitOnFirst("abc", ',');  // { "abc", "" }  <- Suggestion (rather than { "", "abc" })
```
If the first component is all the stuff before the delimiter, and there is no delimiter, I feel the first component should then be the whole string.

src/Common.cpp
DanRStevens:
```cpp
splitOnLast("abc,", ',');  // { "abc", "" }
splitOnLast("ab,c", ',');  // { "ab", "c" }
splitOnLast("a,bc", ',');  // { "a", "bc" }
splitOnLast(",abc", ',');  // { "", "abc" }
splitOnLast("abc", ',');  // { "", "abc" }  <- Suggestion (rather than { "abc", "" })
```

If the second component is all the stuff after the delimiter (scanning backwards), and there is no delimiter, I feel the second component should then be the whole string.

DanRStevens:
Looks like the CI error was fixed on master. Changes look good.

Maybe give Leeor a chance to respond.

cugone:
Yeah it's my intent to wait for approval from both of you. This change isn't a tiny one-liner like fixing the test project settings was.

PR #197, opened 2019-05-20
Update `stringstream` uses

DanRStevens:
This PR adds unit tests, and then updates `stringstream` code to use the more targeted classes `ostringstream` and `istringstream`.

----

@cugone, I believe there is a bug in `split` as illustrated by the commented out unit test line. I've opened Issue #198 to track this.

### DIFF
--- a/test/Common.test.cpp
+++ b/test/Common.test.cpp
@@ -1,0 +1,8 @@
+#include "NAS2D/Common.h"
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+
+TEST(Version, versionString) {
+	EXPECT_THAT(NAS2D::versionString(), testing::MatchesRegex("[[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+"));
+}
--- a/test/Common.test.cpp
+++ b/test/Common.test.cpp
@@ -6,3 +6,26 @@
 TEST(Version, versionString) {
 	EXPECT_THAT(NAS2D::versionString(), testing::MatchesRegex("[[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+"));
 }
+
+TEST(String, split) {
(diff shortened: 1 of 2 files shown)


ldicker83:
Looks good to me. I never considered the 'i' and 'o' versions of string stream for ensuring correct usage.

test/Common.test.cpp
```diff
@@ -0,0 +1,8 @@
+#include "NAS2D/Common.h"
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+
+TEST(Version, versionString) {
+	EXPECT_THAT(NAS2D::versionString(), testing::MatchesRegex("[[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+"));
```
cugone: Using raw-string-literals will make the regex more clear and remove the need to escape the slashes: `R"([[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+)"`

DanRStevens: Ahh yes, I tried [raw string literals](https://en.cppreference.com/w/cpp/language/string_literal), but was stumbling with the syntax. I didn't think to try enclosing the string with parenthesis while using an empty delimiter.

What you posted works. The empty delimiter has me wondering though, what happens if someone inserts parenthesis into the regex for a capture group. Could that end the raw string literal early? I suppose it might also require a quote. It's not too clear to me how the new rules interact.

cugone: As long as the sequence `)"` does not exist within the string it will work. In cases where it does you can add a delimiter you know won't appear:

    std::string testA = R"(Hello World)";
    std::string testB = R"("C(A|B)")";  <-- this fails and ends early.
    std::string testC = R"delim("C(A|B)")delim";  <-- this works because the delimiter is now )delim"

PR #203, opened 2019-05-21
Cleanup MixerSDL

DanRStevens:
Changes extracted from PR #167.


ldicker83:
To explain the use of the virtual keyword in derived classes, it was intended as a case of documentation. At the time that these classes were originally designed and built, that was the common recommendation. The recent changes to C++ (C++11, C++14 and C++17) make this an obsolete practice with keywords that are far better at their purpose of documentation.

Much appreciation for the efforts involved in getting the code base cleaned up and modernized.

PR #213, opened 2019-05-22
Replace boolean parameters with seperate functions

### LINKED ISSUE #199: Replace boolean parameters with extra string methods
The following methods have a boolean flag parameter:
```cpp
std::vector<std::string> split(std::string str, char delim = ',', bool skip_empty = true);
std::string join(std::vector<std::string> strs, char delim, bool skip_empty = true);
std::string join(std::vector<std::string> strs, bool skip_empty = true);
```

The problem with `bool` parameters, is they are not very self documenting at the point of call. In particular, for a call such as the following, there's no way for someone to infer what the boolean does without reading additional documentation, or finding the method definition:
```cpp
auto result = split(string, ',', false);
```

Instead of using boolean flag parameters, two versions of the methods can be provided. For instance, there might be a `split`, and `splitStripEmpty`, or `join`, and `joinSkipEmpty`. This would provide much better documentation at the point of call about what is intended.

As a side benefit, it also allows lifting the boolean checks out of the inner loops of the above methods, and effectively make them static decisions at the point of call. That should result in more optimized code.

cugone:
Closes #199 .

Added splitSkipEmpty;
Added joinSkipEmpty;
Removed boolean versions of split and join;

### DIFF
--- a/test/Common.test.cpp
+++ b/test/Common.test.cpp
@@ -21,7 +21,7 @@ TEST(String, split)
 	EXPECT_EQ((NAS2D::StringList{"", "abc"}), NAS2D::split(".abc", '.'));
 	EXPECT_EQ((NAS2D::StringList{"a", "bc"}), NAS2D::split("a.bc", '.'));
 	EXPECT_EQ((NAS2D::StringList{"ab", "c"}), NAS2D::split("ab.c", '.'));
-	EXPECT_EQ((NAS2D::StringList{"abc"}), NAS2D::split("abc.", '.'));
+	EXPECT_EQ((NAS2D::StringList{"abc", ""}), NAS2D::split("abc.", '.'));
 }
 
 TEST(String, splitSkipEmpty)
--- a/test/Common.test.cpp
+++ b/test/Common.test.cpp
@@ -4,7 +4,11 @@
 
 
 TEST(Version, versionString) {
+#if GTEST_USES_POSIX_RE == 1
 	EXPECT_THAT(NAS2D::versionString(), testing::MatchesRegex(R"([0-9]+\.[0-9]+\.[0-9]+)"));
+#elif GTEST_USES_SIMPLE_RE == 1
+	EXPECT_THAT(NAS2D::versionString(), testing::MatchesRegex(R"(\\d+\.\\d+\.\\d+)"));
+#endif
 }
 
 TEST(String, split)
--- a/test/Common.test.cpp
+++ b/test/Common.test.cpp
@@ -8,6 +8,13 @@ TEST(Version, versionString) {
 }
 
 TEST(String, split) {
+	EXPECT_EQ((NAS2D::StringList{"a", "b", "c"}), NAS2D::split("a,b,c", ','));
+	EXPECT_EQ((NAS2D::StringList{"abc"}), NAS2D::split("abc", ','));
+	EXPECT_EQ((NAS2D::StringList{"", "abc"}), NAS2D::split(",abc", ','));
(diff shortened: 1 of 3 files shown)


test/Common.test.cpp
```diff
@@ -22,10 +22,10 @@ TEST(String, split) {
...
+	EXPECT_EQ((NAS2D::StringList{"a", "b", "c"}), NAS2D::split("a,b,c", ','));
 	EXPECT_EQ((NAS2D::StringList{"abc"}), NAS2D::split("abc", ',', false));
-	EXPECT_EQ((NAS2D::StringList{"", "abc"}), NAS2D::split(",abc", ',', false));
-	EXPECT_EQ((NAS2D::StringList{"a", "bc"}), NAS2D::split("a,bc", ',', false));
-	EXPECT_EQ((NAS2D::StringList{"ab", "c"}), NAS2D::split("ab,c", ',', false));
-	// EXPECT_EQ((NAS2D::StringList{"abc", ""}), NAS2D::split("abc,", ',', false));  // Actual: {"abc"}
+	EXPECT_EQ((NAS2D::StringList{"", "abc"}), NAS2D::split(",abc", ','));
+	EXPECT_EQ((NAS2D::StringList{"a", "bc"}), NAS2D::split("a,bc", ','));
+	EXPECT_EQ((NAS2D::StringList{"ab", "c"}), NAS2D::split("ab,c", ','));
+	// EXPECT_EQ((NAS2D::StringList{"abc", ""}), NAS2D::split("abc,", ','));  // Actual: {"abc"}
```
DanRStevens: Looks like the `false` parameter tests were updated correctly for the method splitting.

The default parameter tests should be updated to the `splitSkipEmpty` variant. As a side note, now that the methods have been split, we should probably split the unit tests into `split` and `splitSkipEmpty`.

We should probably swap the order of the unit tests so that `split` tests are done first, followed by `splitSkipEmpty`. This would match the updated order of the methods in the implementation file. Barring any reason not to, we should try to be consistent about order.

----

Currently I'm seeing some unit test failures. I suspect this is related to the default parameter tests not being updated to use `splitSkipEmpty`.

cugone: Yeah I'm still getting "Can not find gtest.lib" on my machine so I can't test the test project locally, it keeps failing to build. :(

DanRStevens: Hmm, I thought that was resolved after the discussion in #184. The project switched to using `vcpkg` to install Google Test. Maybe reopen that Issue if it's still a problem.

cugone: `vcpkg install gtest:x86/x64-windows` fixed it for the main project but not the test project. If I try to run it again it says it's already installed.

DanRStevens: Err, Google Test is only used by the test project, not by the main project. There's something more going on here. We may need to investigate further.

cugone:
I think it is a round-about way of saying "I don't support digit/alpha/alnum specifiers...".

They tend to be toss up if they are supported or not so I generally don't use them. Ranges are universally supported though.

DanRStevens:
I think this is good for merging now.

The regex problem noted here should be tracked in a new issue. It is unrelated and pre-existing.

cugone:
I just looked at the source code for gtest's `testing::MatchesRegex` function. Internally it converts the given string to the class `RE`. The comment above said class claims the following:

    // A simple C++ wrapper for <regex.h>.  It uses the POSIX Extended
    // Regular Expression syntax.
    class GTEST_API_ RE { //...

PR #214, opened 2019-05-22
Add missing Common.test.cpp to VS project

cugone:
Common.test.cpp was not added to the test project when it was created.

### DIFF
--- a/test/test.vcxproj
+++ b/test/test.vcxproj
@@ -47,6 +47,7 @@
     <OutDir>$(ProjectDir)$(Platform)\$(Configuration)\</OutDir>
   </PropertyGroup>
   <ItemGroup>
+    <ClCompile Include="Common.test.cpp" />
     <ClCompile Include="Filesystem.test.cpp" />
     <ClCompile Include="Utility.test.cpp" />
   </ItemGroup>


DanRStevens:
Side note: I noticed the unit tests didn't run for this branch. I believe you were even added as a collaborator, so I'm confused why the tests didn't run. We might need to look into why that happened.

cugone:
It's my understanding that the CIs don't run on newly created PRs. You must create an additional commit for them to trigger.

DanRStevens:
The branch CI builds should run as soon as you push, even before the PR is opened. You should be able to see the branch CI builds in the preview before creating the PR. Once the PR is opened, the PR CI builds should then start too.

I think something funny is going on with the CI builds. Your push and PR should both have triggered CI builds. They did before. Maybe it's a service outage.

Another thought, have you tried working directly from a lariworks repository clone, rather than a fork? I believe Leeor set you as a collaborator, so you should have push access to the main repository now. There would be less need to pull and push between different remotes.

cugone:
I'll try to remember to work from the clone I have. It'll definitely make less work of pushing/pulling multiple times every time there's a merge to master (pull from upstream, push to origin on fork; pull from master on clone.)

One advantage of working from the fork is I don't add clutter to the branch list on the main repo.

DanRStevens:
Ahh, no worries. The local clone is a full repo, independent from the origin. It only adds clutter if you push the branches to the main repo. By default, `git push` will only upload changes from the current branch. I regularly have local branches that are never pushed to GitHub, or are only pushed when they are ready for review.

cugone:
Side note: just check the status of circleCI, it's specifically having issues with GithHub notifications

https://status.circleci.com/

DanRStevens:
That's probably it then. A service outage on GitHub's end for notifications would explain why all CI providers at the same time are not running any builds.

PR #215, opened 2019-05-25
Fix split bug in issue 198

### LINKED ISSUE #198: Bug in `split` method when `skip_empty = false` with a trailing delimiter
There is a bug in `split` as illustrated by the commented out unit test line:
```cpp
	// EXPECT_EQ((NAS2D::StringList{"abc", ""}), NAS2D::split("abc,", ',', false));  // Actual: {"abc"}
```

It seems the `split` method doesn't respect the `bool skip_empty` parameter when it is `false` and there is a trailing delimiter.

For reference:
[`std::getline`](https://en.cppreference.com/w/cpp/string/basic_string/getline)
[`std::basic_ios::operator bool`](https://en.cppreference.com/w/cpp/io/basic_ios/operator_bool)

The code in question:
```cpp
	while (std::getline(ss, curString, delim))
	{
		if (skip_empty && curString.empty()) { continue; }
		result.push_back(curString);
	}
```
When no characters are extracted, `std::getline` sets the `failbit`. The `std::getline` method returns the input stream. The input stream is converted to a `bool` for the `while` test. The `operator bool` on the `stringstream`/`istringstream` returns `!fail()`. Hence, when no data is extracted the `std::getline` effectively returns `false`, even when a delimiter was present.

cugone:
Closes #198
Uncommented broken tests in split.

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -166,6 +166,9 @@ std::vector<std::string> NAS2D::split(std::string str, char delim /*= ','*/)
 	{
 		result.push_back(curString);
 	}
+    if(ss.eof() && str.back() == delim) {
+		result.push_back(std::string{});
+    }
 	result.shrink_to_fit();
 	return result;
 }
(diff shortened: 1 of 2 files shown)


src/Common.cpp
```diff
@@ -166,6 +166,9 @@ std::vector<std::string> NAS2D::split(std::string str, char delim /*= ','*/)
 	{
 		result.push_back(curString);
 	}
+    if(ss.eof() && str.back() == delim) {
+		result.push_back(std::string{});
+    }
```
DanRStevens: Minor point: The indentation here is off. Probably a spaces versus tabs issue.

PR #216, opened 2019-05-25
Removes pure virtual declaration on d'tor's for Mixer and Renderer.h.…

ldicker83:
… Pure virtual Mixer d'tor caused a compiler issue later on for derived types.

### DIFF
--- a/include/NAS2D/Mixer/Mixer.h
+++ b/include/NAS2D/Mixer/Mixer.h
@@ -32,7 +32,7 @@ public:
 
 public:
 	Mixer() = default;
-	virtual ~Mixer() = 0;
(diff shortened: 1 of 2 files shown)


include/NAS2D/Mixer/Mixer.h
```diff
@@ -32,7 +32,7 @@ class Mixer
 
 public:
 	Mixer() = default;
-	virtual ~Mixer() = 0;
+	virtual ~Mixer() = default;
```
DanRStevens: I'm curious, why set one to explicit `default`, and the other left unspecified. I suppose the result is the same. It just feels like there is a difference in intent.

PR #218, opened 2019-05-28
Split common files

cugone:
Splits `Common.h/.cpp` into `StringUtils.h/.cpp` and `MathUtils.h/.cpp`

Adds `#pragma message` warning of deprecation for usages downstream of engine. Engine updates will be added as separate task.


src/Common.cpp
```diff
@@ -5,328 +5,9 @@
...
- * \param	divisor		Divisor.
- *
- * \return	Returns the divided number rounded up to the nearest whole number.
- */
-int NAS2D::divideUp(int to_divide, int divisor)
-{
-	return (to_divide + (divisor - 1)) / divisor;
-}
+#include "MathUtils.cpp"
+#include "StringUtils.cpp"
```
DanRStevens: I don't think this is required in a .cpp file. Preserving the header and having header includes for the split files should be enough to preserve compatibility. It shouldn't be needed at the .cpp file level.

As long as the methods are defined in some .cpp file that is part of the project, the linker should be able to find the object code and link them. Actually, including source files in another source file could potentially cause link errors, as the same object methods are defined in two different compilation units with the same name. Maybe you'll get lucky and have the linker discard the duplicates, assuming they are exactly the same, though best not to rely on that behavior.

cugone: Fixed. 

Yeah that was a brain-fart. I dunno why I did that. :P

cugone:
@DanRStevens 

What do you suggest for the version functions? `Version.h`?

I only left them in `StringUtils.h` because they dealt with "stringifying" the version number.

ldicker83:
I was scratching my head over that one too. I suppose we can leave them in StringUtil for now for the above stated reason. We can find a more sane place for it as needed.

PR #221, opened 2019-05-30
Mark Filesystem class Deprecated PR

### LINKED ISSUE #209: Mark Filesystem class as deprecated
Future versions of NAS2D will drop the `Filesystem` class entirely. See #118 and #179.

When NAS2D was first born (circa 2008/2009), the C++ standard library had no built in way to properly deal with the native filesystem. C++17's `std::filesystem` effectively solves this issue (minus Apple continuing to not provide an actual implementation).

Anyway, because of this, the `NAS2D::Filesystem` object has basically become obsolete. Users of NAS2D should be encouraged to use the C++17 std::filesystem implementation and move away from `Filesystem`.

cugone:
Closes #209 

Adds a `#pragma message` to warn of deprecation.

### DIFF
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#pragma message("Filesystem Class is deprecated and wil be removed in a future release.")
 #include "StringUtils.h"
 #include "File.h"
 #include <string>


ldicker83:
This should be done using the C++11 attribute specifier instead of a pragma. Pragma's aren't reliable on all compilers but attributes are now part of the standard: https://en.cppreference.com/w/cpp/language/attributes

cugone:
@ldicker83 
This was an attempt to at least show the message without breaking the code. Similar to the `Common.h` split.
The break in appveyor was because of changes in the versionString test specific to windows that had not been pulled in. Fixed now.

MSVC breaks against standard and emits an error when using the `[[deprecated]]` attribute instead of a warning due to SDL checks. @DanRStevens  also had some reservations using it because of the option to "treat warnings as errors" setting.

See: https://github.com/lairworks/nas2d-core/issues/209#issuecomment-494859233

DanRStevens:
In terms of review, it looks like this branch is based on another branch, which has yet to be merged to master. It seems this branch introduces a fairly simple one line edit. I don't believe it really depends on the changes from the other branch it is based upon. As such, I'd recommend branching off from master, rather than another branch. That creates a lot less noise for review, and allows changes to be assessed more independently.

If possible, I'd appreciate a rebase.

----

As for the change, interesting strategy to use a `#pragma message` to avoid issuing an actual warning. On Linux, I'm seeing a bit of "note:" spam while compiling the changes to Common.h. Looks like a few internal parts of the library need to be updated for the header file split.

As a side note, do we even need Common.cpp? It doesn't contain any code. It only has a single include, which generates a compile warning / note, but doesn't actually do anything. No library user should ever interact directly with the source file itself, only the header file, so it should be safe to remove the source file.

DanRStevens:
As per my earlier suggestion, I rebased these changes. It was hard to know what this branch really did since it included all of another branch. Plus, it had grown stale and had merge conflicts.

With that said, and as per the discussion in #209, I think we should close this without merging. The current plan is to maintain a `Filesystem` class, but to change its internal implementation from using PhysFS over to use `<filesystem>` plus a bit of custom code.

DanRStevens:
Ok, I'm going to go ahead and close this. As per the discussion above, we have a new plan. Currently the new plan is being worked on in PR #243.

Issue #220, opened 2019-05-28
Remove Common files


cugone:
Remove `Common.h/.cpp` files and their references in projects entirely.

This is a separate task so as to not break downstream projects without significant warning via compiler messages from #218 and #219 .

DanRStevens:
Tagging related:
Issue #181
Issue #196
PR #218
Issue #219
PR #222
PR #229
PR #224

PR #222, opened 2019-05-31
Replace/Remove all uses of Common.h in the Engine

### LINKED ISSUE #196: Split Common files into String and Math files and nested namespaces
Split the common.h and common.cpp files into String and Math files with nested namespaces `NAS2D::String` and `NAS2D::Math`

### LINKED ISSUE #219: Replace Common usages with StringUtils and/or MathUtils
Engine-side of  #218 . Downstream projects will need to update separately.

Replace all usages of `Common.h` with `StringUtils.h` and/or `MathUtils.h`

cugone:
Closes #219 
Closes #196

### DIFF
--- a/include/NAS2D/Configuration.h
+++ b/include/NAS2D/Configuration.h
@@ -9,7 +9,6 @@
 // ==================================================================================
 #pragma once
 
-#include "Common.h"
 #include "Exception.h"
 
 #include <map>
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -10,7 +10,7 @@
 
 #pragma once
 
-#include "Common.h"
+#include "NAS2D/StringUtils.h"
 #include "File.h"
 #include <string>
 
--- a/include/NAS2D/NAS2D.h
+++ b/include/NAS2D/NAS2D.h
@@ -8,13 +8,15 @@
 // = Acknowledgement of your use of NAS2D is appriciated but is not required.
 // ==================================================================================
 
-#include "NAS2D/Common.h"
 #include "NAS2D/Configuration.h"
 #include "NAS2D/EventHandler.h"
 #include "NAS2D/Exception.h"
 #include "NAS2D/Filesystem.h"
 #include "NAS2D/FpsCounter.h"
 #include "NAS2D/Game.h"
+#include "NAS2D/StringUtils.h"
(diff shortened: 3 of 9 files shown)


DanRStevens:
As I just noted in PR https://github.com/lairworks/nas2d-core/pull/221#issuecomment-498156045, we may want to simply remove "Common.cpp". It doesn't need to exist for backwards compatibility, and may simply cause diagnostic output by including the now deprecated header file.

cugone:
> 
> 
> As I just noted in PR [#221 (comment)](https://github.com/lairworks/nas2d-core/pull/221#issuecomment-498156045), we may want to simply remove "Common.cpp". It doesn't need to exist for backwards compatibility, and may simply cause diagnostic output by including the now deprecated header file.

Removal of the files is a separate issue #220 in order to give downstream projects enough time to update. I'd prefer removing them all in one fell swoop because it requires modifying the project file (`.vcxproj`) itself and splitting that up between two PRs is a headache.

DanRStevens:
Yes, the Common.h file can affect downstream projects, if they've included it directly. The Common.cpp file won't affect downstream projects. When NAS2D is compiled into a library file, all object code is archived in a single .lib file, regardless of where the methods came from. The .lib file doesn't track the source object file. Moving the method implementation from one .cpp file to another .cpp is a distinction that will be lost during the compile/archive step. All the methods in Common.cpp have simply been moved to other .cpp files. It is safe to remove the Common.cpp file now, so long as you preserve the .h file for downstream projects.

That is of course assuming that Common.h was a public header. If it's a private header, and a downstream project has included it directly, that's their fault, and they should be broken. :wink: 

I suppose this brings up the related question of whether or not the other header files are public, or only the main NAS2D.h file is considered public.

Anyway, doesn't matter which PR the removal is done in. I think it's safe to do it now, but you're welcome to do it in a future PR. Either way, this is good to merge now.

DanRStevens:
Reminder this still needs to be merged. This fixes a number of warnings which show up in downstream projects.

----

Additional tasks, which can be completed post merge:
- Split Common.test.cpp
- Remove Common.cpp

DanRStevens:
Ok, so Leeor isn't able to review right now. As this is a pretty innocuous change, and should remove warnings for downstream projects, I'm going to go ahead and merge this.

PR #223, opened 2019-06-01
Add splitOnUnquoted functions

cugone:
Adds split functions that preserve quoted delimiters. Useful for key-value parsers.

### DIFF
--- a/include/NAS2D/StringUtils.h
+++ b/include/NAS2D/StringUtils.h
@@ -24,6 +24,9 @@ namespace NAS2D
 	std::vector<std::string> splitSkipEmpty(std::string str, char delim = ',');
 	std::pair<std::string, std::string> splitOnFirst(const std::string& str, char delim);
 	std::pair<std::string, std::string> splitOnLast(const std::string& str, char delim);
+	std::vector<std::string> splitOnUnquoted(const std::string& str, char delim = ',');
+	std::vector<std::string> splitOnUnquotedSkipEmpty(const std::string& str, char delim = ',');
+
 	std::string join(std::vector<std::string> strs);
 	std::string join(std::vector<std::string> strs, char delim);
 	std::string joinSkipEmpty(std::vector<std::string> strs);
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -131,6 +131,14 @@ std::vector<std::string> NAS2D::splitOnUnquoted(const std::string& str, char del
 		}
 		substring_end = iter++;
 		std::string substring(substring_start, substring_end);
+		if (substring.front() == '"')
+		{
+			substring.erase(0, 1);
+		}
+		if (substring.back() == '"')
+		{
+			substring.pop_back();
+		}
 		result.push_back(substring);
 		substring_start = iter;
 	}
@@ -164,10 +172,19 @@ std::vector<std::string> NAS2D::splitOnUnquotedSkipEmpty(const std::string& str,
 		}
 		substring_end = iter++;
 		std::string substring(substring_start, substring_end);
-		if (substring.empty()) { 
-            substring_start = iter;
-            continue;
-        }
+		if (substring.front() == '"')
+		{
+			substring.erase(0, 1);
+		}
+		if (substring.back() == '"')
+		{
+			substring.pop_back();
+		}
+		if (substring.empty())
+		{
+			substring_start = iter;
+			continue;
+		}
 		result.push_back(substring);
 		substring_start = iter;
 	}
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -110,73 +110,79 @@ std::pair<std::string, std::string> NAS2D::splitOnLast(const std::string& str, c
 
 std::vector<std::string> NAS2D::splitOnUnquoted(const std::string& str, char delim /*= ','*/)
 {
-	bool in_quote = false;
+	bool inQuote = false;
 	std::vector<std::string> result{};
-	auto potential_count = 1u + std::count(std::begin(str), std::end(str), delim);
+	std::size_t potential_count = 1u + std::count(std::begin(str), std::end(str), delim);
 	result.reserve(potential_count);
-	auto substring_start = std::begin(str);
-	auto substring_end = std::end(str);
+	auto start = std::begin(str);
+	auto end = std::end(str);
 	for (auto iter = std::begin(str); iter != std::end(str); /* DO NOTHING */)
 	{
 		if (*iter == '"')
 		{
-			in_quote = !in_quote;
+			inQuote = !inQuote;
 			++iter;
 			continue;
 		}
-		if (in_quote || *iter != delim)
+		if (!inQuote)
 		{
-			++iter;
-			continue;
+			if (*iter == delim)
+			{
+				end = iter++;
+				std::string s(start, end);
+				result.push_back(std::string(start, end));
+				start = iter;
+				continue;
+			}
 		}
-		substring_end = iter++;
-		std::string substring(substring_start, substring_end);
-		result.push_back(substring);
-		substring_start = iter;
+		++iter;
 	}
-	substring_end = std::end(str);
-	auto last_string = std::string(substring_start, substring_end);
-	result.push_back(last_string);
+	end = std::end(str);
+	auto last_s = std::string(start, end);
+	result.push_back(std::string(start, end));
 	result.shrink_to_fit();
 	return result;
 }
 
 std::vector<std::string> NAS2D::splitOnUnquotedSkipEmpty(const std::string& str, char delim /*= ','*/)
 {
-	bool in_quote = false;
+	bool inQuote = false;
 	std::vector<std::string> result{};
-	auto potential_count = 1u + std::count(std::begin(str), std::end(str), delim);
+	std::size_t potential_count = 1u + std::count(std::begin(str), std::end(str), delim);
 	result.reserve(potential_count);
-	auto substring_start = std::begin(str);
-	auto substring_end = std::end(str);
+	auto start = std::begin(str);
+	auto end = std::end(str);
 	for (auto iter = std::begin(str); iter != std::end(str); /* DO NOTHING */)
 	{
 		if (*iter == '"')
 		{
-			in_quote = !in_quote;
+			inQuote = !inQuote;
 			++iter;
 			continue;
 		}
-		if (in_quote || *iter != delim)
+		if (!inQuote)
 		{
-			++iter;
-			continue;
+			if (*iter == delim)
+			{
+				end = iter++;
+				std::string s(start, end);
+				if (s.empty())
+				{
+					start = iter;
+					continue;
+				}
+				result.push_back(std::string(start, end));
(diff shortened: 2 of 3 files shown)


include/NAS2D/StringUtils.h
```diff
@@ -24,6 +24,9 @@ namespace NAS2D
 	std::vector<std::string> splitSkipEmpty(std::string str, char delim = ',');
 	std::pair<std::string, std::string> splitOnFirst(const std::string& str, char delim);
 	std::pair<std::string, std::string> splitOnLast(const std::string& str, char delim);
+	std::vector<std::string> splitOnUnquoted(const std::string& str, char delim = ',');
+	std::vector<std::string> splitOnUnquotedSkipEmpty(const std::string& str, char delim = ',');
```
DanRStevens: The `splitOnUnquotedSkipEmpty` method strikes me as almost never desirable. An empty quoted string is often used to explicitly specify a string value which is empty. Discarding such strings seems to go against the very reason for specifying them.

src/StringUtils.cpp
```diff
@@ -108,6 +108,76 @@ std::pair<std::string, std::string> NAS2D::splitOnLast(const std::string& str, c
...
+	auto substring_end = std::end(str);
+	for (auto iter = std::begin(str); iter != std::end(str); /* DO NOTHING */)
+	{
+		if (*iter == '"')
+		{
+			in_quote = !in_quote;
+			++iter;
+			continue;
+		}
+		if (in_quote || *iter != delim)
```
DanRStevens: I'm a little uncertain about this suggestion, but....

Here there is a stateful `in_quote` variable, and the nested `if` statements in a loop. Often the state variable can be removed by splitting the loop into two. I suspect that can be done here. One loop to skip over ignored stuff or an opening quote, and another loop to find the end of the current string (close quote, delimiter, end of string). Or perhaps even just two calls to [`std::find`](https://en.cppreference.com/w/cpp/algorithm/find) instead of two loops. Admittedly it's complicated by both delimiters and quotes, and you'd still need an overall loop here to handle multiple strings for the split result.

test/Common.test.cpp
```diff
@@ -56,6 +56,10 @@ TEST(String, splitOnUnquoted)
 	EXPECT_EQ((NAS2D::StringList{"", "abc", "def"}), NAS2D::splitOnUnquoted("\"\".abc.def", '.'));
 	EXPECT_EQ((NAS2D::StringList{"abc", "", "def"}), NAS2D::splitOnUnquoted("abc.\"\".def", '.'));
 	EXPECT_EQ((NAS2D::StringList{"abc", "def", ""}), NAS2D::splitOnUnquoted("abc.def.\"\"", '.'));
+	EXPECT_EQ((NAS2D::StringList{"Hello", "Hello World", "World"}), NAS2D::splitOnUnquoted("Hello \"Hello World\" World", ' '));
+	EXPECT_EQ((NAS2D::StringList{"Hello", " ", "World"}), NAS2D::splitOnUnquoted("Hello \" \" World", ' '));
+	EXPECT_EQ((NAS2D::StringList{"Hello", "", "World"}), NAS2D::splitOnUnquoted("Hello \"\" World", ' '));
```
DanRStevens: What about quotes in the middle of a string? I assume those are ignored?
```
EXPECT_EQ((NAS2D::StringList{"abc"}), NAS2D::splitOnUnquoted("a\"b\"c")); // Or broken into {"a", "b", "c"}?
```

----

I suppose you've actually tested this, though it wasn't immediately obvious. The parser is stateful. Does it handle arbitrary mixes of quoted and unquoted strings? A specific case I don't see is two quoted strings separated by an unquoted string.
```
EXPECT_EQ((NAS2D::StringList{"a", "b", "c"}), NAS2D::splitOnUnquoted("\"a\",b,\"c\""));
```

cugone: It's a simple on/off state like block quotes in C++ `/*...*/` attempting to nest either will make you sad.

I wrote this one specifically for a `KeyValueParser`.

I wrote one for an `ArgumentParser` that uses a different algorithm. It uses a stream and checks the first character of an extracted word to determine if it's in a quote and checks the last character of an extracted word to determine when it's not anymore. I did it that way because the argument parser does not preserve the quotes like the key value parser does.

include/NAS2D/StringUtils.h
cugone: It's useful for parsing files you know to have blank lines, like config files: `auto lines = splitOnUnquotedSkipEmpty(entireFileStr, '\n');`. This will output all the non-empty lines in the file.

src/StringUtils.cpp
cugone: I'm not quite understanding what you are suggesting. Do you have an example?

include/NAS2D/StringUtils.h
DanRStevens: That's a pretty good example. That should be included in the documentation, and perhaps in a test case.

src/StringUtils.cpp
DanRStevens: I'm afraid I don't have an example, and would need to defer on finding one. I haven't played around with the details enough to know if the particular concerns here (multiple delimiters: quotes, specified delimiter, end of string) would get in the way of using higher-level methods. If not, it might be possible to use shorter higher-level code. Probably not a high priority item though, so feel free to ignore.

test/Common.test.cpp
DanRStevens: A simple on/off switch is a simple and sensible way to handle things.

From the test cases I read through, it looked like quotes were stripped. I suspect it would be more useful to strip the quotes than to preserve them. I'm wondering if I misunderstood what you're saying about preserving quotes. Or perhaps that was a later change I didn't check in detail.

DanRStevens:
One additional thought I had. What happens (or should happen) if the user specifies a quote as the delimiter?

cugone:
> 
> 
> One additional thought I had. What happens (or should happen) if the user specifies a quote as the delimiter?

My brain says it will return the entire input string. Let's check:

e.g.

    Input:
        string: Hello "Hello World" World
        delimiter: "
    Algorithm:
    quote := false
    substring_start := beginning of input string
    substring_end := end of input string
    for each character in input string:
        H: not a quote, !quote  and not a delimiter, go to next char.
        e: not a quote, !quote and not a delimiter, go to next char.
        l: not a quote, !quote and not a delimiter, go to next char.
        l: not a quote, !quote and not a delimiter, go to next char.
        o: not a quote, !quote and not a delimiter, go to next char.
        [space]: not a quote, !quote and not a delimiter, go to next char.
        ": is a quote, set quote, go to next char.
        H: not a quote, quote is true but not a delimiter, go to next char.
        e: not a quote, quote is true but not a delimiter, go to next char.
        l: not a quote, quote is true but not a delimiter, go to next char.
        l: not a quote, quote is true but not a delimiter, go to next char.
        o: not a quote, quote is true but not a delimiter, go to next char.
        [space]: not a quote, quote is true but not a delimiter, go to next char.
        W: not a quote, quote is true but not a delimiter, go to next char.
        o: not a quote, quote is true but not a delimiter, go to next char.
        r: not a quote, quote is true but not a delimiter, go to next char.
        l: not a quote, quote is true but not a delimiter, go to next char.
        d: not a quote, quote is true but not a delimiter, go to next char.
        ": is a quote, clear quote, go to next char.
        [space]: not a quote, !quote and not a delimiter, go to next char.
        W: not a quote, !quote but not a delimiter, go to next char.
        o: not a quote, !quote but not a delimiter, go to next char.
        r: not a quote, !quote but not a delimiter, go to next char.
        l: not a quote, !quote but not a delimiter, go to next char.
        d: not a quote, !quote but not a delimiter, go to next char.


by now the algorithm has reached the end of the for loop, `substring_end` is set to the end of the input string (which it already is) and a substring is created from `substring_start` to `substring_end` and pushed into the result vector. Since `substring_start` and `substring_end` A) default to the beginning and end of the input string and B) never change: the entire input string is returned. QED.

Long story short: Don't do that.

DanRStevens:
Heh, nice results presentation.

Might be good to document somewhere that quotes have higher parse precedence than delimiters, and so using a quote at the delimiter means the string won't be split. Or maybe that's a valid error condition, and the method could throw if the delimiter is set to a quote character.

cugone:
> 
> 
> Curious, where would this be used? I was thinking command line arguments, but then those are generally already parsed by the shell before the program even sees them.

On windows the `WinMain` entry point does not split commandline arguments into an array like `argc, argv` of `main` does. You get it all as one string: `cmdLine`.

As noted above, this can be used in a KeyValueParser or even an IniParser (`.ini`...yuck) to grab all the non-empty lines by passing a newline in as the delimiter: `auto lines = splitOnUnquotes(entireFileStr, '\n');`

DanRStevens:
I'm a bit confused by the use of this, as in why is it being added. It could be used for processing `WinMain` arguments, though `WinMain` is platform specific, and there's no clear reason to use it over the standard C/C++ `main` method, which has the arguments already parsed. It could be used for .ini file parsing, though we don't currently use .ini files, and it sounds like there is a preference to avoid them.

I'm not sure what use case or feature needs to have this added.

cugone:
> 
> 
> I'm a bit confused by the use of this, as in why is it being added. It could be used for processing `WinMain` arguments, though `WinMain` is platform specific, and there's no clear reason to use it over the standard C/C++ `main` method, which has the arguments already parsed. It could be used for .ini file parsing, though we don't currently use .ini files, and it sounds like there is a preference to avoid them.
> 
> I'm not sure what use case or feature needs to have this added.

We already talked about this. Data-driven design; Key-value-pair-based files; etc

DanRStevens:
I think we've missed each other's point here. The context given is very generic. Sure, the method can be used in those contexts, but do those contexts matter to the NAS2D library? What's the roadmap of feature design? What are the priorities for the library. I'm not sure how this fits in to the grand scheme of development.

cugone:
> 
> 
> I think we've missed each other's point here. The context given is very generic. Sure, the method can be used in those contexts, but do those contexts matter to the NAS2D library? What's the roadmap of feature design? What are the priorities for the library. I'm not sure how this fits in to the grand scheme of development.

@ldicker83 

Granted, a roadmap for future features and design would be helpful. :P

DanRStevens:
I'm looking back at this old PR. The reason for it is not entirely clear, which leaves me tempted to close without merging. To be fair though, the code didn't seem like it was bad code, just that it wasn't clear why specifically it was being added. On top of that, the presentation of the PR had a lot of bug fixing, which makes it hard to review.

If we were to merge this, I think we need a clear and compelling specific reason as to where and how the new code will actually be used, and the code would need to be rebased onto the current master to eliminate the merge conflicts, and to clean up the presentation so reviewers don't need to read through bug fixes.

I'd be hesitant to ask for the cleanup though without first knowing the why.

cugone:
I vote to close it. It's not an immediate concern right now and if later a project needs the feature the topic can be revisited.

DanRStevens:
Ok, closing. I'll leave the branch for a couple of days, in case you want to copy the code anywhere.

PR #224, opened 2019-06-03
Split common.test.cpp

cugone:
Splits Common.test.cpp into separate files

After Merge of #222 the current failing build will pass due to the changes not being available to master yet.


test/MathUtils.test.cpp
```diff
@@ -0,0 +1,237 @@
+#include "NAS2D/MathUtils.h"
+#include "NAS2D/Renderer/Primitives.h"
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+TEST(MathUtils, divideUp)
+{
+	EXPECT_DEATH(NAS2D::divideUp(0, 0), "divideUp(0,0) divide by zero.");
```
DanRStevens: This looks like it could be an [`EXPECT_THROW`](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#exception-assertions) rather than an `EXPECT_DEATH`.

I'm wondering if the divide by 0 cases should be separated into their own test case, or perhaps segregated from the other test lines. I think I would have done that due to the different behavior in that case.

test/MathUtils.test.cpp
```diff
@@ -0,0 +1,237 @@
...
+	EXPECT_EQ(26, NAS2D::divideUp(256, 10));
+	EXPECT_EQ(24, NAS2D::divideUp(256, 11));
+	EXPECT_EQ(22, NAS2D::divideUp(256, 12));
+	EXPECT_EQ(20, NAS2D::divideUp(256, 13));
+}
+
+TEST(MathUtils, isPointInRectComponent)
+{
+	//Neghbors
+	EXPECT_EQ(false, NAS2D::isPointInRect(4, 4, 5, 5, 1, 1));
```
DanRStevens: I think this test really illustrates how hard the current method is to read at the point of call. There are 6 numbers, but it's not clear which ones represent the point and which ones represent the rect.

To be clear, this isn't a problem with the test case, but rather the test case highlighting a weakness of the design. It would be nice to update the method parameters to make for easier to read test code, which generally serves as good example code for how things will actually be used.

test/MathUtils.test.cpp
```diff
@@ -0,0 +1,237 @@
...
+	EXPECT_EQ(true, NAS2D::isPointInRect(6, 6, 5, 5, 2, 2));
+	EXPECT_EQ(true, NAS2D::isPointInRect(7, 6, 5, 5, 2, 2));
+	EXPECT_EQ(false, NAS2D::isPointInRect(8, 6, 5, 5, 2, 2));
+}
+
+TEST(MathUtils, isPointInRectClass)
+{
+	using namespace NAS2D;
+	//Neghbors
+	EXPECT_EQ(false, NAS2D::isPointInRect({4, 4}, {5, 5, 1, 1}));
```
DanRStevens: This set of tests is much easier to read than the one above. Even though individual values are passed here, the `{}` adds minimal syntactic noise, and yet makes it much easier to read and understand the code at the point of call.

As the test here shows what is possible, I think I'd recommend dropping the other variant of the method that allows passing unpacked values.

From an assembly level perspective, I don't believe the unpacked values would have any efficiency advantage over the packed values. Particularly not if the structs are passed by value, as it would all just be pushed onto the stack either way.

test/MathUtils.test.cpp
```diff
@@ -0,0 +1,237 @@
...
+	EXPECT_EQ(true, NAS2D::isPointInRect({5, 6}, {5, 5, 2, 2}));
+	EXPECT_EQ(true, NAS2D::isPointInRect({6, 6}, {5, 5, 2, 2}));
+	EXPECT_EQ(true, NAS2D::isPointInRect({7, 6}, {5, 5, 2, 2}));
+	EXPECT_EQ(false, NAS2D::isPointInRect({8, 6}, {5, 5, 2, 2}));
+}
+
+TEST(MathUtils, isRectInRectComponent)
+{
+	//Neghbors
+	EXPECT_EQ(false, NAS2D::isRectInRect(3, 3, 1, 1, 5, 5, 1, 1));
```
DanRStevens: Again, this test shows the unpacked parameter value variant of the method is much harder to read at the point of call. I'm thinking this variant should be dropped.

test/MathUtils.test.cpp
DanRStevens: I think using explicit class names makes it even more clear. Should be the programmer's choice though. Maybe we should provide examples of both syntax in the documentation.

test/MathUtils.test.cpp
DanRStevens: I just realized, this is technically incorrect. For integers, division by 0 is undefined. We can't even rely on the code dying. It's just extremely common for the OS to throw a system level exception that causes the code to die.

We should probably update the code to detect this case and throw an exception, and then update the test code to check for an exception. That way we would have defined behaviour.

test/MathUtils.test.cpp
DanRStevens: I think I'm starting to agree more with explicit constructors. At the very least, I think example code should use the type names when using initializer lists.

DanRStevens:
I tried fixing some of the failing code/tests. I left the `isRectInRect` methods for now. I figure we should have a discussion about how they should work.

For `isRectInRect`, we should consider renaming the method if what we are really testing is overlap rather than containment. For overlap, we should decide if the test should be end-point inclusive or exclusive. There is often a good case to be made for using endpoint exclusive logic. That is assuming touching at a point or line is considered overlap. Maybe we want area of overlap, rather than simply touching.

DanRStevens:
I noticed something a bit odd. The two overloads of `isRectInRect` take different parameters. The unpack version takes `(x1, x2, y1, y2)` sets, where as the packed version takes `(x, y, width, height)` sets.

Regardless, I think the method is going to need to be renamed at some point. It checks for overlap, rather than containment. I checked OPHD and OP2-Landlord, neither of which have direct references to `isRectInRect`, so assuming no other projects use it, this would be an internal rename.

Actually, if we made it a little easier to obtain structs and work with them, we could probably remove the unpacked overload.

cugone:
Maybe `DoRectsOverlap`? Removing the unpacked versions and using only the packed versions would also remove ambiguity, there's only one way to use the function and only one way to construct Rectangle_2ds: `DoRectsOverlap(const Rectangle_2d& a, const Rectangle_2d& b)`

I'd only ever see a reason to use the unpacked versions for interfacing with C functions or GPU APIs.

Same and similar with Point_2d.

DanRStevens:
Let's defer the rename to a new issue. (#235).

I'd say this PR closes #220 (in case it's not considered closed already).

Alright. Tests are passing. This change is long overdue. Let's merge this.

PR #227, opened 2019-06-04
Build up math library

cugone:
The beginnings of a proper math library.

Implements:
 - Vector2
 - Vector3
 - Vector4
 - AABB2
 - convertDegreeToRadians in MathUtils
 - convertRadiansToDegrees in MathUtils
 - Adds Constants namespace to MathUtils
 - Adds the Constants
   - pi
   - 2pi
   - pi/2
   - pi/4
   - 1/pi
   - 2/pi
   - 2/sqrt(pi)

The `Constants` namespace exists because the ones in `cmath` are not standard and may not be available (VS requires a compile flag to even turn them on). VS2019 even moved them to their own file in preparation for modules support in C++20/2y. Putting them behind a namespace prevents collisions with downstream projects defining their own. 

I haven't implemented `Matrix4` yet because the implementation depends on a discussion on memory layout, column/row-major order, named member functions vs overloading the `operator*`, and DirectX projection matricies vs OpenGL projection matricies (or both, probably).

Also because it's big. 140 functions in my implementation and over 1000 LOC.

### DIFF
--- a/include/NAS2D/Math/AABB2.h
+++ b/include/NAS2D/Math/AABB2.h
@@ -1,0 +1,55 @@
+// ==================================================================================
+// = NAS2D
+// = Copyright © 2008 - 2019 New Age Software
+// ==================================================================================
+// = NAS2D is distributed under the terms of the zlib license. You are free to copy,
+// = modify and distribute the software under the terms of the zlib license.
+// =
+// = Acknowledgment of your use of NAS2D is appreciated but is not required.
+// ==================================================================================
+#pragma once
+
+#include "NAS2D/Math/Vector2.h"
+
+namespace NAS2D
+{
+
+	class AABB2
+	{
+	  public:
+		Vector2 mins{};
+		Vector2 maxs{};
+
+		static const AABB2 Zero_to_One;
+		static const AABB2 Neg_One_to_One;
+
+		AABB2() = default;
+		AABB2(const AABB2& other) = default;
+		AABB2(AABB2&& other) = default;
+		AABB2& operator=(const AABB2& other) = default;
+		AABB2& operator=(AABB2&& other) = default;
+		~AABB2() = default;
+
+		AABB2(float x, float y) noexcept;
+		AABB2(float minX, float minY, float maxX, float maxY) noexcept;
+		AABB2(const Vector2& mins, const Vector2& maxs) noexcept;
+		AABB2(const Vector2& center, float radiusX, float radiusY) noexcept;
+
+		void stretchToIncludePoint(const Vector2& point) noexcept;
+		void addPaddingToSides(float paddingX, float paddingY) noexcept;
+		void addPaddingToSidesClamped(float paddingX, float paddingY) noexcept;
+		void translate(const Vector2& translation) noexcept;
+
+		Vector2 calcCenter() const noexcept;
+		Vector2 calcDimensions() const noexcept;
+
+		AABB2 operator+(const Vector2& translation) const noexcept;
+		AABB2 operator-(const Vector2& antiTranslation) const noexcept;
+		AABB2& operator+=(const Vector2& translation) noexcept;
+		AABB2& operator-=(const Vector2& antiTranslation) noexcept;
+
+	  protected:
+	  private:
+	};
+
+} // namespace NAS2D
--- a/include/NAS2D/Math/Vector2.h
+++ b/include/NAS2D/Math/Vector2.h
@@ -30,7 +30,7 @@ namespace NAS2D
 		Vector2& operator=(Vector2&& r_other) = default;
 		~Vector2() = default;
 		Vector2(float x, float y) noexcept;
-        //TODO: Construct from std::string
+		//TODO: Construct from std::string
 
 		Vector2 operator+(const Vector2& rhs) const noexcept;
 		Vector2& operator+=(const Vector2& rhs) noexcept;
@@ -80,7 +80,7 @@ namespace NAS2D
 
 		friend void swap(Vector2& a, Vector2& b) noexcept;
 
-        //TODO: streaming operations
+		//TODO: streaming operations
 
 		float x = 0.0f;
 		float y = 0.0f;
--- a/include/NAS2D/Math/Vector2.h
+++ b/include/NAS2D/Math/Vector2.h
@@ -1,0 +1,92 @@
+// ==================================================================================
+// = NAS2D
+// = Copyright © 2008 - 2019 New Age Software
+// ==================================================================================
+// = NAS2D is distributed under the terms of the zlib license. You are free to copy,
+// = modify and distribute the software under the terms of the zlib license.
+// =
+// = Acknowledgment of your use of NAS2D is appreciated but is not required.
+// ==================================================================================
+
+#pragma once
+
+namespace NAS2D
+{
+
+	class Vector2
+	{
+	  public:
+		static const Vector2 Zero;
+		static const Vector2 X_Axis;
+		static const Vector2 Y_Axis;
+		static const Vector2 One;
+		static const Vector2 XY_Axis;
+		static const Vector2 YX_Axis;
+
+		Vector2() = default;
+		Vector2(const Vector2& rhs) = default;
+		Vector2(Vector2&& rhs) = default;
+		Vector2& operator=(const Vector2& other) = default;
+		Vector2& operator=(Vector2&& r_other) = default;
+		~Vector2() = default;
+		Vector2(float x, float y) noexcept;
+        //TODO: Construct from std::string
+
+		Vector2 operator+(const Vector2& rhs) const noexcept;
+		Vector2& operator+=(const Vector2& rhs) noexcept;
+
+		Vector2 operator-() const noexcept;
+		Vector2 operator-(const Vector2& rhs) const noexcept;
+		Vector2& operator-=(const Vector2& rhs) noexcept;
+
+		friend Vector2 operator*(float lhs_scalar, const Vector2& rhs) noexcept;
+		Vector2 operator*(float rhs_scalar) const noexcept;
+		Vector2& operator*=(float rhs_scalar) noexcept;
+		Vector2 operator*(const Vector2& rhs) const noexcept;
+		Vector2& operator*=(const Vector2& rhs) noexcept;
+
+		Vector2 operator/(float rhs_scalar) const noexcept;
+		Vector2& operator/=(float rhs_scalar) noexcept;
+		Vector2 operator/(const Vector2& rhs) const noexcept;
+		Vector2& operator/=(const Vector2& rhs) noexcept;
+
+		bool operator==(const Vector2& rhs) const noexcept;
+		bool operator!=(const Vector2& rhs) const noexcept;
+
+		void getXY(float& x, float& y) const noexcept;
+		float* getAsFloatArray() noexcept;
+
+		float calcHeadingRadians() const noexcept;
+		float calcHeadingDegrees() const noexcept;
+		float calcLength() const noexcept;
+		float calcLengthSquared() const noexcept;
+
+		void setXY(float newX, float newY) noexcept;
+		void setHeadingRadians(float headingRadians) noexcept;
+		void setHeadingDegrees(float headingDegrees) noexcept;
+		void setUnitLengthAndHeadingRadians(float headingRadians) noexcept;
+		void setUnitLengthAndHeadingDegrees(float headingDegrees) noexcept;
+		float setLength(float length) noexcept;
+		void setLengthAndHeadingRadians(float headingRadians, float length) noexcept;
+		void setLengthAndHeadingDegrees(float headingDegrees, float length) noexcept;
+
+		float normalize() noexcept;
+		Vector2 getNormalize() const noexcept;
+
+		void rotate90Degrees() noexcept;
+		void rotateNegative90Degrees() noexcept;
+		void rotateRadians(float radiansToAdd) noexcept;
+		void rotateDegrees(float degreesToAdd) noexcept;
+
+		friend void swap(Vector2& a, Vector2& b) noexcept;
+
+        //TODO: streaming operations
+
+		float x = 0.0f;
+		float y = 0.0f;
+
+	  protected:
+	  private:
+	};
+
+} // namespace NAS2D
--- a/include/NAS2D/Math/Vector2.h
+++ b/include/NAS2D/Math/Vector2.h
@@ -87,6 +87,7 @@ namespace NAS2D
 
 	  protected:
 	  private:
+		float calcInverseLength() const noexcept;
 	};
 
 } // namespace NAS2D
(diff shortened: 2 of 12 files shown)


cugone:
The Codacy/PR QR is throwing false positives for assigned, but unused variables and I'm just not seeing why.

DanRStevens:
I'm not sure what features this addition will support. I'm hesitant to spend time reviewing this without knowing why it's being added.

----

It looks like Vector2/3/4 are largely declared the same. I'm thinking the common aspects should be templated / pulled out into a base class to reduce the repetition.

----

Using `auto` more might protect against internal changes, such as replacing `float` with `double` or `long double`.

----

The Codacy warnings don't make sense to me either. Have you tried doing the division inline? I suppose you're trying to optimize away a slow division. What about providing a `calcInverseLength()` method, (which returns `0.0f` when the length is zero).

cugone:
> 
> 
> I'm not sure what features this addition will support. I'm hesitant to spend time reviewing this without knowing why it's being added.
> 
Are you confused what features a math library adds to a *Game Engine*?! Here let me show you:
 - Physics
 - Lighting
 - Collision Detection
 - Collision Response
 - Artificial Intelligence
 - Trigonometry
 - Linear Algebra
 - Vector Algebra
 - Vector Calculus
 - Matrix Transformations
 - Linear Transformations
 - Nonlinear Transformations
 - Easing functions
 - Noise
 - Perlin Noise
 - Cameras
 - Frustum culling
 - Bounds checking
 - Rendering
 - Ray tracing
 - Picking
 - Spatial partitioning
   - Quad trees
   - Oct Trees
  - Red/Black trees
 - Vertex manipulation
 - GPU programming
 - Physically-based Rendering
 - Hidden surface removal
 - Texture manipulation
 - 3D Audio
 - Separating Axis Theorem
 - Range Mapping
 - Interpolation
 - Intrinsic equations
 - Parametric equations
 - Terrain generation
 - Map generation
 - Voroni regions
 - Distance
   - Euclidian
   - Manhattan
 - Vector fields
 - Path finding
   - A*
   - Dijkstra's algorithm

And that's what I could think of just off the top of my head.

> It looks like Vector2/3/4 are largely declared the same. I'm thinking the common aspects should be templated / pulled out into a base class to reduce the repetition.
> 
They are different classes because they have different uses. For example, `Vector4` is the only one that's allowed to interface with `Matrix4` by design. That design being "because math says so". `Vector3` is the only one that has a concept of computing a Cross Product (to be added to MathUtils). You can't set the length and heading of a `Vector3` because there are infinite possibilities. Each one is implicitly constructable from the next highest version, but not one lower because of ambiguities in what the value of the undefined component should be (especially for `Vector3` to `Vector4`, that causes all kinda of problems). Making something derived or templated "because you can" isn't always a good idea. A classic is the academic example of `Shape` derived classes or worse, `Rectangle` derived from `Square` or vice versa. These are bad examples and should never be taught.

> Edit: Using [`std::array`](https://en.cppreference.com/w/cpp/container/array) may be appropriate here.
> 
Using an array hinders readability and comprehension for only 2/3/4 values. Which is more understandable at a glance?: `direction.x` or `direction[0]` it's also less typing.

That said, for `Matrix4` using a `std::array<float, 16> m_indices` is a given. Provided the use of functions to abstract away the gross (meaning "many" not "undesirable") use of `m_indicies` in some algorithms.

> Using `auto` more might protect against internal changes, such as replacing `float` with `double` or `long double`.

When someone wants to make a game with world units at sub-millimeter scale or larger than 340,282,340,000,000,000,000,000,000,000,000,000,000 I'll think about changing it from `float`. Right now, speed, space, and interfacing with shader code were my concerns.

> The Codacy warnings don't make sense to me either. Have you tried doing the division inline? I suppose you're trying to optimize away a slow division. What about providing a `calcInverseLength()` method, (which returns `0.0f` when the length is zero).
> 
> Edit: I wonder if it might be related to optimization. As calculations are often done internally using higher precision math than `float`, perhaps the `float` variable is actually never used, and instead some internal unnamed temporary value of higher precision is being used?
I'll try that. Depending how it's written, the function call should be compiled out.

DanRStevens:
Using `std::array` to implement a templated Vector class allows generic code to be written that works for any dimension with a single method implementation. The Vector2/3/4 classes could potentially be reduced to 1 line `using` declarations. Reducing 3 classes to 1 substantially improves maintainability.

----

Probably not quite what you meant, but there is no runtime overhead in terms of speed or space for using `auto`. The actual type is determined at compile time. By not having to be explicit about the type, it often allows the same method to work even if the underlying data type changes. This can be particularly useful in templates where a method may be instantiated for multiple types, and a local variable needs a type that matches a template argument. There is no runtime performance cost to using `auto`.

cugone:
> 
> 
> Using `std::array` to implement a templated Vector class allows generic code to be written that works for any dimension with a single method implementation. The Vector2/3/4 classes could potentially be reduced to 1 line `using` declarations. Reducing 3 classes to 1 substantially improves maintainability.
> 

In terms of video games, creating Vectors or Matricies with dimensions larger than 4 is not needed. Due to the nature of Matrix multiplication and composition, creating Matrix2 and Matrix3 classes is also not needed. Therefore, creating a template class to service creating large vectors and matrix dimensions is not needed.

There exist operations on a Vector2 class that cannot be performed on a Vector3 or Vector4; there exist operations on a Vector3 class that cannot be performed on a Vector2 or Vector4; there exist operations on a Vector4 class that cannot be performed on a Vector2 or Vector3.

Let's take Vector2 as an example and see what's left after we remove all functionality that can only be performed on a Vector2:

        class Vector2
        {
        public:
            //
            // Constructing convenience vector3s and vector4s for vector2s is right out.
            //
            static const Vector2 Zero;
            static const Vector2 X_Axis;
            static const Vector2 Y_Axis;
            static const Vector2 One;
            static const Vector2 XY_Axis;
            static const Vector2 YX_Axis;
            
            Vector2() = default;
            Vector2(const Vector2& rhs) = default;
            Vector2(Vector2&& rhs) = default;
            Vector2& operator=(const Vector2& other) = default;
            Vector2& operator=(Vector2&& r_other) = default;
            ~Vector2() = default;

            //
            // This constructor is ambiguous for Vector3 and Vector4
            //
            Vector2(float x, float y) noexcept;
            
            //
            // Arithmetic operations cannot be performed on mismatched sizes
            //
            Vector2 operator+(const Vector2& rhs) const noexcept;
            Vector2& operator+=(const Vector2& rhs) noexcept;
            Vector2 operator-() const noexcept;
            Vector2 operator-(const Vector2& rhs) const noexcept;
            Vector2& operator-=(const Vector2& rhs) noexcept;
            friend Vector2 operator*(float lhs_scalar, const Vector2& rhs) noexcept;
            Vector2 operator*(float rhs_scalar) const noexcept;
            Vector2& operator*=(float rhs_scalar) noexcept;
            Vector2 operator*(const Vector2& rhs) const noexcept;
            Vector2& operator*=(const Vector2& rhs) noexcept;
            Vector2 operator/(float rhs_scalar) const noexcept;
            Vector2& operator/=(float rhs_scalar) noexcept;
            Vector2 operator/(const Vector2& rhs) const noexcept;
            Vector2& operator/=(const Vector2& rhs) noexcept;
            
            //
            // Equality of mismatched sizes is not defined.
            //
            bool operator==(const Vector2& rhs) const noexcept;
            bool operator!=(const Vector2& rhs) const noexcept;
            
            void getXY(float& x, float& y) const noexcept;
            float* getAsFloatArray() noexcept;
            
            //
            // Calculating the direction of a vector3 requires another vector3
            // i.e. (target - source).normalize();
            // Calculating the direction of a vector4 is impossible
            //
            float calcHeadingRadians() const noexcept;
            float calcHeadingDegrees() const noexcept;
            
            //
            // Calculating the length of a Vector4 is different and ambiguous
            // Do you want the length in 3D or 4D?
            //
            float calcLength() const noexcept;
            float calcLengthSquared() const noexcept;
            
            void setXY(float newX, float newY) noexcept;

            //
            // Setting the direction of a vector3 requires another vector3
            // and is non-trivial
            // i.e. vec = (target - source).getNormalize() * vec.calcLength();
            // or vec = (target - source).getNormalize(); for unit length
            //
            void setHeadingRadians(float headingRadians) noexcept;
            void setHeadingDegrees(float headingDegrees) noexcept;
            void setUnitLengthAndHeadingRadians(float headingRadians) noexcept;
            void setUnitLengthAndHeadingDegrees(float headingDegrees) noexcept;

            //
            // Setting the length of a vector3/4 is non-trivial unlike Vector2
            // i.e. vec = vec.getNormalize() * newLength;
            //
            float setLength(float length) noexcept;
            void setLengthAndHeadingRadians(float headingRadians, float length) noexcept;
            void setLengthAndHeadingDegrees(float headingDegrees, float length) noexcept;
            
            float normalize() noexcept;
            Vector2 getNormalize() const noexcept;

            //
            // Rotating a vector3 or vector4 "by 90 degrees" is ambiguous
            // Rotating a vector3 or vector4 without specifying an axis of rotation is undefined.
            //
            void rotate90Degrees() noexcept;
            void rotateNegative90Degrees() noexcept;
            void rotateRadians(float radiansToAdd) noexcept;
            void rotateDegrees(float degreesToAdd) noexcept;
            
            friend void swap(Vector2& a, Vector2& b) noexcept;
            
            float x = 0.0f;
            float y = 0.0f;

	      protected:
	      private:
	    };

    } // namespace NAS2D

Let's see what we are left with:

        class Vector2
        {
        public:
            
            Vector2() = default;
            Vector2(const Vector2& rhs) = default;
            Vector2(Vector2&& rhs) = default;
            Vector2& operator=(const Vector2& other) = default;
            Vector2& operator=(Vector2&& r_other) = default;
            ~Vector2() = default;
            
            void getXY(float& x, float& y) const noexcept;
            float* getAsFloatArray() noexcept;
            
            float calcLength() const noexcept;
            float calcLengthSquared() const noexcept;
            
            void setXY(float newX, float newY) noexcept;
            
            float normalize() noexcept;
            Vector2 getNormalize() const noexcept;
            
            friend void swap(Vector2& a, Vector2& b) noexcept;
            
            float x = 0.0f;
            float y = 0.0f;

	      protected:
	      private:
	    };

    } // namespace NAS2D

Default Constructors, multi-argument getter/setter convenience functions for lower dimensions, calculating the length, normalization, and swapping; all of which are trivial to implement. Congratulations, you've saved 12 lines of code and added unneeded complexity.


> Probably not quite what you meant, but there is no runtime overhead in terms of speed or space for using `auto`. The actual type is determined at compile time. By not having to be explicit about the type, it often allows the same method to work even if the underlying data type changes. This can be particularly useful in templates where a method may be instantiated for multiple types, and a local variable needs a type that matches a template argument. There is no runtime performance cost to using `auto`.

Not what I meant at all. I meant changing the underlying type from `float` to `double` or `long double` is a bad idea. Especially when interfacing with the GPU and shader code.

ldicker83:
Sorry, been down and out for the count (again). Burned out from work.

Long weekend though so I'll take the time over the next couple of days to really read through this and see if there's anything that stands out.

cugone:
> 
> 
> Sorry, been down and out for the count (again). Burned out from work.
> 
> Long weekend though so I'll take the time over the next couple of days to really read through this and see if there's anything that stands out.

@ldicker83 

It's been about a month since last communication. Any update on this PR review (or any of the others...)? Did you manage to look any of it over on your three-day weekend?

DanRStevens:
I'm afraid I've been a little out of things here for the past couple of months. I'm just starting to get back to things now. I may need a bit of time to regain context on this one.

I did have a few comments in regards to using templates here. Some method vary a bit between the different vector dimensions, but they vary in very formulaic ways, hence I think templates would still be useful.

In regards to the underlying type of the matrix, it's possible to set a default type with templates, which can be used unless explicitly specified to be otherwise. That means you can set a default type to `float` for performance reasons with graphical hardware, but still allow the library user to specify something else if needed for their particular needs. Vectors are often useful in scientific realms outside of graphics processing, and in some cases the extra precision can be warranted. A parameterized type with default of `float` would make for a good common ground.

I'd like to add some code examples for the above two points, though I may need to come back to this after I've had time to re-read and research some stuff.

I'm also still a little uncertain as to the overall plan concerning this addition. Is there a particular use case where this new feature addition will be needed, which is blocked by the absence? Or are these being added more as a generally useful thing, which you happen to have some code for?

DanRStevens:
I'm taking a look back at this. I'm not really sure what to do with it. It's big. Too big. I kind of feel like we can't properly review and incorporate all of this at once. I would however like to work on some of these methods. I saw a few things I wanted to experiment with, perhaps reducing some of the bulk by using templates.

Would it be possible to keep this code in a separate repository for now? Perhaps we could merge it in the future, or setup a Git submodule to access it as an independent subproject. That might even be a cleaner setup done that way.

If I remember correctly, this code came from another project. I'm not sure of the exact origin, but I suspect https://github.com/cugone/Abrams2019 may be an appropriate upstream? If so, perhaps we can review and work on the code there first, and then once it becomes familiar we can look into incorporating it here.

cugone:
Changing the vector/matrix implementations in https://github.com/cugone/Abrams2019 in any way is a no-go.

I've got dozens of projects that use it and the vector/matrix implementations there are so ingrained into them that any change would require a complete overhaul of hundreds of thousands of lines of code.

That said, you may use it as a reference for specific feature implementations.

The matrix4 implementation is a little unorthodox and requires RtL multiplication to work correctly. That might be somewhere it can be improved.

DanRStevens:
Sounds daunting. Though to be fair, I was hoping to only try out small changes that would be backwards compatible. If you had a hundred thousand lines of code using it, that would make for a good test bed to ensure changes were backwards compatible.

Most of my ideas were about refactoring to reduce code bulk, while maintaining identical functionality.

Would be open to other suggestions though.

DanRStevens:
So any more thoughts on how to proceed here? I still don't know what the roadmap is on this stuff, and it doesn't feel close to merge. I'm tempted to close this without merging, but I don't really want to lose the code either. Is this all contained in the Abrams2019 repository? If so, I think it should be safe to close, as the code will still exist somewhere else.

If you have a specific reason to use the code, we can also consider a Git submodule to access the code from another repository. For an example of how that works, check out the [OPHD](https://github.com/ldicker83/OPHD) or (OP2-Landlord](https://github.com/OutpostUniverse/op2-landlord) projects, which both use NAS2D as a Git submodule. That would allow for a bit more isolation and independent development of the different parts.

cugone:
You can go ahead and close it. There isn't a need for heavy Vector/Matrix math yet and the code can be easily re-implemented later. Maybe when the code is cleaned up enough and I get around to writing a DirectX renderer it can be revisited.

@ldicker83  Please consider https://trello.com/ for organizing a features roadmap. :)

DanRStevens:
Ok closing.

----

And a bit off-topic, but I did finally start using the project board in another repository, and it's actually quite nice. I haven't quite figured out how to set a long term roadmap with it, but for managing lots of short term changes it was helpful. Might be worth a discussion issue, or forum post, or IRC chat.

PR #228, opened 2019-06-06
Fix missing hard-coded path for code analysis ruleset

cugone:
The hard-coded path provided was pointing to an older version of the "Microsoft Native Recommended Ruleset" and if only VS2019 is installed it will not be there and fail.

Changing this to the built-in "Microsoft Recommended Native Ruleset" fixes this issue and accounts for any discrepancy in install locations.

### DIFF
--- a/proj/vs2019/NAS2D.vcxproj
+++ b/proj/vs2019/NAS2D.vcxproj
@@ -100,6 +100,8 @@
     <TargetName>$(ProjectName)_d</TargetName>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <CodeAnalysisRuleSet>NativeRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <RunCodeAnalysis>true</RunCodeAnalysis>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release Gold|x64'">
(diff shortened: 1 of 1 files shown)


cugone:
@ldicker83 
@DanRStevens 

Side note: This fix now allows me to debug compiler and test errors locally because the Release build now works.

Prior to this, the Debug build of Test would fail because it couldn't find the gtest libs, and the Release build of NAS2D would fail because of this issue, but the Release build of Test wouldn't (wtf is up with that?!)

cugone:
@DanRStevens 

Only halfway. Still can't build Debug successfully, which means I can't step through code. Usually creating a "FastDebug" profile (i.e., Release mode with optimizations turned off) fixes it. Will test later.

DanRStevens:
Hmm, I suspect this is related to the error I see in #229.

I noticed this bit of information:
[How to disable code analysis in MSBuild target ClCompile?](https://stackoverflow.com/questions/9438987/how-to-disable-code-analysis-in-msbuild-target-clcompile)

I can't say I entirely understand what is going on here, thought I think I want this change merged.

DanRStevens:
Ok, I looked into this a bit more. I think we rather want to get this change in now. This might be addressing the root cause of some of the other problems I've been seeing. I can't be sure without trying though, and one of the best ways for me to try is to merge and see what the CI builds start doing. Worst case scenario, we revert if I'm wrong about this.

PR #229, opened 2019-12-16
Remove common cpp

DanRStevens:
This is based on PR #222 (and hence commits from that branch will show up here). It adds an additional commit to remove Common.cpp. This file is not needed for backwards compatibility. Only Common.h would be needed for that.

Removing Common.cpp gets rid of deprecation warnings caused by its inclusion of Common.h.


DanRStevens:
Hmm, I've seen this failure on another project in the last few days. It doesn't appear to be related to the particular change introduced here. I'll have to experiment a bit. I read a bit about it in the following thread, which seems to provide a solution:
[MSBuild error MSB4018 with VS2013 while doing analyze](https://social.msdn.microsoft.com/Forums/vstudio/en-US/cc0dfb93-9f18-40bd-9c6f-377808bcd34e/msbuild-error-msb4018-with-vs2013-while-doing-analyze?forum=msbuild)

DanRStevens:
The failure shown here is already on master. It has recently been affecting other projects as well. Oddly it only affects the `Release|Win32` configuration. I suspect this is a bug in the tooling, possibly caused by a Visual Studio/MSVC update on the build servers. It appears to be caused by bad default project settings for the one configuration.

Other configurations show no problems, and similarly other platforms show no problems. As such, I'm going to go ahead and merge this. This is a fairly simple change, and it should reduce warnings, allowing us to focus on the remaining errors and warnings.

PR #238, opened 2019-12-20
Update code analysis rules

DanRStevens:
Noticed a bit of a discrepancy between configurations in the project file.

Would be nice if someone with Visual Studio reviews this.


cugone:
Looks good.

PR #240, opened 2019-12-20
Fix Various warnings

### LINKED ISSUE #237: Fix warnings code pass
VS2019's default settings brings up a lot of easily fixable warnings. Fixing warnings promotes better and more performant code.

Since we are not compiling with warnings as errors, periodically making sure all the warnings are cleaned up should be a regular task.

cugone:
Closes #237

This fixes all the non-third party warnings emitted from x86 and x64 targets emitted by Visual Studio 2019.

I'm not sure how to fix the SDL_rect.h warning. Normally I'd just wrap the include in a `#pragma warning` block to ignore it, but because it's from the nuget package I've yet to find a solution.


src/EventHandler.cpp
```diff
@@ -630,7 +661,7 @@ bool EventHandler::shift(KeyModifier mod) const
  */
 bool EventHandler::alt(KeyModifier mod) const
 {
-	return ((mod & KEY_MOD_LALT) || (mod & KEY_MOD_RALT));
+	return KeyModifier::KEY_MOD_NONE != ((mod & KeyModifier::KEY_MOD_LALT) | (mod & KeyModifier::KEY_MOD_RALT));
```
DanRStevens: Minor point, but this can probably be written as:
```cpp
return KeyModifier::KEY_MOD_NONE != (mod & KEY_MOD_ALT);
```

Edit: Actually, to make it more consistent with the other functions that follow, reverse that order:
```cpp
return (mod & KEY_MOD_ALT) != KeyModifier::KEY_MOD_NONE;
```

src/EventHandler.cpp
```diff
@@ -619,7 +650,7 @@ bool EventHandler::textInputMode()
  */
 bool EventHandler::shift(KeyModifier mod) const
 {
-	return ((mod & KEY_MOD_SHIFT) || (mod & KEY_MOD_CAPS));
+	return KeyModifier::KEY_MOD_NONE != ((mod & KEY_MOD_SHIFT) | (mod & KeyModifier::KEY_MOD_CAPS));
```
DanRStevens: Similarly, this could be written as:
```cpp
return (mod & (KEY_MOD_SHIFT | KeyModifier::KEY_MOD_CAPS)) != KeyModifier::KEY_MOD_NONE;
```

src/Filesystem.cpp
```diff
@@ -128,7 +128,7 @@ StringList Filesystem::searchPath() const
  *
  * \note	This function will also return the names of any directories in a specified search path
  */
-StringList Filesystem::directoryList(const std::string& dir, const std::string& filter) const
+StringList Filesystem::directoryList(const std::string& dir, const std::string& filter /*= std::string{}*/) const
```
DanRStevens: I have a mild preference to not include the comment here. Comments have a tendency to grow stale. Plus I prefer to not repeat information. Repeated information tends to increase maintenance costs, or allows for skew when not all values are updated together.

src/Xml/XmlDocument.cpp
```diff
@@ -29,7 +29,9 @@ XmlDocument::XmlDocument() :	XmlNode(XmlNode::XML_DOCUMENT),
  *
  * \param documentName	Name of the document.
  */
-XmlDocument::XmlDocument(const std::string& documentName) :	XmlNode(XmlNode::XML_DOCUMENT),
+XmlDocument::XmlDocument(const std::string& documentName)
+: XmlNode(XmlNode::NodeType::XML_DOCUMENT)
+,
```
DanRStevens: The formatting here looks kind of funny, though perhaps we should address that as a separate issue. I've opened issue #242 to track this.

src/Xml/XmlNode.cpp
```diff
@@ -19,9 +19,9 @@ using namespace NAS2D::Xml;
 /**
  * Default c'tor.
  */
-XmlNode::XmlNode() :	XmlBase(),
-						_parent(nullptr),
-						_type(XML_UNKNOWN),
+XmlNode::XmlNode() :	XmlBase(), _parent(nullptr)
+, _type(NodeType::XML_UNKNOWN)
+,
```
DanRStevens: Again, the formatting on this looks particularly off.

src/EventHandler.cpp
cugone: For long-winded comparisons I tend to put the expected comparison result on the left so it's immediately visible.

src/Xml/XmlNode.cpp
cugone: See previous

src/Xml/XmlDocument.cpp
cugone: I blame Clang-Tidy and the adherence to tabs. :P Really it was me rapidly cycling through the output errors list and pasting the enum scope: F8, Ctrl+V. F8, Ctrl+V, etc. Will fix in #242.

src/EventHandler.cpp
DanRStevens: I do too. Just that the next 3 or 4 functions had it the other way around.

Plus, some programmers prefer to reverse the two, since if the left side is a variable, and you accidentally write `=` instead of `==`, it can compile silently and do the wrong thing. I'm pretty sure we have warnings for that now though. On the other hand, if the left side is an (lvalue) expression, then accidental assignment will give a compiler error.

A similar case is when comparing a variable to a constant. Putting the constant on the left is less prone to typo errors.

In this case, we're comparing an lvalue expression to a constant, so we should be good either way.

DanRStevens:
I was looking back at this, and the odd initializer list formatting change kind of got to me. I went ahead and updated it to more or less match the old formatting, but with the new code updates. From there I just sort of went and added my other suggestions.

I haven't heard anything back from Leeor, so I may just go ahead and merge this. I'll wait for the CI builds to get caught up first.

DanRStevens:
Oh shoot, I just noticed this PR makes a backwards incompatible change that breaks OPHD builds. Moving the `enums` out of the `EventHandler` class has changed their name scoping.

Old name: `NAS2D::EventHandler::MouseButton`
New name: `NAS2D::MouseButton`

Not sure if we should fix in NAS2D or OPHD.

cugone:
Considering OPHD has had a broken master build for 6 months I vote fixing OPHD as part of that. :)

DanRStevens:
Keep in mind Leeor works on a "develop" branch, rather than "master". That branch wasn't broken before, but it is broken by this.

Maybe we can add some simple 'using' aliases to fix this. Could always add a deprecation attribute on one if we really do want to move things to a new namespace.

It would be easier to develop OPHD, and upgrade to newer versions of NAS2D if we stopped breaking things all the time. :P

Issue #212, opened 2019-05-22
Determine special folders without using PhysFS


DanRStevens:
On the path towards replacing / removing the current `Filesystem` component (Issue #209), we need a way of determining appropriate platform specific write folder for things like game settings and saved games.

Currently the platform specific write folder is handled by PhysFS from within the `Filesystem` component. As this is likely to be removed soon, it needs a replacement.

----

@cugone, I believe you did some work towards such a method in the past. Are you able to dig up that work somewhere?

cugone:
I asked for some clarification on the issue last month: https://github.com/lairworks/nas2d-core/pull/120#discussion_r272785613 . I know there was some contention on whether or not to remove some of the stored directories, like `mUserDirectory` or `mPrefDirectory` at the time and my concerns were that the platform differences were very different and potentially ambiguous, particularly for windows, and I didn't get a response. Removing some of the directories will mitigate, but not eliminate, some concerns.

Have there been any changes to this line of thinking between @ldicker83 and you about what is and is not being removed? What about the concerns I've outlined?

DanRStevens:
Ahh, sorry, seems I missed replying to that comment.

That comment contains a decent summary of various options. Some aspects of the design we need are perhaps not entirely clear as there are quite a few subtle differences between paths and environment variables. We may need to do some experimentation to figure out what we really want.

In terms of what we need to support, the two most important are:
- Executable folder for reading static game assets
- Write folder for user configuration settings
- Write folder for saved games

These may or may not be the same folder, or a subfolder of another.

If we want, we might optionally allow retrieving other folders as well:
- User's home folder (Examples: `C:\Users\UserName\`, `/home/UserName/`)
- Current working directory (`std::filesystem::current_path()`)
- Startup working directory (the current working directory might change after startup)

These are optional, and maybe not so useful for games. The User's home folder, or related things such as Documents or Pictures might be useful in other contexts, such as image editors or document editors. That's out of scope though.

----

We may also want to give some consideration to things like Roaming or Local folders, as Windows distinguishes between the two in certain contexts. I believe we'd want the Roaming location.

----

Useful references:
https://docs.microsoft.com/en-gb/windows/desktop/shell/about-user-profiles
https://ss64.com/nt/syntax-variables.html
https://specifications.freedesktop.org/basedir-spec/latest/index.html#introduction

cugone:
[...]

Issue #179, opened 2019-05-10
Review Filesystem requirements and design


DanRStevens:
We should take a look at what we really need from the Filesystem component, and if the current design meets those well.

One consideration is that some aspects of the Filesystem interface seem tied to exposing specifics of the current PhysFS implementation, rather than any particular high level interface. Others may be in contrast with common practice in other libraries such as `<filesystem>`, and may be worth looking into why.

One aspect mentioned by @cugone is the `extension` method, which returns extensions without the leading ".". This means there is no way to distinguish between the following two cases:
```cpp
extension("file.");  // Currently "", perhaps should be "."
extension("file");  // Currently "", perhaps should be ""
```

We may also want to handle extensions used in directory names:
```cpp
extension("DirName.ext/");  // Currently "ext/", perhaps should be ".ext"
```

One thought I had, is perhaps splitting out virtual filesystem methods from real filesystem methods. An example of a virtual filesystem concern is to restrict writing to a specific data folder, or finding static game assets packaged with the executable file. An example of a real filesystem concern is finding the user folder, or the app specific write folder, which would then be used as a mount point for the virtual filesystem.

----

Tagging Issue #175 as related, and PR #176 as related.
Edit: Tagging Issue #118 as related.

ldicker83:
The original intention of the Filesystem object was to provide a platform independent way of reading from archive files that was transparent to the end user of NAS2D. Basically ,it allowed packing all data assets into a ZIP file (or similar) and reading from that. This was especially important at the time that NAS2D/The Legend of Mazzeroth was being developed (circa early 2008) as C++ had not really been updated, it was messy when dealing with filesystem code, etc.

With the recent updates to C++, filesystem operations are basically built into the language (with macOS continuing to be difficult). It basically means that using an external C-based library for filesystem operations is no longer necessary. Reading out of archive files is getting less and less important so at this point it's more about providing a consistent interface. But, seeing as C++17 effectively solved that problem the Filesystem object may now be superfluous.

The only advantage at this point is seamlessly reading from zip archives. Still though, that's more of a 'nice to have' type of deal and seems like it would be better for the end user to make said decisions to use the library. By not using it it would remove yet another dependency (zlib).

[...]

PR #243, opened 2019-12-22
Filesystem Update

cugone:
- [x] Remove the data folder parameter (API interface change, downstream projects affected) (#315)
 - [x] Improve API for arbitrary mounts of special real filesystem folders (API extension) (#317)
 - [x] Remove default mount of all folders (API behaviour change, downstream projects affected) (#320)
 - [x] Replace use of PhysFS
 - [x] Replace real filesystem methods (#212)

Refactors Filesystem to remove internal dependency on PhysFS and use C++17 `<filesystem>`.

This also serves as an initial push to see if any platforms break, particularly macOS and MinGW, because I do not have a mac nor linux dev machine.

(Edit: Added task list to PR for easier tracking)


DanRStevens:
Huh, Linux build is showing an error:
```cpp
src/Filesystem.cpp:23:10: fatal error: filesystem: No such file or directory
 #include <filesystem>
          ^~~~~~~~~~~~
```

Not too sure of the cause. I might be a bit too tired to do a full review or investigation today.

DanRStevens:
I looked at some of this. For the part I looked at, most of the changes were pretty well presented and easy to read through. I like the highly focused commits.

The number of changes are rather long, so I didn't get through all of them. Would be kind of nice if the changes could be broken up a bit into separate branches to make the review chunk smaller. Right now it's a bit hard to keep the different interrelated changes in mind. The current change set encompasses hundreds of lines, over 20 commits, and not all dependent changes are near each other. I understand though that some aspects of this change will necessarily be a bit large, and it would be difficult to break up certain parts of it. There do seem to be a number of small refactorings and setting changes that could probably be applied independently. The MacOS CI update stands out. A few of the smaller function updates stand out.

Right now I have some concern over commit e1765f44808270b506d2f16a72f5d0b0c79a7e6b, which makes a few breaking changes. I suspect some of those are fixed up in later commits, though I haven't been able to review them all yet. That also seems to be the commit that's responsible for the current CI failure. I'm also a little concerned about a lack of pre-existing tests.

I'll have to finish reviewing later. Meanwhile, if there are independent parts that could be applied without the breaking change, I would welcome new branches. That might allow some of these changes to be reviewed and merged more quickly. We could potentially rebase the remainder afterwards to reduce the size of the branch. With the current workflow, a branch is pass/fail merge/don't-merge as a whole, so even a small thing can hold up a whole branch.

cugone:
After some sleep I have some thoughts and comments:

---
I suspect the linux build is failing because the docker image being used is not new enough to have support for `<filesystem>` on all compilers.
I found the docker image on dockerhub: https://hub.docker.com/r/outpostuniverse/nas2d-circleci/tags (currently we're using 1.1);
If I understand correctly, the underlying linux distro is this one: https://ubuntu.pkgs.org/18.04/ubuntu-main-amd64/build-essential_12.4ubuntu1_amd64.deb.html
I have two concerns with this image: It hasn't been updated since 2017-9-18 (scroll down to change log) and more distressing, the versions of GCC and g++ being used are exactly 7.2; This is fine for g++ but GCC didn't add filesystem support until version 8! https://en.cppreference.com/w/cpp/compiler_support

Is there a way to use a newer linux distro in the docker?

---
There are a few commits where Clang-Tidy went behind my back and formatted the *entire file* between writing the new code and testing it. I would have to manually discard those changes before commiting. I wasn't paying attention and missed it in dac4632 :( Sorry about that.

---
Lots of little commits are better than one massive one. They were structured to only modify one function or feature at a time. The commits may feel big but that's because I was removing large portions of code. It would show up either way, commenting it out or removing it, as a large +/- change.

---
The use of the app name in the search path was lost on me as well as what exactly the test was supposed to be testing. Is it supposed to set up and test a file structure similar to `appname/data/`? The test only checked for a contained string, not a path.

DanRStevens:
> Lots of little commits are better than one massive one. They were structured to only modify one function or feature at a time. The commits may feel big but that's because I was removing large portions of code. It would show up either way, commenting it out or removing it, as a large +/- change.

The commits are fine. I like how they were done. The number of commits all grouped into one branch is perhaps part of the difficulty. I suspect some of these changes could be broken up a bit into multiple branches. Not all of them needed to be done together.

There seem to be two types of changes here. Some are simple refactoring, which no change in behaviour (or so little change that I don't expect them to impact downstream code). Others change behaviour, a bit at a time (though perhaps taken as a whole don't change behaviour overall in the end?). They two types are somewhat interspersed, which makes it a little harder to sort out what is happening. The two types of changes are mostly broken up into separate commits though, which is good.

I was pretty tired when reviewing last, so that may have impacted my perception of it.

----

I suspect we may need conditional compilation for the `<filesystem>` header, as most other projects I work on with a Linux build have a fallback to `<experimental/filesystem>`. The case is a bit worse for the Mingw-w64 compiler, as the `<experimental/filesystem>` header doesn't have every method in the C++17 standard. Notably, there are some path related methods missing, which I think you might be using.

Ubuntu 18.04 is the latest LTS (Long Term Support) release, so most people will be using that. I would like to be able to compile with a stock 18.04 image if possible. I believe that will be possible with `<experimental/filesystem>`. Failing that, I believe Ubuntu 18.04 offers updated packaged builds of GCC and Clang, though they must be specifically installed, and must be called with an explicit version number, as the default is still the older version.

I'll try to get back to this a bit later.

DanRStevens:
Tagging issue #179 as related.

src/Filesystem.cpp
```diff
@@ -406,40 +406,50 @@ bool Filesystem::closeFile(void* file) const
...
+	std::ofstream myFile{FS::path{file.filename()}, std::ios_base::binary};
 	if (!myFile)
 	{
-		if (mVerbose) { std::cout << "Couldn't open '" << file.filename() << "' for writing: " << getLastPhysfsError() << std::endl; }
+		if (mVerbose) { std::cerr << "Couldn't open '" << file.filename() << "' for writing.\n"; }
 		return false;
 	}
 
-	if (PHYSFS_writeBytes(myFile, file.bytes().c_str(), static_cast<PHYSFS_uint32>(file.size())) < static_cast<PHYSFS_sint64>(file.size()))
+	if(!myFile.write(reinterpret_cast<const char*>(file.bytes().data()), file.bytes().size()))
```
ldicker83: reinterpret_cast<const char*> raises giant warning flags. I suspect it's because you need const char* -- use raw_bytes() instead.

cugone: Will do. Out of curiosity, what's the warning you're getting?

ldicker83:
So I'm still reviewing this -- I'd still like to see CircleCI build this successfully but beyond that there are a number of question marks for me in the code. E.g., why so much of the previous code is left commented out? Does this behave the way it's intended to? Does the 'virtual' part of the filesystem continue to be virtual?

I've pulled a local copy of it and will be performing some tests on it to see how it behaves especially when integrated into a test branch with an active project that's built on NAS2D (e.g., OutpostHD). I have the next couple days off so I'll be looking into it.

cugone:
The code was commented out in case I needed it back if macOS or MinGW broke and I needed to put the physfs code back and I didn't want to trawl through commits to try and get it all back. Easier to just uncomment it again.

I can delete the dead code if you want.

Regarding #263, how would you want that handled? Using `argv_0` is no longer needed and the use of an organization name is confusing. Should that be a prefix to the `data/` path? Should it be specific to operating systems similar to `%APPDATA%/[organizationName]` and %APPDATA%/[organizationName]/[dataPath]`? or just removed entirely?

DanRStevens:
My extremely biased opinion also wants to see this build on CircleCI first. :wink:

----

The commented out code also worries me that this isn't feature complete. In particular, the comments concerning `argv_0` and `organizationName` suggests there is a change of behaviour. I would say the use of `argv_0` and an `organizationName` may still have some relevance, and would recommend keeping them for now.

Related to this is issue #212. That issue would use `appName`, though could potentially also use `organizationName` in some contexts, or on some platforms.

The `argv_0` argument may still have relevance to the virtual filesystem concept, and being able to locate the executable file, and finding static assets. In particular, the current working directory is quite likely not going to be the same as the executable directory. Further, the executable can be run through a symlink on Linux, and you may need the location of the symlink, rather than the target of it. It's actually fairly common practice on Linux to compile all options and variants into a single executable, but then trigger them by creating symlinks, where the name of the symlink determine the function of the executable, or in some cases, the default set of options.

I liked seeing the `std::map` to implement the virtual filesystem concept. We probably need a quick setup method to add certain expected folders to the virtual filesystem. On some platforms, such as Linux, that auto setup would likely make use of `argv_0`.

----

There are a couple of changes in this thread which feel a bit like they are being added in just to be taken out again. One is the `mIsInit` variable, and the other is `noexcept`. These do make sense in the context of this thread. It's the overall larger picture that I see where their use seems short lived.

If issue #263 were done first, it would remove the need for `mIsInit`. That's been in the back of my mind for awhile now, though I never managed to get it out into issue form until now, so it seems my delay has gone and made extra work for people.

For the `noexcept` part, I think I also need to open an issue to discuss that plan. The error reporting for `Filesystem` is not very good. Errors are logged internally, and then silently swallowed. The calling code never knows if an operation succeeded or not. They only way they could find out, would be to scan the log file, which is a horrible horrible idea. Instead, we should change the code to report errors with C++ exceptions, and push the error logging out into the client code. As most `Filesystem` methods will rely on an underlying filesystem, which is subject to external changes, and even hardware errors, almost all methods will be able to throw exceptions. Hence there are very few methods that should be marked as `noexcept` in the end. The current ability to mark many of them as `noexcept` is simply the result of the current error handling and internal swallowing of errors. I think that needs to change.

----

I'm hoping to touch on some of the CI build config soon. I'd like to possibly update some of the Docker images, and perhaps improve compiler coverage a bit. That'll be easier now that we've fixed some of the build breaks, and got other open PRs merged. I admit though, I'm still tempted to tackle some other low hanging fruit before diving back into Docker image updates.

ldicker83:
I've gone through all the changes here and I generally like what I see though I'd really like to see all of the test builds succeed even if it means using compiler directives to drop to <experimental/filesystem>. I'm also going to test it with a couple of NAS2D apps to see how much it really changes, etc.

This would be part of a different PR/Project where getting the user directory is part of the Filesystem class as that seems to be the way Microsoft is pushing Windows applications to behave. As I understand it, it works basically the same way that macOS X and Linux operate.

ldicker83:
Welp, those were supposed to be squashed but eh, whatever. You can see my frustration with GCC and the make system in general.

Anyway, linux now builds but only OSX and Windows are actually passing the Test's:

```
[ RUN      ] FilesystemTest.workingPath
test/Filesystem.test.cpp:48: Failure
      Expected: "data/"
To be equal to: fs.workingPath("data/file.extension")
      Which is: "data"
test/Filesystem.test.cpp:49: Failure
      Expected: "data/subfolder/"
To be equal to: fs.workingPath("data/subfolder/file.extension")
      Which is: "data/subfolder"
test/Filesystem.test.cpp:50: Failure
      Expected: "anotherFolder/"
To be equal to: fs.workingPath("anotherFolder/file.extension")
      Which is: "anotherFolder"
[  FAILED  ] FilesystemTest.workingPath (1 ms)
```

From what it's showing the directory separator is being stripped off in the Linux build. I'm not sure if it's because the Linux build is relying on experimental::filesystem or if it's because GCC is just garbage.

src/Filesystem.cpp
```diff
@@ -29,70 +29,33 @@
...
 	std::error_code ec{};
 	p = FS::canonical(p, ec);
 	if (ec)
 	{
-		std::cerr << "Filesystem::init failed:\n"
-				  << "dataPath: " << p
-				  << "\nReasons for failure:\n"
+		std::cerr << "Filesystem::init failed:" << std::endl
+				  << "dataPath: " << p << std::endl
+				  << "Reasons for failure:" << std::endl
```
cugone: `std::cerr` automatically flushes the stream, so does `std::endl`. Either the `std::endl` is unnecessary or switch to using the, incorrect in this context, `std::cout`.

Ref: https://www.youtube.com/watch?v=GMqQOEZYVJQ
Ref: https://www.youtube.com/watch?v=lHGR_kH0PNA

cugone:
@ldicker83 

Added a comment about `std::cerr` and using `std::endl` to 
0a00c9c

src/Filesystem.cpp
ldicker83: In this context the intention is to keep the code consistent with the rest of the code base in terms of line endings. While it may be unnecessary as two flushes are being done the consistency improvement is worth the change.

DanRStevens: Or implement #280, and remove this code as no longer relevant. :wink:

cugone:
The default behavior for `p.parent_path().make_preferred().string()` in `workingPath` is to drop the trailing slash. Currently the function is adding it back via `(p.parent_path() / "").make_preferred().string()`.

Is there a requirement that the path return the slash? If not, all tests pass.

src/Filesystem.cpp
cugone: Sounds good. :)

ldicker83:
So the good news is all the tests pass. (yay!)

The not so great news is that the filesystem itself is very broken and doesn't work as expected. (boo!)

Dan and I did a pair programming session today and started working on some smaller change sets that change and simplify the interface for the Filesystem which has the unfortunate side effect of conflicting with this PR.

That stated, I think it's a good base to work from and there's plenty here that can be cherry picked for a drop-in replacement to the Filesystem class.

I've been using OP2-Landlord (https://github.com/OutpostUniverse/op2-landlord) as a base project to link against NAS2D's latest changes to the Filesystem to see how it behaves in an actual use case. PhysicsFS hid a lot of the more complex implementation details of a virtual filesytem with multiple mount paths, which is the real culprit here; this version of Filesystem doesn't properly search the directory graph for named files.

---

**TL;DR**
Changes to the Filesystem interface breaks this PR and this PR's code doesn't behave as intended. It needs work but it's a good base to work from.

DanRStevens:
So I was thinking about what happened here. It's curious that this was made to compile and pass tests, though did not function as intended. That suggests the unit tests are inadequate, as they don't cover all required behaviour. The specification of the `Filesystem` class, provided by the unit tests, is incomplete.

I believe this incompleteness of testing revolves around the virtual filesystem details, and mounting of the data folder. I summarized my thoughts on how we might move forward into a few issues. Hopefully they provide a bit of a roadmap for incremental change.

New issues:
#315 Filesystem double mounting
#317 Filesystem special folders (using PhysFS) (related #212, without PhysFS)
#318 Filesystem `unmount`
#320 Remove Filesystem auto mounting
#321 Sort Filesystem methods into categories

----

In terms of replacing PhysFS, I believe the following path may work well (separate PRs):
1. Remove the data folder parameter (API interface change, downstream projects affected) (#315)
2. Improve API for arbitrary mounts of special real filesystem folders (API extension) (#317)
3. Remove default mount of all folders (API behaviour change, downstream projects affected) (#320)
4. Replace use of PhysFS
 a. Replace real filesystem methods (#212)
 b. What was attempted here

I believe with the other steps in place first, it would simplify the last part.

cugone:
There are changes to master that I want to pull into this PR be able to cleanly compile before getting started. (e.g. The C4706 warnings in `Sprite.cpp`) Is there a way to selectively merge one file from master into this PR without merging the entire branch?

DanRStevens:
You can use `git cherry-pick` to pull in a specific commit. That may be close enough if the commit was fairly isolated and only changed that one file.

I'm not sure if we want to continue hacking on this one long branch though. The history is already pretty complex and hard to follow. It might be better to pick from this branch into new branches. Well, for review and merging anyway. Could still use it as a testing grounds.

cugone:
> 
> 
> You can use `git cherry-pick` to pull in a specific commit. That may be close enough if the commit was fairly isolated and only changed that one file.

@DanRStevens  What if the branch has already been deleted both locally and on the server?

DanRStevens:
You need the commit somewhere in your local repository to cherry-pick it. The commit doesn't need to be on a branch. A non-garbage collected orphaned commit is sufficient. If the commit is from another repository, you may need to add a remote and fetch changes from the remote.

I assume though, you are trying to cherry-pick a change that has already been merged to master?

You run the cherry-pick from the branch you want the changes added to.

cugone:
The documentation for macOS is horrible. Practicing Google Fu for getting the exe path in macOS points to `_NSGetExecutablePath` but doesn't mention what header to include. Even more time spent googling says it should be `<mach-o/dydl.h>` but even with that AND `<CoreFoundation/CoreFoundation.h>` included the function is claimed to not exist.

I hate Apple.

At least I got Linux to build, now just to fixup the code so the tests pass.

DanRStevens:
A bit of an aside: To avoid long install times, usually GCC would be installed in the Dockerfile. The Dockerfile would be used to build a Docker image, the image would be uploaded to DockerHub. When the CI build runs, it downloads the image from DockerHub, and compiles that code in that. The compiler is then already installed, so there's no need to install it for each run. At least that's how it's done for Linux.

Of course that's easier to do once you have it all figured out. I guess it would be a little awkward to jump through all that when you're still testing.

Though, you could also install Docker Engine on Windows, which would give you effectively a VM you can test on and run builds from, without having to upload things to GitHub, and wait for CI builds to provide feedback. Might be a little faster.

cugone:
Might be.

Is there a way to install `g++-8` on the docker images and change the value of `CXX` to `g++-8` we have access to? That should mitigate trying to do it through the make file which is obviously not working as expected.

DanRStevens:
To work with Docker, you'll probably want to install the Windows version of Docker Engine. I highly recommend giving it a try. I'm afraid I only have experience with Docker on Linux, so I'm not quite sure what the experience is like on Windows.

Might take a bit of reading to figure out how it works. If you catch me early tomorrow morning, I might have time to go over some stuff with you on TeamViewer/Discord.

cugone:
Looks like `Docker For Windows` is a no-go. The newer version requires Windows 10 Pro at least and I only have Windows 10 Home. Additionally, the newer AND older versions of the software, `Docker Toolbox`, both require Virtualization on the CPU and I don't have that either. :/

cugone:
@DanRStevens 

Good news, the macOS build compiles but doesn't link, that's one step closer!
The linker doesn't like `-lstdc++fs` that is required for the linux build to work. Is there a way to conditionally add `-lstdc++fs` to link for Linux but leave it out for macOS builds?

cugone:
@DanRStevens 

`~/.local/share -> `/home/user/.local/share` doesn't exist on the circleci build. Could you step through the debugger it on your local system to see if it works and the issue is just a circleci configuration problem?

DanRStevens:
I'm afraid I'm unable to test for a few days.

In some contexts, the `~/` won't expand to the home folder, so that is something to watch for.

You might very well need to deal with creating that folder if it doesn't already exist.

DanRStevens:
First time running the tests, they passed. I then went and deleted the `~/.local/share/NAS2DUnitTests` folder, and the tests failed.

It looks like the `Filesystem` code may need to create the folder it needs if it doesn't already exist. Looking at some of the methods though, it seems surprising some of them would fail because a folder doesn't exist. This might indicate a weakness in the implementation.

Since this could be related, the `<filesystem>` methods often come in pairs. Some work on the real filsystem. Others work on just the path itself, without regard to the filesystem. The later ones that work just on the path itself typically have `lexical` in the name. If you're not using those methods, it may be worth looking into switching.

----

Also noted, a number of tests pass, but output logging information in the middle of the GoogleTest report output. We probably shouldn't be generating log output from the library itself.

----

This PR has grown too big to review. At some point we need to start picking out changes and packaging them into new smaller PRs.

I'm looking at closing this PR, though I don't want anyone to accidentally delete the branch. It may be safer to keep this open for now. We still want access to the code changes.

The branch can still be used for testing. Actually, even if the PR was closed, the branch could still be used for testing. The CI builds would continue to run (though the PR build on AppVeyor might stop running on a closed PR - speed optimization?).

PR #245, opened 2019-12-23
Fix Clang warnings

DanRStevens:
Closes #244

Fixes new Clang warnings concerning special member functions.

Issue #244, opened 2019-12-23
Clang 8+ warnings - explicitly defaulted operations are implicitly deleted


DanRStevens:
Clang 8+ (tested with Clang 8, 9, 10) produces a new warning:
```
In file included from src/Mixer/MixerNull.cpp:1:
include/NAS2D/Mixer/MixerNull.h:12:3: warning: explicitly defaulted copy constructor is implicitly deleted [-Wdefaulted-function-deleted]
                MixerNull(const MixerNull& other) = default;
                ^
include/NAS2D/Mixer/MixerNull.h:8:20: note: copy constructor of 'MixerNull' is implicitly deleted because base class 'NAS2D::Mixer' has a deleted copy constructor
        class MixerNull : public Mixer
                          ^
include/NAS2D/Mixer/Mixer.h:171:2: note: 'Mixer' has been explicitly marked deleted here
        Mixer(const Mixer&) = delete;
        ^
In file included from src/Mixer/MixerNull.cpp:1:
include/NAS2D/Mixer/MixerNull.h:13:3: warning: explicitly defaulted move constructor is implicitly deleted [-Wdefaulted-function-deleted]
                MixerNull(MixerNull&& other) = default;
                ^
include/NAS2D/Mixer/MixerNull.h:8:20: note: move constructor of 'MixerNull' is implicitly deleted because base class 'NAS2D::Mixer' has a deleted move constructor
        class MixerNull : public Mixer
                          ^
include/NAS2D/Mixer/Mixer.h:171:2: note: 'Mixer' has been explicitly marked deleted here
        Mixer(const Mixer&) = delete;
        ^
In file included from src/Mixer/MixerNull.cpp:1:
include/NAS2D/Mixer/MixerNull.h:14:14: warning: explicitly defaulted copy assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
                MixerNull& operator=(const MixerNull& other) = default;
                           ^
include/NAS2D/Mixer/MixerNull.h:8:20: note: copy assignment operator of 'MixerNull' is implicitly deleted because base class 'NAS2D::Mixer' has a deleted copy assignment operator
        class MixerNull : public Mixer
                          ^
include/NAS2D/Mixer/Mixer.h:172:9: note: 'operator=' has been explicitly marked deleted here
        Mixer& operator=(const Mixer&) = delete;
               ^
In file included from src/Mixer/MixerNull.cpp:1:
[...]

PR #246, opened 2019-12-23
Update `MixerSDL`

DanRStevens:
~~This is based on PR #245 (The first 3 commits are from the other branch). It is dependent on those changes to not introduce additional Clang warnings.~~ Edit: Rebased on master after merging base branch.

This completes some additional changes touched on by issue #244, but which were not core parts of the issue affecting the original warnings.

### DIFF
--- a/include/NAS2D/Mixer/MixerSDL.h
+++ b/include/NAS2D/Mixer/MixerSDL.h
@@ -29,6 +29,8 @@ public:
 	MixerSDL();
 	MixerSDL(const MixerSDL&) = delete;
 	MixerSDL& operator=(const MixerSDL&) = delete;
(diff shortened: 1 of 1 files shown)


include/NAS2D/Mixer/MixerNull.h
```diff
@@ -5,7 +5,7 @@
 namespace NAS2D
 {
 
-	class MixerNull : public Mixer
+	class MixerNull final : public Mixer
```
cugone: Consider the use of `final` carefully before use. Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-final

PR #232, opened 2019-12-17
Fix Clang parenthesis warning

DanRStevens:
Extra parenthesis seems to indicate intent and silence the warning for all compilers.

PR #271, opened 2019-12-24
Fixes C4706

### LINKED ISSUE #201: Fix Warning C4706 - Using result of assignment as condition
A recent change has created a lot of compiler warnings:
```cpp
src/Configuration.cpp:219:18: warning: using the result of an assignment as a
      condition without parentheses [-Wparentheses]
                while (xmlNode = root->iterateChildren(xmlNode)) // warn...
                       ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is the result of many lines similar to the following:
```cpp
		while (xmlNode = root->iterateChildren(xmlNode)) // warning C4706: intended
```

Previously, these lines had an extra set of parenthesis which were silencing the warning:
```cpp
		while ((xmlNode = root->iterateChildren(xmlNode)))
```

----

It may be possible to provide a bit of support code to enable [range-based for loops](https://en.cppreference.com/w/cpp/language/range-for) to replace those `while` loops. In particular, the existence of `begin` and `end` methods can allow the compiler to generate a range-based for loop.

There are two ways to use `begin`/`end`. Either as member methods on a class, or as free floating methods which take an iterable parameter. If we wish to avoid modifying existing classes (perhaps from within the XML library), we can instead use the later form of free floating methods.

cugone:
Closes #201

Fixes the C4706 warnings

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -200,8 +200,9 @@ bool Configuration::readConfig(const std::string& filePath)
 
 
 		// Start parsing through the Config.xml file.
-		XmlNode *xmlNode = nullptr;
-		while ((xmlNode = root->iterateChildren(xmlNode)))
+		for (auto xmlNode = root->iterateChildren(nullptr);
+			 xmlNode != nullptr;
+			 xmlNode = root->iterateChildren(xmlNode))
 		{
 			if (xmlNode->value() == "graphics") { parseGraphics(xmlNode); }
 			else if (xmlNode->value() == "audio") { parseAudio(xmlNode); }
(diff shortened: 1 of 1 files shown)


DanRStevens:
Tagging PR #232 as related. I think that PR may have actually eliminated the warnings in Issue #201. It wasn't exactly the most satisfactory fix though.

cugone:
Weird. I was still getting them in my debug build (hence this PR).

DanRStevens:
There may be some confusion over two different error messages here. In Issue #201, the error quoted is a Clang error message. The C4706 in the comment is an MSVC reported error message.

I haven't been seeing the MSVC error message on AppVeyor, though that may be a difference in analysis settings for the `Debug` configuration.

I think what happened is the two compilers were fighting about what was worse. MSVC complained about something, so code was changed to fix the warning. Then Clang started complaining about the fixed code. Or maybe the other way around. I can't remember now.

Regardless, neither compiler seems to be complaining now, so that's good.

DanRStevens:
After re-reading issue #201 I noticed the solution here is very similar to a related solution that Leeor suggested for a different block of code. As the tests are passing, I'm going to go ahead and merge this.

PR #279, opened 2019-12-28
Adds Pre-Build Event for VS2019

### LINKED ISSUE #278: Pulling fresh from master doesn't build on Windows
I pulled a fresh copy of master for my laptop and several issues popped up:

 - GL/glew.h couldn't be found.
 - physfs.h couldn't be found (this will go away with future updates).
 - Several `.pdb`s are listed as using an obsolete format and need to be deleted and recompiled.

This is probably because vcpkg isn't being setup correctly on build. Setting up a `Properties > Configuration > Build Event > Pre-Build Event` to have vcpkg download and setup the required submodules might fix the issue.

cugone:
Closes #278 

Runs `($SolutionDir)../../VcpkgDeps.bat` as a Pre-Build Event to install glew/physfs dependencies prior to trying to build.

### DIFF
--- a/proj/vs2019/NAS2D.vcxproj
+++ b/proj/vs2019/NAS2D.vcxproj
@@ -104,6 +104,12 @@
     <Link>
       <GenerateDebugInformation>true</GenerateDebugInformation>
     </Link>
+    <PreBuildEvent>
+      <Command>$(SolutionDir)../../InstallVcpkgDeps.bat</Command>
+    </PreBuildEvent>
+    <PreBuildEvent>
+      <Message>Installing vcpkg dependencies</Message>
+    </PreBuildEvent>
   </ItemDefinitionGroup>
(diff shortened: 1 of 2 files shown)


cugone:
The batch file is being run for both NA2D and test projects. Since the installs are per machine we can try to only run it on NAS2D since it's the base project.

cugone:
I double-checked the file itself. In cases where `PLATFORM` isn't defined it explicitly runs both the x64 and x86 versions.

DanRStevens:
Oh, I meant after the initial AppVeyor config to run the batch file, it gets run twice more, once for NAS2D, and then again for the test project (until the later edit to remove it). Each time it goes through each dependency and outputs a large block of text. Hence a bunch of build spam, and some possibly noticeable extra build time.

In terms of the `PLATFORM` environment variable, it will be defined throughout the AppVeyor run. I checked the latest run, and there is only output for a single platform. The duplication is simply coming from the extra calls to run the batch file multiple times.

PR #288, opened 2019-12-29
Remove verbose logging from Filesystem

### LINKED ISSUE #284: Remove verbose logging from `Filesystem`
Most verbose logging in the `Filesystem` class is perhaps indicative of unit tests that should exist. We should add unit tests to cover these cases, and then remove the logging code.

If library users want to log, they can log at the point of call. That would place logging in control of the library user, rather than dictated within the library.

There was some earlier discussion of this in:
https://github.com/lairworks/nas2d-core/issues/280#issuecomment-569524380

ldicker83:
Closes #284

### DIFF
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -51,15 +51,12 @@ public:
 	bool isDirectory(const std::string& path) const;
 	void makeDirectory(const std::string& path) const;
 
-	void toggleVerbose() const;
-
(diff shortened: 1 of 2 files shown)


DanRStevens:
Oh I just noticed something wonderful about this update. The unit test runs are no longer spammed by verbose logging messages, which used to get interspersed with the unit test output. I love how much cleaner the unit test output is now.

PR #297, opened 2019-12-31
Fixes C4706 warnings

cugone:
Fixes the VS2019 C4706 warnings emitted in Sprite


cugone:
AppVeyor isn't setup to build in Debug mode.

DanRStevens:
Ahh, right, I forgot about that. Would be nice to build in all configurations, thought we'd need to address the build performance before doing that.

I looked up the similar related change. Reference PR #271.

PR #298, opened 2019-12-31
Moves Color_4ub to Color_4ub file and renames from all caps

### LINKED ISSUE #259: Move pre-defined Color_4ub colors from Renderer to Primitives file
It's quite likely that another part of the engine or downstream project will want to use them. Keeping them local to the Renderer translation unit promotes bad practices of duplicating them.

At the same time they should be renamed to not use the out-dated "all caps for constants/enums" "rule" as that has been relegated to macros only.

N.B. I have actually had this bite me in the past. I named a scoped enum value for keys as `DELETE`. By including the `<windows.h>` header it created a macro named `DELETE` and since macros are strict and direct text replacement, things exploded.

Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps
Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-not-CAPS
Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-ALL_CAPS
Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-all-caps

cugone:
Closes #259

### DIFF
--- a/include/NAS2D/Renderer/Color.h
+++ b/include/NAS2D/Renderer/Color.h
@@ -21,6 +21,23 @@ namespace NAS2D {
 class Color_4ub
 {
 public:
+	static const Color_4ub Black;
+	static const Color_4ub Blue;
+	static const Color_4ub Bright_green;
+	static const Color_4ub Cyan;
+	static const Color_4ub Green;
+	static const Color_4ub Grey;
+	static const Color_4ub Magenta;
+	static const Color_4ub Navy;
+	static const Color_4ub Orange;
+	static const Color_4ub Red;
+	static const Color_4ub Silver;
+	static const Color_4ub White;
+	static const Color_4ub Yellow;
+	static const Color_4ub Normal;
+	static const Color_4ub NormalZ;
+	static const Color_4ub NoAlpha;
+
 	Color_4ub() = default;
 	Color_4ub(uint8_t r, uint8_t g, uint8_t b, uint8_t a);
 
@@ -38,7 +55,10 @@ public:
 	void alpha(uint8_t alpha);
 
 private:
-	unsigned char mR = 255, mG = 255, mB = 255, mA = 255;
+	unsigned char mR = 255;
+	unsigned char mG = 255;
+	unsigned char mB = 255;
+	unsigned char mA = 255;
 };
 
 } // namespace
--- a/include/NAS2D/Renderer/Color.h
+++ b/include/NAS2D/Renderer/Color.h
@@ -55,10 +55,10 @@ public:
 	void alpha(uint8_t alpha);
 
 private:
-	unsigned char mR = 255;
-	unsigned char mG = 255;
-	unsigned char mB = 255;
-	unsigned char mA = 255;
+	uint8_t mR = 255;
+	uint8_t mG = 255;
+	uint8_t mB = 255;
+	uint8_t mA = 255;
 };
 
 } // namespace
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -21,23 +21,6 @@
 
 namespace NAS2D {
 
-// Color Presets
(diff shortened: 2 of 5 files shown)


include/NAS2D/Renderer/Color.h
```diff
@@ -38,7 +55,10 @@ class Color_4ub
 	void alpha(uint8_t alpha);
 
 private:
-	unsigned char mR = 255, mG = 255, mB = 255, mA = 255;
+	unsigned char mR = 255;
+	unsigned char mG = 255;
+	unsigned char mB = 255;
+	unsigned char mA = 255;
```
DanRStevens: This should probably have been done in a separate commit.

include/NAS2D/Renderer/Color.h
```diff
@@ -21,6 +21,23 @@ namespace NAS2D {
...
+	static const Color_4ub Magenta;
+	static const Color_4ub Navy;
+	static const Color_4ub Orange;
+	static const Color_4ub Red;
+	static const Color_4ub Silver;
+	static const Color_4ub White;
+	static const Color_4ub Yellow;
+	static const Color_4ub Normal;
+	static const Color_4ub NormalZ;
+	static const Color_4ub NoAlpha;
```
DanRStevens: Can probably just declare these as `static constexpr`, and then initialize them in the header. A `constexpr` variable is also implicitly an `inline` variable.

Reference: [How do inline variables work?](https://stackoverflow.com/questions/38043442/how-do-inline-variables-work)

include/NAS2D/Renderer/Color.h
DanRStevens: Oh, and on the subject of that line, it should probably use type `uint8_t`. That would be more consistent with the constructor parameter types, and accessor method types.

cugone: > 
> 
> Oh, and on the subject of that line, it should probably use type `uint8_t`. That would be more consistent with the constructor parameter types, and accessor method types.

I was just thinking the same thing.

include/NAS2D/Renderer/Color.h
cugone: > 
> 
> Can probably just declare these as `static constexpr`, and then initialize them in the header. A `constexpr` variable is also implicitly an `inline` variable.
> 
> Reference: [How do inline variables work?](https://stackoverflow.com/questions/38043442/how-do-inline-variables-work)

I keep getting either `C2078: Too many initializers` when I use `static const Color_4ub Black{0,0,0,255}` or `C2059: syntax error 'constant'` when I use `static const Color_4ub Black(0,0,0,255)` and `expression must have a constant value` or `expected a type specifier` when each is used with `constexpr` respectively.

DanRStevens: Hmm, I played around with this a bit. I managed to get something working, but not with the constants inside of the `Color_4ub` class. In hindsight, it does seem a little bit strange to have them there. The constants are static instances contained in their own class type. Their own class type will necessarily be incomplete until after the scope has ended, and class scope can't be re-opened for addition.

Instead, to use `constexpr`, I had to declare them after the close of the class. I also needed to declare the constructor as `constexpr`. That requires all initialization to be done through the initializer list (which is already true here). It also requires the full constructor definition (not just the declaration) be visible before defining the objects that use the constructor. If the `constexpr` variables are defined in the header, then the constructor definition also needs to be contained in the header.

Given the design considerations, I'm tempted to merge this as is. If we wish to investigate using `constexpr`, perhaps that would be best as its own issue. Will revisit tomorrow, unless Leeor wants to make a decision meanwhile.

ldicker83: Was going to say, constexpr should be a separate issue. This will suffice for now. Thanks!

PR #300, opened 2020-01-01
Converts Point and Rectangle classes to struct

### LINKED ISSUE #268: Convert Point and Rectangle from class to struct
The Point and Rectangle classes behave more like structs than classes. As such, they should simply be declared as such. In particular, the private fields all have public getter/setter methods, making the fields essentially public anyway. Direct access should just be given.

For backwards compatibility, we could still offer the same getter/setter methods, though they should be deprecated and removed at some point.

From the C++ Core Guidelines:
[C.2: Use class if the class has an invariant; use struct if the data members can vary independently](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-struct)

There are no invariants to be maintained. The `x`, `y`, `width`, `height` values may all vary independently.

cugone:
Closes #268

This is a breaking change to any file that forward declares `Point_2d`, `Point_2df`, `Rectangle_2d`, and `Rectangle_2df` as a `class` instead of `struct`.

### DIFF
--- a/include/NAS2D/Renderer/Point.h
+++ b/include/NAS2D/Renderer/Point.h
@@ -12,15 +12,15 @@
 
 namespace NAS2D {
 
-class Point_2d;
-class Point_2df;
+struct Point_2d;
+struct Point_2df;
 
 
 /**
  * \class	Point_2d
  * \brief	2D point.
  */
-class Point_2d
+struct Point_2d
 {
 public:
 	Point_2d() = default;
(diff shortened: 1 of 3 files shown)


ldicker83:
I think removal/deprecation of the get/set methods should be a separate change from this PR and changeset. And that's if they even really need to be removed.

There are functions that provide references for whatever reason to the internals -- I don't recall that being intended due to encapsulation but again, a separate change set.

PR #304, opened 2020-01-02
Remove name from mixer and renderer

### LINKED ISSUE #250: Remove or refactor `name()` method in `Mixer` and `Renderer`
The `name()` method doesn't seem to be of much use in the `Mixer` and `Renderer` classes. We should consider removing it. If for some reason it is still desired, we should perhaps consider changing the design.

For `Mixer` it is never called anywhere in NAS2D or OPHD. (Any other projects?)

For `Renderer`, it is only ever called in the `RendererOpenGL` constructor (where the value is initially set), and in one other method where it looks to be called by accident and appears to be a bug (see issue #249).

I suggest removing this function. Due to lack of use, removing it shouldn't cause downstream breaks in OPHD.

----

If the method is desired, perhaps for debugging output reasons, then we should consider changing the design.

The current design requires a name to be set in a constructor initializer list of the derived class. It also requires storage within any derived class. This is a bit wasteful, as the value does not appear designed to change during the lifetime of the class. There are no methods to update it, nor any internal code that changes it.

Instead, a design that allows for a human readable class name to be returned, and not require instance storage, is to declare the `name()` function as virtual. It's already returning a `const std::string&`, so returning a hardcoded constant would be just fine.

Even better, would be to change the return type to `std::string_view`. With `const std::string&`, a `std::string` object must be constructed at runtime. If the function were to return a string literal (`return "OpenGL Renderer";`), that construction would happen every time the `name()` method was called. If the `std::string` object was moved global, it would only ever be constructed once each time the executable is run. However, it would be constructed globally regardless of whether or not that particular class was ever instantiated. Instead, if a `std::string_view` were used, it could be compile time constructed as it is a `constexpr` compatible type. Returning it would invoke the copy constructor, but it's a cheap operation for `std::string_view`, and as the string has global semantics, there are no ownership issues with the non-owning nature of `std::string_view`.

----

The easier solution though is to just remove it.

ldicker83:
Closes #250

### DIFF
--- a/include/NAS2D/Mixer/Mixer.h
+++ b/include/NAS2D/Mixer/Mixer.h
@@ -147,13 +147,6 @@ public:
 	virtual void musicVolume(int level) = 0;
 
 	/**
-	 * Gets the name of the Mixer.
-	 *
-	 * \return	A /c std::string containing the name of the Mixer.
-	 */
-	const std::string& name() const;
-
-	/**
 	 * Gets a reference to a NAS2D::Signals::Signal0<void>, a signal raised
 	 * when a Music track has finished playing.
 	 */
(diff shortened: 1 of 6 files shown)


include/NAS2D/Mixer/Mixer.h
```diff
@@ -164,17 +157,9 @@ class Mixer
...
-	 *
-	 * This c'tor is not public and can't be invoked externally.
-	 */
-	Mixer(const std::string& name);
-
 	NAS2D::Signals::Signal0<> mMusicComplete; /**< Callback used when music finished playing. */
 
 private:
-	std::string mName; /**< Internal name of the Renderer. */
 };
```
DanRStevens: The `private:` part should be removed too.

ldicker83: Whoops

PR #312, opened 2020-01-06
Update `string_format`

### LINKED ISSUE #293: GCC Warning - null destination pointer
Newer versions of GCC are showing the following 4 warnings (they are all basically the same):
```
In file included from include/NAS2D/Filesystem.h:13:0,
                 from src/Resources/Image.cpp:14:
include/NAS2D/StringUtils.h: In function 'std::__cxx11::string NAS2D::string_format(const string&, Args ...) [with Args = {const char*, int}]':
include/NAS2D/StringUtils.h:40:30: warning: null destination pointer [-Wformat-truncation=]
   std::size_t size = snprintf(nullptr, std::size_t{0u}, format.c_str(), args...) + std::size_t{1u};

                      ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In file included from include/NAS2D/Filesystem.h:13:0,
                 from src/Resources/Font.cpp:14:
include/NAS2D/StringUtils.h: In function 'std::__cxx11::string NAS2D::string_format(const string&, Args ...) [with Args = {int}]':
include/NAS2D/StringUtils.h:40:30: warning: null destination pointer [-Wformat-truncation=]
   std::size_t size = snprintf(nullptr, std::size_t{0u}, format.c_str(), args...) + std::size_t{1u};

                      ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

include/NAS2D/StringUtils.h: In function 'std::__cxx11::string NAS2D::string_format(const string&, Args ...) [with Args = {unsigned int}]':
include/NAS2D/StringUtils.h:40:30: warning: null destination pointer [-Wformat-truncation=]
   std::size_t size = snprintf(nullptr, std::size_t{0u}, format.c_str(), args...) + std::size_t{1u};

                      ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

include/NAS2D/StringUtils.h: In function 'std::__cxx11::string NAS2D::string_format(const string&, Args ...) [with Args = {int, int}]':
include/NAS2D/StringUtils.h:40:30: warning: null destination pointer [-Wformat-truncation=]
   std::size_t size = snprintf(nullptr, std::size_t{0u}, format.c_str(), args...) + std::size_t{1u};

                      ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The code looks basically good. The destination pointer is `nullptr`, but the size is also `0`, so should be safe. The intention of the code is to calculate how much buffer space is needed, and then do an appropriate allocation before doing the operation for real.

Nevertheless, the warning should be eliminated somehow.

DanRStevens:
Closes #293

GCC no longer complains about null destination buffer.

Eliminates copy from temporary buffer.

### DIFF
--- a/include/NAS2D/StringUtils.h
+++ b/include/NAS2D/StringUtils.h
@@ -1,6 +1,6 @@
 // ==================================================================================
 // = NAS2D
-// = Copyright © 2008 - 2019 New Age Software
+// = Copyright ï¿½ 2008 - 2019 New Age Software
 // ==================================================================================
 // = NAS2D is distributed under the terms of the zlib license. You are free to copy,
 // = modify and distribute the software under the terms of the zlib license.
@@ -37,10 +37,12 @@ namespace NAS2D
 	template<typename... Args>
 	std::string string_format(const std::string& format, Args... args)
 	{
-		std::size_t size = snprintf(nullptr, std::size_t{0u}, format.c_str(), args...) + std::size_t{1u};
-		std::unique_ptr<char[]> buffer(new char[size]);
-		snprintf(buffer.get(), size, format.c_str(), args...);
-		return std::string(buffer.get(), buffer.get() + size - 1);
+		std::string buffer;
+		std::size_t size = snprintf(buffer.data(), buffer.size(), format.c_str(), args...);
+		buffer.resize(size + 1); // Including null (avoid clipping/undefined behavior)
+		snprintf(buffer.data(), buffer.size(), format.c_str(), args...);
+		buffer.resize(size); // Strip null temrinator
+		return buffer;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


include/NAS2D/StringUtils.h
```diff
@@ -1,6 +1,6 @@
 // ==================================================================================
 // = NAS2D
-// = Copyright � 2008 - 2019 New Age Software
+// = Copyright � 2008 - 2019 New Age Software
```
cugone: Small nit happened here. Consider changing to non-unicode string of three characters: `(c)`.

DanRStevens: Gah! I hate when that happen. Atom seems to silently and automatically mess with certain encoded symbols.

So it looks like the © was ASCII encoded, rather than Unicode encoded. I believe other files in the project do similar. Perhaps we should switch to Unicode encoding. I'll open a new issue to track this.

PR #314, opened 2020-01-06
Convert copyright symbols to Unicode

### LINKED ISSUE #313: The Copyright symbols are ASCII encoded
The © symbols in header comments are often ASCII encoded. Perhaps we should switch to using a Unicode encoding (UTF-8) for this symbol.

When editing files with the Atom editor, it will automatically and silently garble high ASCII encoded characters, such as ©. This is easy to miss, and sometimes gets checked in along with other changes. If we pre-convert all symbols to Unicode, that may prevent them from being converted/garbled in the future. These days, most editors assume files are UTF-8 encoded, rather than ASCII encoded.

----

Edit: Slight clarification. US-ASCII is technically 7-bits, so "high ASCII" codes, really implies some sort of non-US-ASCII encoding. Using the Linux `file` utility to identify files, it lists some files as `UTF-8`, and some files as `ISO-8859`.

DanRStevens:
Closes #313

This changes the encoding of the copyright symbol from ISO-8859 (aka Latin-1) to Unicode UTF-8. This is generally the only high-ASCII symbol used in the source files, and hence the only symbol with a distinct binary encoding between ISO-8859 and UTF-8.

### DIFF
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -1,6 +1,6 @@
 // ==================================================================================
 // = NAS2D
-// = Copyright � 2008 - 2019 New Age Software
+// = Copyright © 2008 - 2019 New Age Software
 // ==================================================================================
 // = NAS2D is distributed under the terms of the zlib license. You are free to copy,
 // = modify and distribute the software under the terms of the zlib license.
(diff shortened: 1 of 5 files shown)


DanRStevens:
I assume that's the only problem symbol. I'm not quite sure of a way to check.

PR #305, opened 2020-01-02
Use parameter packs for delegates

DanRStevens:
I got to looking at this code after doing the `musicComplete` callback change. It reminded me of the idea to reduce the bulk of the code by using template parameter packs. I thought I'd try my hand at it, now that I'm a little more familiar with templates. It seemed to go well.

~~I think we should probably write test code first, before merging it. Meanwhile, I thought I'd share this now and see what people think.~~ Edit: Unit tests added by PR #326.

----

~~Edit: I haven't yet touched the related `Signal` class.~~ Edit: Done.

PR #319, opened 2020-01-06
Fix Reserved Identifiers

### LINKED ISSUE #258: Fix implementation reserved identifiers
While reviewing #247 I noticed several variable names and classes that started with an underscore followed by a capital letter. Though not illegal, identifier names of this construction (along with any name that contains an underscore followed by another underscore) are reserved for the implementation and should never be used. Since this is a targeted change but may be pervasive I think one big feature branch with smaller commit changes to the entire project are required so the changes can be applied at once.

cugone:
Closes #258

### DIFF
--- a/src/EventHandler.cpp
+++ b/src/EventHandler.cpp
@@ -13,7 +13,7 @@
 
 // UGLY ASS HACK!
 // This is required for mouse grabbing in the EventHandler class.
-extern SDL_Window* _WINDOW;
+extern SDL_Window* underlyingWindow;
 
 
 using namespace std;
@@ -492,7 +492,7 @@ EventHandler::QuitEventCallback& EventHandler::quit()
  */
 void EventHandler::grabMouse()
 {
-	if (_WINDOW) { SDL_SetWindowGrab(_WINDOW, SDL_TRUE); }
+	if (underlyingWindow) { SDL_SetWindowGrab(underlyingWindow, SDL_TRUE); }
 }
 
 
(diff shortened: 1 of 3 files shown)


cugone:
Commit 08a371c conflicts with PR #305. Prefer cherry-picking all commits afterwards and incorporating changes to Signal class directly into #305.

src/Renderer/RendererOpenGL.cpp
```diff
@@ -829,18 +830,6 @@ void fillTextureArray(GLfloat x, GLfloat y, GLfloat u, GLfloat v)
...
- * The following code was developed by Chris Tsang and lifted from:
- *
- * http://www.codeproject.com/KB/openGL/gllinedraw.aspx
- *
- * Modified: Removed option for non-alpha blending and general code cleanup.
- *
- * This is drop-in code that may be replaced in the future.
- */
-static inline float _ABS(float x) { return x > 0 ? x : -x; }
-
```
DanRStevens: I don't think this comment pertains to the `_ABS` method alone, but also to the code that follows it.

src/EventHandler.cpp
```diff
@@ -13,7 +13,7 @@
 
 // UGLY ASS HACK!
 // This is required for mouse grabbing in the EventHandler class.
-extern SDL_Window* _WINDOW;
+extern SDL_Window* _underlyingWindow;
```
DanRStevens: How come this one was renamed, but retains a leading underscore?

The variable looks to be defined at top level scope, so may still be reserved.

DanRStevens:
Ok, I did a bit of cleanup. I split off the `Delegate` change into a new temporary branch `removeUnderscoreDelegate`. I then rebased this branch to remove that commit so there would be no conflicts with #305. As part of the rebase, I also reordered some of the commits, so the fixup commits came right after the commits they were fixing, and then squashed them together. The result should be a bit cleaner, in that every commit should compile on its own now, and there's no interleaving of changes.

If these changes looks good, I'd like to merge this now.

PR #327, opened 2020-01-07
Add new source files to unit test project

DanRStevens:
PR #326 added new unit tests, but accidentally missed adding the new unit tests to the Visual Studio unit test project file. This was soon caught in PR #305.

Curiously, this crashes on Windows. Meanwhile, the new code runs and passes on Linux and MacOS. I might need a bit of help figuring out what is going on here.

### DIFF
--- a/test/Delegate.test.cpp
+++ b/test/Delegate.test.cpp
@@ -12,7 +12,8 @@ namespace {
 }
 
 
-TEST(Delegate, DelegateCall) {
+// Disabled due to GoogleMock crashes on Windows
+TEST(Delegate, DISABLED_DelegateCall) {
 	MockHandler handler;
 	auto delegate = NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod);
 	EXPECT_CALL(handler, MockMethod(0));
@@ -21,7 +22,8 @@ TEST(Delegate, DelegateCall) {
 	delegate(1);
 }
 
-TEST(Delegate, DelegateCallConst) {
+// Disabled due to GoogleMock crashes on Windows
+TEST(Delegate, DISABLED_DelegateCallConst) {
 	const MockHandler handler;
 	auto delegate = NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod);
 	EXPECT_CALL(handler, MockMethod(0));
--- a/test/Signal.test.cpp
+++ b/test/Signal.test.cpp
@@ -21,8 +21,9 @@ TEST(Signal, ConnectEmitDisconnect) {
 	signal.connect(delegate);
 	EXPECT_FALSE(signal.empty());
 
-	EXPECT_CALL(handler, MockMethod());
-	signal.emit();
+	// Disabled due to GoogleMock crashes on Windows
+	// EXPECT_CALL(handler, MockMethod());
+	// signal.emit();
 
 	signal.disconnect(delegate);
 	EXPECT_TRUE(signal.empty());
--- a/test/test.vcxproj
+++ b/test/test.vcxproj
@@ -47,8 +47,10 @@
     <OutDir>$(ProjectDir)$(Platform)\$(Configuration)\</OutDir>
   </PropertyGroup>
   <ItemGroup>
+    <ClCompile Include="Delegate.test.cpp" />
     <ClCompile Include="Filesystem.test.cpp" />
     <ClCompile Include="MathUtils.test.cpp" />
+    <ClCompile Include="Signal.test.cpp" />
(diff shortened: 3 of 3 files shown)


cugone:
The crash is located here:
```
namespace {
	class MockHandler {
	public:
		MOCK_CONST_METHOD1(MockMethod, void(int value)); //<---- 2. ...Crashes deep inside this macro
		MOCK_CONST_METHOD1(MockMethod2, void(int value));
	};
}

TEST(Delegate, DelegateCall) {
	MockHandler handler;
	auto delegate = NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod);
	EXPECT_CALL(handler, MockMethod(0));
	EXPECT_CALL(handler, MockMethod(1));
	delegate(0); //<---- 1. During this call...
	delegate(1);
}
```

MSVC debugger is returning a crash on a read access violation:
```
Exception thrown: read access violation.
this->**_Myproxy** was 0xCCCCCCCC. occurred
```

The `0xCCCCCCCC` is a magic debug value indicating an uninitialized pointer that is being dereferenced.

DanRStevens:
Hmm, most curious. This may be a challenge to debug.

I notice some of the `Delegate` code contains platform detects, so some aspects are MSVC specific.

cugone:
Does the newer recommended `MOCK_METHOD` macro not work in all compilers?

DanRStevens:
The newer `MOCK_METHOD` macro works, with the GoogleTest v1.10.0 upgrade, however, Clang spits out reams of warning text about it if you try to compile with `-Wpedantic`. An issue was opened for it on the GoogleTest project about 7 days ago.

cugone:
I switched over to the newer `MOCK_METHOD` and the error still occurs. That tells me it has something to do with either the test code itself or how the underlying delegate is constructed.

ldicker83:
The Delegate class is based on very old code that makes a lot of assumptions, was from a CodeProject article (I think it's linked in the source file?)

Anyway, it may be worth looking at https://www.codeproject.com/Articles/1170503/The-Impossibly-Fast-Cplusplus-Delegates-Fixed on a C++11 implementation of the Delegate class. Probably plays nicer with modern compilers.

cugone:
Commenting out the `EXPECT_CALL` macros causes things to not crash.

If there are any uncaught exceptions the tests are designed to fail.

You could add a debug-only `static bool` that detects if the delegate was called and then just test it with `EXPECT_TRUE/EXPECT_FALSE`

DanRStevens:
Well, that solves the immediate problem of the crashing, though leaves the tests doing pretty limited checks. In particular, there are no longer any tests that actually call a delegate method.

I kind of suspect this may be a problem within GoogleMock itself, or perhaps the packaging of it, which might not be binary compatible with the settings used by the project. In particular, this is the first use of the `MOCK_METHOD` macros in this project.

With that said, I can't exactly rule out the `Delegate` code itself, particularly since it has Windows only code paths. It may function differently on Windows than on Linux or MacOS, where all tests pass. However, since OPHD has been using the `Delegate` code for quite some time now, mostly on Windows, and hasn't been crashing, I somehow doubt the cause could entirely be the `Delegate` code.

DanRStevens:
Ok, after a lot of digging, reading through GoogleMock source code, reading through the Delegate code, ... I found nothing. But then I got lucky. I tried searching GitHub for some of the error message posted by cugone, hoping to find some relevant section of code within GoogleMock, but instead was lead to a recent issue (27 days old):
https://github.com/google/googletest/issues/2628

It describes the error we saw, and ties it quite specifically to:
GoogleTest v1.10.0, using Visual Studio 2019, targeting the C++17 standard

Notably, it doesn't affect older versions of GoogleTest, nor Linux or MacOS compilers, nor older versions of Visual Studio, nor the current version of Visual Studio when targeting an older standard.

The issue links to a bug report for Visual Studio, where it looks like some bad compiler code generation is affecting a number of projects.

----

Given how specific and recent the issue is, I'd vote for merging as is, with the problem tests disabled (for now), and waiting for upstream projects to see if they can resolve the issue. Whether that comes from an update to GoogleTest, or an update to Visual Studio (which makes its way to AppVeyor), I'm not yet certain.

Given the above finding, and the successful test results on Linux and MacOS, I would be comfortable merging both this PR and PR #305.

ldicker83:
That pretty much sums up the issue I'm seeing during the crash when I run it locally. We'll keep an eye on this but in the mean time I'll go ahead and merge it.

cugone:
The visual studio team is aware of the issue and will be pushing a fix for it in 16.4.3 or 16.5

PR #329, opened 2020-01-09
Clean makefile

DanRStevens:
Linux only change.

Simplifies and eliminates old code for Docker images using Ubuntu 16.04.

Ubuntu 16.04 has an older version of GCC, which doesn't support needed C++17 features. As such, the code no longer compiles with the stock GCC compiler on Ubuntu 16.04. The old unmaintained Docker image for this build environment has been removed.


DanRStevens:
Not sure if anyone wants to review this, or if it should just be merged. I'm guessing other people aren't particularly keen to review `Makefile`s or `Dockerfile`s.

PR #331, opened 2020-01-09
Add `Point` and `Rectangle` conversion operator

### LINKED ISSUE #281: Point and Rectangle should have explicit conversion for underlying base type
The `Point` and `Rectangle` classes have a lot of overloaded functions to convert between an underlying type `int` and an underlying type of `float`. This in effect allows for implicit conversions, which may have unexpected rounding effects that could impact correctness, and also lead to hidden performance implications.

Rather than rely on what are effectively implicit conversions, there should exist explicit conversion methods. This would allow the caller to express conversion intent much more clearly.

----

As a first step, we could provide such explicit conversion functions, while making no other modifications that would affect backwards compatibility. Use of these explicit conversions would then be voluntary.

As a second step, the extra overloads that effectively do silent implicit conversions could be deprecated or removed. This would break old code. It would however make it easy to find code that should be updated, and the update itself would be fairly simple. Just add an explicit conversion call.

----

Related: Issue #267
If the common methods are templated, the conversion method could also be templated to convert to any other underlying type, using a default implementation of `static_cast`.

----

Old code:
```cpp
Point_2d pointInt{0, 0};
Point_2df pointFloat{0, 0};

// Implicitly trigger conversions (potentially bad)
if (pointInt == pointFloat) {
  // ...
}
```

New code:
```cpp
Point_2d pointInt{0, 0};
Point_2df pointFloat{0, 0};

// Explicitly convert type
if (pointInt.toFloat() == pointFloat) {
  // ...
}
```

New code (template version):
```cpp
Point_2d pointInt{0, 0};
Point_2df pointFloat{0, 0};

// Explicitly convert type
if (pointInt.to<float>() == pointFloat) {
  // ...
}
```

DanRStevens:
Closes #281

Uses a conversion operator to reduce the need for operator overloads.

### DIFF
--- a/src/Renderer/Rectangle.cpp
+++ b/src/Renderer/Rectangle.cpp
@@ -266,19 +266,6 @@ void Rectangle_2df::operator()(float _x, float _y, float _w, float _h)
 /**
  * Compares two rectangles.
  *
- * \param rect	Rectangle_2d to compare against.
- *
- * \returns True if the two rectangles are equivalent.
- */
-bool Rectangle_2df::operator==(const Rectangle_2d& rect)
-{
-	return (x() == rect.x()) && (y() == rect.y()) && (width() == rect.width()) && (height() == rect.height());
-}
-
-
-/**
- * Compares two rectangles.
- *
  * \param rect	Rectangle_2df to compare against.
  *
  * \returns True if the two rectangles are equivalent.
@@ -286,19 +273,6 @@ bool Rectangle_2df::operator==(const Rectangle_2d& rect)
 bool Rectangle_2df::operator==(const Rectangle_2df& rect)
 {
 	return (x() == rect.x()) && (y() == rect.y()) && (width() == rect.width()) && (height() == rect.height());
-}
-
-
-/**
- * Compares two rectangles.
- *
- * \param rect	Rectangle_2d to compare against.
- *
- * \returns True if the two rectangles are not equivalent.
- */
-bool Rectangle_2df::operator!=(const Rectangle_2d& rect)
-{
-	return (x() != rect.x()) || (y() != rect.y()) || (width() != rect.width()) || (height() != rect.height());
 }
 
 
--- a/src/Renderer/Rectangle.cpp
+++ b/src/Renderer/Rectangle.cpp
@@ -59,7 +59,7 @@ void Rectangle_2d::operator()(int _x, int _y, int _w, int _h)
  *
  * \returns True if the two rectangles are equivalent.
  */
-bool Rectangle_2d::operator==(const Rectangle_2d& rect)
+bool Rectangle_2d::operator==(const Rectangle_2d& rect) const
 {
 	return (x() == rect.x()) && (y() == rect.y()) && (width() == rect.width()) && (height() == rect.height());
 }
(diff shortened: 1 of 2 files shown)


src/Renderer/Rectangle.cpp
```diff
@@ -319,26 +279,24 @@ bool Rectangle_2df::operator==(const Rectangle_2df& rect)
...
-bool Rectangle_2df::operator!=(const Rectangle_2df& rect)
+Rectangle_2df::operator Rectangle_2d() const
 {
-	return (x() != rect.x()) || (y() != rect.y()) || (width() != rect.width()) || (height() != rect.height());
+	return Rectangle_2d{
+		static_cast<int>(mX),
+		static_cast<int>(mY),
+		static_cast<int>(mW),
+		static_cast<int>(mH)
+	};
```
cugone:
```
#include <cmath>
//...
return Rectangle_2d {
static_cast<int>(std::floor(mX))
static_cast<int>(std::floor(mY))
static_cast<int>(std::floor(mW))
static_cast<int>(std::floor(mH))
};
```
Floor float values before casting to int to prevent unintended behavior when dealing with negative values.
(You'll need to include `<cmath>`)

Note the graph and examples here; if you only truncate you get -2.0 instead of -3.0 when dealing with negative values.
Ref: https://en.cppreference.com/w/cpp/numeric/math/floor

DanRStevens: You know, I was tempted to leave that method out entirely, since the original didn't have auto conversion in that direction. I eventually decided to include it for symmetry, and because it seemed to fit with the idea of templatizing things later on.

----

I think delving into rounding modes is starting to get a bit subjective. I would not consider the proposed behaviour to be very intuitive as a default. It does seem potentially useful, and I could see us supporting that. I think though, due to the various forms of rounding, we would need to start using named methods to choose the proper rounding mode. The programmer could then explicitly choose which rounding mode they want.

```cpp
Rect<int> rectIntFloor = rectFloat.floor();
Rect<int> rectIntCeil = rectFloat.ceil();
Rect<int> rectIntRound = rectFloat.round();
Rect<int> rectIntTrunc = rectFloat.trunc();
```

Note that the standard library floating point rounding methods return floating point results. Hence the rounding method on the `Rect` may actually return a `Rect<float>` still, rather than a `Rect<int>`. In the above code, the assignment might then do an implicit cast from `Rect<float>` (of the temporary result) to `Rect<int>`.

For the default conversion operator, which can be triggered with `static_cast` of the struct, I think the most intuitive way to implement it, is to apply the `static_cast` to each field using the underlying types of the two structs. In particular, the reverse operation from `int` to `float` is implemented that way.

```cpp
// Applies `static_cast<float>` to each member field
Rect<float> rectFloat = static_cast<Rect<float>>(rectInt);
```

If both directions were consistent, then when we templatize the class, both directions can collapse into a single template method. That's one reason why I feel using a simple `static_cast` of each member field is most intuitive here.

```cpp
Rect<int> rectIntExplicitCast = static_cast<Rect<int>>(rectFloat); // Equivalent to truncation
Rect<int> rectIntImplicitConversion = rectFloat; // Equivalent to truncation
```

----

My suggestion here, is to remain with a simple `static_cast` for this operation. I would however quite welcome another issue to offer support for other rounding modes.

cugone: My original thought of "do as math does" should be the default "intuitive" implementation but then I found out apparently "doing what math does" is just as ambiguous!

Retracting my requested changes.

Ref: https://math.stackexchange.com/questions/3448/rules-for-rounding-positive-and-negative-numbers
Ref: https://en.wikipedia.org/wiki/Rounding

DanRStevens: Huh, that Wikipedia article looks like a pretty good summary of rounding.

Issue #337, opened 2020-01-13
Remove Point and Rectangle operator call


DanRStevens:
The `operator()` overload should be removed from `Point` and `Rectangle`.

The call operator is being overloaded as a quick way to set all values of these objects:
```cpp
someRect(newX, newY, newWidth, newHeight);
```

Although the code resembles object construction and initialization, it's a little unusual to do with an already existing object, and I've found the usage a bit surprising.

Resetting all values is perhaps better accomplished with the assignment operator:
```cpp
someRect = Rectangle_2d(newX, newY, newWidth, newHeight);
```

This would perhaps be more standard, more expected, and more clear.

As a curious side note, I tried checking if the call operator overload would be more efficient. A really minor point, I know, but I was curious. Turns out to not be the case. At high optimization settings, the two approaches compiled to identical code with GCC and Clang. At low or no optimization settings, the later, using the assignment operator, produced better code. I'm not sure why, other than perhaps the language offers more optional and mandatory optimizations that might apply to the assignment, which wouldn't be applied to other operators. Main point though, is the intention is perhaps a bit more clear.

One downside to the assignment code above, is the repetition of the type name. Repetition tends to add to maintenance costs, and should be avoided. In particular, the coming template update in issue #267, may involve a renamed base class. To get around that issue, we can omit the type name, and simply include values in braces:
```cpp
someRect = {newX, newY, newWidth, newHeight};
```

----

I propose we update downstream projects to use the brace syntax. It works right now, so the update can be done now. Then, we should update NAS2D to remove the call operator overload for the `Point` and `Rectangle` classes.

DanRStevens:
This has now been closed by a combination of PR #340 and PR #341.

PR #341, opened 2020-01-13
Remove Point `operator()`

DanRStevens:
This addresses the Point portion of issue #337. The corresponding Rectangle changes were done in PR #340.

~~It's recommended to merge the downstream OPHD changes first~~:  (Edit: Done).
ldicker83/OPHD#141

### DIFF
--- a/include/NAS2D/Renderer/Point.h
+++ b/include/NAS2D/Renderer/Point.h
@@ -27,8 +27,6 @@ public:
 	Point_2d(int x, int y);
 
 public:
-	void operator()(int x, int y);
-
 	bool operator==(const Point_2d& pt) const;
 	bool operator!=(const Point_2d& pt) const;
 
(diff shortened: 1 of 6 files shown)


DanRStevens:
Heh. For reference, I usually go commit by commit. Occasionally I'll start commenting on something that was changed later. It doesn't happen too often though, and I like the breakdown of being able to see things by individual commits.

Interesting comment about most vexing parse. It does look quite similar. Perhaps another reason to avoid that syntax here. Though since that was the call operator (`operator()`) on a pre-existing object, and not the constructor for a new object, I think it was safe. In particular, there is no type name to the left. At any rate, I compiled and ran unit tests at each step, and it didn't catch any problems. Your comment does get me thinking about it a bit deeper though.

cugone:
Ever since reading Scott Meyers' books and watching nearly all of his videos I've nearly universally switched to uniform initialization syntax (curly braces) unless absolutely necessary (I'm looking at you `std::vector`!), due in part to the MVP.

I realize now that yes, the code wouldn't have been affected because it wasn't an attempt to declare a variable; it just looked really close to a variable declaration.

PR #344, opened 2020-01-13
Rename `_Delegate` to `DelegateType`

### LINKED ISSUE #339: Replace _underscore prefixed names in Delegate code
Now that major Delegate and Signal changes have been merged, we should update to remove underscore prefixed names, such as `_Delegate`.

Related: PR #319 for changes to the rest of the code base.

DanRStevens:
Closes #339

Reference: See PR #319 for original rename thread.


DanRStevens:
Heh, you know, I was actually briefly considering `delegate_t`. I eventually decided against it though as it didn't match the naming convention used in the rest of the project. Plus, I didn't want the class to be confused as something coming from the standard library.

Or maybe it's just that I have a really strong bias against underscores. :wink:

cugone:
That's what the `std` namespace is for and why you should never use `using namespace std;` 😛

Issue #267, opened 2019-12-24
Use template for common code of Point and Rectangle


DanRStevens:
Point and Rectangle have specializations for both `int` and `float`. We should extract the commonality into a template base class to reduce repetition. Subclasses or `using` aliases can be used to generate the needed specializations with backwards compatible names.

As a bonus, users could also specialize for any type they wanted, such as `double`, or `long`.

DanRStevens:
This has now been closed by a combination of PR #345 and PR #346.

PR #345, opened 2020-01-14
Rectangle template

DanRStevens:
These are the `Rectangle` changes for issue #267.

----

I declared `extern template` types for the two existing types, and used explicit instantiation in the source file. This means the template only ever needs to be expanded for those two types once in a project (when compiling `Rectangle.cpp`), rather than in every file that uses them (by including `Rectangle.h`). I did that largely to keep behaviour consistent with how the compiler would have handled this code before. I don't know that it is necessarily needed though. All the member functions are pretty small and easy to inline, so that might be a better way to go. For all I know, that might even be how it does it anyway, since the methods being in the header file means they are all fully visible to the caller. Anyway, if we don't find it useful, we could remove the `extern template` declarations, and then remove the associated source file `Rectangle.cpp`. We can defer to a new PR if people want time to discuss or experiment with it.


include/NAS2D/Renderer/Rectangle.h
```diff
@@ -13,93 +13,109 @@
...
 
-struct Rectangle_2d;
-struct Rectangle_2df;
+template <typename BaseType>
+struct Rectangle
+{
+	Rectangle() = default;
+	Rectangle(const Rectangle& other) = default;
+	Rectangle& operator=(const Rectangle& other) = default;
+
```
cugone: Consider `=default`ing the move operations and the Dtor to adhere to Rule of Five/Six and [C.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five).

It's not necessarily a requirement (and I'm still not sure if moves of POD types aren't silently converted to copies anyway) but may be a missed opportunity for compiler optimization.

DanRStevens: Well, that got me thinking.

It doesn't seem to make much sense to move a POD type. Really what you'd want is to just copy it. There's no need to move from it, and update where it comes from. Effectively, there is no external resource being managed here. That's is largely why the special operations are all defaulted, rather than being user provided. The only reason they were explicitly defaulted, is because the constructor which takes initial values would have caused them to be deleted. And that only exists because we haven't yet transitioned this fully from a `class` to a `struct`, and we still have private fields.

According to the [The rule of three/five/zero](https://en.cppreference.com/w/cpp/language/rule_of_three), it talks about user defined special operations. None of those operations are user defined, only defaulted. It also talks about managing resources, which we don't do.

I would say the rule of three/five doesn't apply here.

----

We'll probably need to revisit the special member functions once we compete the transition and make the member fields public. My guess is, explicit mention of the special member functions will all disappear, along with the constructor taking initial values. The compiler should provide appropriate defaults at that time.

cugone: If we can get rid of *all* the default operations, i.e., the rule of zero, that would be even better! :)

PR #346, opened 2020-01-14
Point template

DanRStevens:
These are the `Point` changes for issue #267. See PR #345 for the associated `Rectangle` changes.

### DIFF
--- a/include/NAS2D/Renderer/Point.h
+++ b/include/NAS2D/Renderer/Point.h
@@ -89,79 +89,20 @@ private:
 };
 
 
-struct Point_2d;
-struct Point_2df;
-
-
 /**
  * \class	Point_2d
  * \brief	2D point.
  */
-struct Point_2d
-{
-public:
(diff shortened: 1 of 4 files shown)


include/NAS2D/Renderer/Point.h
```diff
@@ -12,79 +12,97 @@
...
 
-struct Point_2d;
-struct Point_2df;
+
+template <typename BaseType>
+struct Point {
+	Point() = default;
+	Point(const Point& point) = default;
+	Point& operator=(const Point& other) = default;
+
```
cugone: Consider `=default`ing the move operations and the Dtor to adhere to Rule of Five/Six and [C.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five).

It's not necessarily a requirement (and I'm still not sure if moves of POD types aren't silently converted to copies anyway) but may be a missed opportunity for compiler optimization.

PR #347, opened 2020-01-14
Remove always true condition flagged by Codacy

DanRStevens:
The `if` check above will already catch when `version` is `nullptr`.


cugone:
I have several false positives from Codacy in my own codebase that are very annoying. It's causing me to start to hate Codacy.

PR #348, opened 2020-01-14
Remove unused assignment flagged by Codacy

DanRStevens:
This variable is better scoped within the for loops.

### DIFF
--- a/src/Resources/Font.cpp
+++ b/src/Resources/Font.cpp
@@ -22,6 +22,7 @@
 #include <iostream>
 #include <math.h>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


cugone:
Since it never changes, declaring it `const` would be a good idea. Since it's an index into a collection, changing the type to `std::size_t` would be even better.

DanRStevens:
Good idea.

DanRStevens:
There's a stable line in between the header include adjustments, which gives enough context for the merge algorithm.

On another note, if both change sets make an identical change, that also doesn't cause a merge conflict.

PR #349, opened 2020-01-14
Avoid using C-includes

DanRStevens:
Should be `#include <cmath>` (C++) rather than `#include <math.h>` (C).

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -26,7 +26,6 @@
 #include <SDL_image.h>
 
 #include <iostream>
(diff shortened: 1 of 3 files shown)


cugone:
I was just about to mention this. It appears in other places as well:

 - `RendererOpenGL.cpp`
 - `Font.cpp`
 - `XMLParser.cpp` (These are other C-includes, not `math.h` specifically)

DanRStevens:
Oh geeze, seems my grep missed a few.

cugone:
Visual Studios' Find in Files Regular expressions and "Search Entire Solution" ftw.

Issue #235, opened 2019-12-20
Rename `isRectInRect`


DanRStevens:
The method `isRectInRect` would seem to imply containment, as in one rect is entirely contained in the other. In actuality, the method tests if two rects overlap at all. The method should be renamed to reflect that.

Related, the method test overlap in an endpoint inclusive way. Looking at method usage, it seems endpoint exclusive might be more appropriate to how it is used. The exclusive way may also be more common and expected by end users. When the method is renamed, we should give some consideration to the exact definition of what it does, and if any behaviour update is required.

My suggestion is to transition it to endpoint exclusive. That may require some updates to the test cases.

cugone:
It can be renamed `DoRectsOverlap` and a separate function `Contains` can be added if needed.

ldicker83:
Honestly, it was intended as a basic intersect test. It could be moved into the Rectangle_2d and Rectangle_2df classes as Rectangle_2d/f::intersects(const Rectangle_2d/f& r). Would actually simplify syntax in a number of areas.

DanRStevens:
I was thinking similar. We could easily offer member functions for both `intersects` and `contains`. The later being more sensible as a member function, as the relationship is not symmetric, so a free form function would not make it obvious which rect contains the other. With `intersects` the relationship is symmetric, so would be clear as either a member function or a free standing function.

PR #351, opened 2020-01-15
Add Rectangle `contains` and `overlaps` methods

DanRStevens:
This is support work for issue #235.

Note there is a change in specification between these methods and the existing methods from `MathUtils`. The original methods were inclusive of both start point and endpoint. The new methods are start point inclusive, and end point exclusive. I believe the end point exclusive behaviour is more correct for most uses of the `overlaps` test. In particular, the new methods are consistent with 0-based array indexing, and are suitable for testing overlap of images with dimensions given in pixels.

### DIFF
--- a/include/NAS2D/Renderer/Rectangle.h
+++ b/include/NAS2D/Renderer/Rectangle.h
@@ -10,6 +10,9 @@
 
 #pragma once
 
+#include "Point.h"
+
+
 namespace NAS2D {
 
 
@@ -46,6 +49,14 @@ struct Rectangle
 			static_cast<NewBaseType>(mW),
 			static_cast<NewBaseType>(mH)
 		};
+	}
+
+	// Start point inclusive (x, y), endpoint exclusive (x + width, y + height)
+	// Area in interval notation: [x .. x + width), [y .. y + height)
+	bool contains(const Point<BaseType>& point) const {
+		auto px = point.x();
+		auto py = point.y();
+		return ((mX <= px) && (px < mX + mW)) && ((mY <= py) && (py < mY + mH));
 	}
 
 	void x(BaseType x) {
--- a/include/NAS2D/Renderer/Rectangle.h
+++ b/include/NAS2D/Renderer/Rectangle.h
@@ -59,6 +59,12 @@ struct Rectangle
 		return ((mX <= px) && (px < mX + mW)) && ((mY <= py) && (py < mY + mH));
 	}
 
+	// Start point inclusive (x, y), endpoint exclusive (x + width, y + height)
+	// Area in interval notation: [x .. x + width), [y .. y + height)
+	bool overlaps(const Rectangle& rect) const {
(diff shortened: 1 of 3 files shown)


cugone:
For a sufficiently small world size such that the calculated viewport causes the world units to be noticeably large relative to the screen (think tile-based games where the tiles are dozens to hundreds of pixels in dimension) the tests will mathematically pass but visually fail when testing a collision against the right sides by as much as half the rectangle. Two tenets I generally follow when dealing with perceived bugs:

 - Is the player going to notice?
 - Is the player going to care?

I think in this case, the player will notice and care.

DanRStevens:
Doh! I hate having to update the Visual Studio project files. I kind of wish they could just include all `*.cpp` and `*.h` in a subtree and be done with it. I believe the project file syntax actually supports that, though it seems to be discouraged for some reason.

----

The test code folder layout should match the folder layout of the source. A better question is why are `Point` and `Rectangle` in the `Renderer` folder. I kind of think they could be more general than that.

----

I'm not sure what you're trying to say about world views. Let me illustrate with an example though:
Screen size is 640 x 480.
X ranges from 0 .. 639.
Y ranges from 0 .. 479.
If you draw a pixel at (640, 480), it is off the screen.

DanRStevens:
To add to that example, if the rect is using tile coordinates, with tiles of size 32 x 32, then you might assume a viewport of size 640 x 480 can show:
640 / 32 = 20 tiles (X direction)
480 / 32 = 15 tiles (Y direction)
However, that is only true when the viewport dimension is an integral multiple of the tile size, and you're using integer tile scrolling.

For non-integral viewport dimensions, you need to round up when dividing.

For smoother scrolling, you need to account for the maximum offset of any given tile. In the above example, if the view offset is (1, 1), then you still need to show 31 x 31 pixels of the first tile, and 1x1 pixel of the last tile, which is now beyond the previously displayed last tile. In that case, the range of displayed tiles must be increased by 1 in each dimension.

Accounting for smooth scrolling and viewport dimensions that are non-integer multiples of the tile size, then you need to account for both the viewport dimension and a maximum offset within a tile:
```
maxWidthInTiles = divideRoundUp(viewport.width + tileWidth - 1, tileWidth)
maxHeightInTiles = divideRoundUp(viewport.height + tileHeight - 1, tileHeight)
```

I believe that works out to:
```
maxWidthInTiles = (viewport.width + (tileWidth - 1) * 2) / tileWidth
maxHeightInTiles = (viewport.height + (tileHeight - 1) * 2) / tileHeight
```

----

Of course, the calculation given above is a different concern from bounding box clipping, which `overlaps` is concerned with. In particular, given the above calculation, the viewport area in tiles might be given by some sort of rect:
```
viewportInTilesRect{Point{tileX, tileY}, Vector{maxWidthInTiles, maxHeightInTiles}};
```
In which case, testing for overlap, you still want end point exclusive comparisons.

DanRStevens:
Well, that was an interesting exercise.

Issue #306, opened 2020-01-03
Add Point and Vector component operations to Rectangle


ldicker83:
From #303: `Rectangle` should provide ways to get corner `Point`s, or a size `Vector`. From there, arithmetic can be done on the `Point` and `Vector` components, which can then be used to create a new `Rectangle`.

As part of that, `Rectangle` should be constructible from either two `Point`s, or from a `Point` and a `Vector` (size).

_Originally posted by @DanRStevens in https://github.com/lairworks/nas2d-core/issues/303#issuecomment-570127304_

As a note, the title of this issue is very poor but I'm tired and can't think of a better way to name this.

cugone:
We don't have a `Vector` class yet, but constructing from two `Point_2d` or `Point_2df` can be done as `mins` and `maxs` (i.e. the smallest x/y and largest x/y) then derive the `mW` and `mH` values to reduce confusion with `Rectangle(x, y, w, h)`.

DanRStevens:
I'm a little unclear what you mean by `mins` and `maxs`, though typically a `Vector` is defined by `Point` subtraction:
```cpp
Vector vector = p2 - p1;
```

This implies the `Point` class has an `operator-` that looks something like:
```cpp
Vector Point::operator-(Point rhs);
```

cugone:
`mins`: The smallest x value and smallest y value pair contained by the rectangle i.e. a corner of the rectangle.
`maxs`: The largest x value and largest y value pair contained by the rectangle i.e. the opposite corner of the rectangle.

I have to be deliberately vague because "smallest" by default according to Allegro 5/SDL2/windows is towards the upper left of the screen but to OpenGL it's the lower left and DirectX/Vulkan is whatever you want it to be when you set up your viewports. Similar but different for "largest".

DanRStevens:
Ahh. Yes. I'd have been tempted to name them something like `BottomLeft` and `TopRight`. That would be context specific though. I suppose more general names would be in order.

I would also be tempted to number the points, though that's not very descriptive.

PR #352, opened 2020-01-15
Add Vector template class

DanRStevens:
This is to support issue #306.

### DIFF
--- a/include/NAS2D/Renderer/Vector.h
+++ b/include/NAS2D/Renderer/Vector.h
@@ -1,0 +1,52 @@
+#pragma once
+
+namespace NAS2D {
+
+
+template <typename BaseType>
+struct Vector {
+	BaseType x = 0;
+	BaseType y = 0;
+
+	bool operator==(const Vector& vector) const {
+		return (x == vector.x) && (y == vector.y);
+	}
+	bool operator!=(const Vector& vector) const {
+		return !(*this == vector);
+	}
+
+	Vector& operator+=(const Vector& vector) {
+		x += vector.x;
+		y += vector.y;
+		return *this;
+	}
+	Vector& operator-=(const Vector& vector) {
+		x -= vector.x;
+		y -= vector.y;
+		return *this;
+	}
+
+	Vector operator+(const Vector& vector) const {
+		return {
+			x + vector.x,
+			y + vector.y
+		};
+	}
+	Vector operator-(const Vector& vector) const {
+		return {
+			x - vector.x,
+			y - vector.y
+		};
+	}
+
+	template <typename NewBaseType>
+	operator Vector<NewBaseType>() const {
+		return {
+			static_cast<NewBaseType>(x),
+			static_cast<NewBaseType>(y)
+		};
+	}
+
+};
+
+}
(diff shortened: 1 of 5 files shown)


proj/vs2019/NAS2D.vcxproj
```diff
@@ -249,6 +249,7 @@
     <ClInclude Include="..\..\include\NAS2D\Renderer\RendererNull.h" />
     <ClInclude Include="..\..\include\NAS2D\Renderer\Color.h" />
     <ClInclude Include="..\..\include\NAS2D\Renderer\Point.h" />
+    <ClInclude Include="..\..\include\NAS2D\Renderer\Vector.h" />
```
cugone: This shouldn't be in the Renderer folder. It's a math concept and would make more sense in a `Math` subfolder.

proj/vs2019/NAS2D.vcxproj.filters
```diff
@@ -236,6 +236,9 @@
     <ClInclude Include="..\..\include\NAS2D\Renderer\Point.h">
       <Filter>Header Files\Renderer</Filter>
     </ClInclude>
+    <ClInclude Include="..\..\include\NAS2D\Renderer\Vector.h">
+      <Filter>Header Files\Renderer</Filter>
```
cugone: Ditto

test/test.vcxproj
```diff
@@ -48,6 +48,7 @@
   </PropertyGroup>
   <ItemGroup>
     <ClCompile Include="Renderer/Rectangle.test.cpp" />
+    <ClCompile Include="Renderer/Vector.test.cpp" />
```
cugone: Ditto

include/NAS2D/Renderer/Vector.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+namespace NAS2D {
+
+
+template <typename BaseType>
+struct Vector {
+	BaseType x = 0;
+	BaseType y = 0;
+
```
cugone: No support for 3D or 4D vectors? How is matrix math going to be handled?

proj/vs2019/NAS2D.vcxproj
DanRStevens: Agreed, though I think we need to address that for all geometries at the same time, in a separate PR.

include/NAS2D/Renderer/Vector.h
DanRStevens: That's not a concern at this time. The class was modelled to match existing `Point` and `Rectangle` classes. We could do this at some point, but it likely needs to involve all classes at the same time. We should defer until after the other classes have been switched over to proper structs.

proj/vs2019/NAS2D.vcxproj
cugone: Some PR has to make the Math folder, why not start with this one? It'll make adding more later much easier.

DanRStevens: There are inter-dependencies between the headers. Either relative includes need to be updated, which is messy, or they all move at once.

The `Vector` class is strongly related to the `Point` class and the `Rectangle` class. It should go where those are.

I would really like to move this, I just don't think this is the PR to do it in.

cugone: 'K. :)

DanRStevens: Opened issue #358 to track.

PR #353, opened 2020-01-15
Update vcpkg install rule

DanRStevens:
This should be relative to the project file, rather than the solution. That way the project can be embedded in other solution files, without the relative path being wrong.

Edit:
The AppVeyor build of downstream project op2-landlord is broken due to not finding `InstallVcpkgDeps.bat`. There is no AppVeyor build for OPHD.

How does this affect people's local builds? I assume it would have failed locally as well? Or is the error not noticed because packages are already installed locally?


cugone:
The batch file runs and I get no errors because everything reports as already installed.

PR #355, opened 2020-01-15
Add Rectangle factory functions

DanRStevens:
Based on PR #352 (Add `Vector`). (Only the last commit is new).

Adds factory methods to create a `Rectangle` from either a `Point` and a `Vector`, or from two `Point`s.

### DIFF
--- a/include/NAS2D/Renderer/Rectangle.h
+++ b/include/NAS2D/Renderer/Rectangle.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include "Point.h"
+#include "Vector.h"
 
 
 namespace NAS2D {
@@ -29,6 +30,26 @@ struct Rectangle
 		mW(width),
 		mH(height)
 	{}
+
+	// Factory method
+	static Rectangle<BaseType> Create(Point<BaseType> startPoint, Vector<BaseType> size) {
+		return {
+			startPoint.x(),
+			startPoint.y(),
+			size.x,
+			size.y
+		};
+	}
+
+	// Factory method
+	static Rectangle<BaseType> Create(Point<BaseType> startPoint, Point<BaseType> endPoint) {
+		return {
+			startPoint.x(),
+			startPoint.y(),
+			endPoint.x() - startPoint.x(),
+			endPoint.y() - startPoint.y()
+		};
+	}
 
 	bool operator==(const Rectangle& rect) const {
 		return (mX == rect.mX) && (mY == rect.mY) && (mW == rect.mW) && (mH == rect.mH);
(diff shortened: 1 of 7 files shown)


include/NAS2D/Renderer/Rectangle.h
```diff
@@ -30,6 +31,26 @@ struct Rectangle
 		mH(height)
 	{}
 
+	// Factory method
+	static Rectangle<BaseType> Create(Point<BaseType> startPoint, Vector<BaseType> size) {
```
cugone: `startPoint` is ambiguous. Is it a corner? The center? Why not make it a `Vector`?

DanRStevens: A `Rectangle` would not be well defined by two `Vector` objects. It needs either two corner `Point` objects, or it  needs one corner `Point` object, and a `Vector` for the size. The other corner point is then `startPoint + vector`.

And yes, `startPoint` is a corner, which has the minimum `(x, y)` value. Wasn't sure what to call it.

cugone: I suggest renaming `startPoint` to remove external ambiguity.

DanRStevens: Yes, but do you have a suggestion?

cugone: The easiest is to name is `center`, `size` should be `radius` or `halfExtents` and then internally derive the resulting `Rectangle`:

```
static Rectangle<BaseType> Create(Point<BaseType> center, Vector<BaseType> halfExtents) {
    return {
        center.x - halfExtents.x(),
        center.y - halfExtents.y(), //Assuming topleft origin.
        halfExtents.x() * static_cast<BaseType>(2.0),
        halfExtents.y() * static_cast<BaseType>(2.0),
    };
}
```

DanRStevens: Well, I wouldn't mind having additional overloads, such as being able to create from a center point. That is something I would like to support eventually.

In this case though, the intention is to create from a corner point and total size along each side. There are many use cases where a corner point is easily known, but a center point would need to be calculated. Plus the corner point and size representation closely matches the internal representation, so if that's what's already available, there should be a quick way to create a `Rectangle` from that information, without needing to convert to some intermediate representation and then back again.

----

As for the parameter names `startPoint` and `size`, they do have a certain symmetry to the other factory method which uses the names `startPoint` and `endPoint`.

----

I would advise against using `radius` for something that isn't a circle. I find that pretty confusing. My first thought there is either a square inscribed in a circle, or a square which forms the bounding box of a circle.

cugone: Looking back on it, `startPoint` and `endPoint` are just synonyms for `mins` and `maxs`. I'm okay with this. Just not the internal representation as "corner and size". 😝

Issue #320, opened 2020-01-06
Remove `Filesystem` auto mounting


DanRStevens:
We should remove auto mounting of folders from the `Filesystem` constructor. Not all programs will need the same set of real filesystem folders mounted into their virtual filesystem. To improve flexibility, and allow support for more general use cases, we should strive for a more flexible design. As it may be hard to summarize all desired mount points in constructor parameters, perhaps mounting should be decoupled from the constructor.

A convenience method can be offered to setup default mounting that matches previous behaviour. Potentially this method could return a self reference to the object, allowing it to be easily chained with object construction. This would allow for easy default initialization, that still only takes one line of code:
```cpp
Filesystem& f = Utility<Filesystem>::init<Filesystem>(...).mountDefault();
```

----

Related:
#315: Double mounting of files (remove data folder auto mount)
#317: Filesystem special folders (real filesystem paths, allow for easy custom mounting)
#318: Filesystem `unmount` (flexibility, and stop gap adjustment of default mounts)
This issue: Remove auto mounting of all folders

DanRStevens:
I just realized the `Filesystem` interface has no way to set a write folder. The `mount` method gives read-only access to the mounted folders.

Before we can remove the auto mounting, which sets up a write folder, we first need an alternate way to set a write folder.

With PhysFS, there is only one write folder. Setting a new one replaces the old one. Any mounted folders also always map to the root of the filesystem. More generally, we could allow write access to multiple folders. Perhaps we should offer an alternate `mountWrite` method, that provides write access to a folder. Eventually the `mount` methods could be updated to provide a virtual filesystem path they are mounted to, with a default equal to the root of the virtual filesystem, which would match the current PhysFS behaviour.

If two write folders are mounted at the same path, the later one should take precedence. That would make sense for a module or extension system, where later loaded resources take precedence over the main default resources. It would also match current PhysFS behaviour, where setting a new write folder replaces the setting for the old write folder.

When mounting a folder for write access, we should probably also mount it for read access. I believe PhysFS allows more generality there, though I don't see much reason in using it. It would certainly make the API more convenient if mounting a write folder also gave read access to it. I also believe the mounted to path within the virtual filesystem should be the same for both read and write, should we eventually support mounting to non root locations within the virtual filesystem.

Issue #315, opened 2020-01-06
Filesystem double mounting


DanRStevens:
The `Filesystem` class is designed such that it double mounts some files, giving them two distinct paths in the virtual filesystem.

The cause of the double mounting is that we are using PhysFS to mount both a folder and it's subfolder. The default "sane" config provided by PhysFS allows for mounting of the executable's folder. We also offer a parameter for a data folder. This is typically a subfolder of the executable's folder. Once both are mounted, any files in the data folder are now visible through two distinct paths:
- `data/file.txt` (visible through exeFolder mount)
- `file.txt` (visible through exeFolder + dataFolder mount)

This seems to be a flaw in the design of `Filesystem`, as the constructor parameters encourage such behaviour. This double mounting basically happens by default. I believe we need an API change involving the constructor, such that it does not encourage double mounting.

----

One way we can remove default double mounting is to simply remove the data folder parameter. The `Filesystem` object could then by default just mount the executable's folder. This would give a single data path:
- `data/file.txt`

Such as change would impact downstream software. They would need to update all resource paths to explicitly include the data folder.

If it is too onerous for a software package to update all paths to explicitly include the data folder, they could manually mount the data folder after object construction, allowing for shorter paths (`file.txt`). This of course would cause the same double mounting, though would be under the control of the calling code, rather than baked into the library.

----

Another option is to remove default mounting of the executable's folder. I'll open another issue shortly with a longer term roadmap for `Filesystem` changes that discusses that.

DanRStevens:
To add to this, while working on PR #357, I noticed another reason for the double mounting flaw in the API design. The `Filesystem::mount` method prepends the `dataPath` to any path being mounted. As `dataPath` should already be mounted, this limits the `mount` operation to only providing duplicate views to already accessible files, just with shorter paths.

We should remove pre-pending of any paths when mounting. In particular, the mount paths might be real filesystem absolute paths, so should be have anything prepended to them. In fact, they quite likely will be absolute paths.

In terms of #357, the `unmount` method was designed to also prepend `dataPath`, since it must necessarily match how `mount` works. Both methods should be updated together.

PR #357, opened 2020-01-15
Add `Filesystem::unmount`

### LINKED ISSUE #318: Provide `Filesystem::unmount` method
We should add an `unmount` method to the `Filesystem` class. The method should take real filesystem paths as an argument.

Although generally useful, this can also be used in conjunction with issue #317 to unmount default mounted folders. That would provide a stop gap solution for adjusting the default set of mounted folders, should a program need something other than the default.

DanRStevens:
Closes #318

### DIFF
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -38,6 +38,7 @@ public:
 	std::string workingPath(const std::string& filename) const;
 	StringList searchPath() const;
 	void mount(const std::string& path) const;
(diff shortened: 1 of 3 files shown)


DanRStevens:
I know this may feel a little backwards to be adding more PhysFS code at this point, though I'm sure it's on the path forward. This should allow downstream projects to adjust easier while we make other planned `Filesystem` changes, such as #315 and #320.

PR #359, opened 2020-01-16
Add Point and Vector component operations to Rectangle

### LINKED ISSUE #306: Add Point and Vector component operations to Rectangle

DanRStevens:
Closes #306

### DIFF
--- a/include/NAS2D/Renderer/Rectangle.h
+++ b/include/NAS2D/Renderer/Rectangle.h
@@ -58,6 +58,18 @@ struct Rectangle
 		return !(*this == rect);
 	}
 
+	Vector<BaseType> size() const {
+		return {mW, mH};
+	}
+
+	Point<BaseType> startPoint() const {
+		return {mX, mY};
+	}
+
+	Point<BaseType> endPoint() const {
+		return Point<BaseType>{mX, mY} + Vector<BaseType>{mW, mH};
+	}
+
 	bool null() const {
 		return (mW == 0) || (mH == 0);
 	}
--- a/include/NAS2D/Renderer/Point.h
+++ b/include/NAS2D/Renderer/Point.h
@@ -10,6 +10,9 @@
 
 #pragma once
 
+#include "Vector.h"
+
+
 namespace NAS2D {
 
 
@@ -29,6 +32,20 @@ struct Point {
 	}
 	bool operator!=(const Point& point) const {
 		return !(*this == point);
+	}
+
+	Point& operator+=(const Vector<BaseType>& vector) {
+		mX += vector.x;
+		mY += vector.y;
+		return *this;
+	}
+
+	Point operator+(const Vector<BaseType>& vector) const {
+		return {mX + vector.x, mY + vector.y};
+	}
+
+	Vector<BaseType> operator-(const Point& point) const {
+		return {mX - point.mX, mY - point.mY};
 	}
 
 	template <typename NewBaseType>
--- a/test/Renderer/Point.test.cpp
+++ b/test/Renderer/Point.test.cpp
@@ -14,6 +14,20 @@ TEST(Point, OperatorEqualNotEqual) {
 	EXPECT_NE((NAS2D::Point<int>{1, 1}), (NAS2D::Point<int>{2, 1}));
 }
 
+TEST(Point, SelfAddVector) {
+	NAS2D::Point<int> point{1, 1};
+	EXPECT_EQ(&point, &(point += NAS2D::Vector<int>{1, 2}));
+	EXPECT_EQ((NAS2D::Point<int>{2, 3}), point);
+}
+
+TEST(Vector, AddVector) {
+	EXPECT_EQ((NAS2D::Point<int>{2, 3}), (NAS2D::Point<int>{1, 1}) + (NAS2D::Vector<int>{1, 2}));
+}
+
+TEST(Vector, SubtractPointToVector) {
+	EXPECT_EQ((NAS2D::Vector<int>{1, 1}), (NAS2D::Point<int>{2, 3}) - (NAS2D::Point<int>{1, 2}));
+}
+
 TEST(Point, Conversion) {
 	// Allow explicit conversion
 	EXPECT_EQ((NAS2D::Point<int>{1, 1}), static_cast<NAS2D::Point<int>>(NAS2D::Point<float>{1.0, 1.0}));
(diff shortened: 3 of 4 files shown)


include/NAS2D/Renderer/Rectangle.h
```diff
@@ -58,6 +58,18 @@ struct Rectangle
...
+	}
+
+	Point<BaseType> startPoint() const {
+		return {mX, mY};
+	}
+
+	Point<BaseType> endPoint() const {
+		return Point<BaseType>{mX, mY} + Vector<BaseType>{mW, mH};
+	}
+
```
cugone: Adding a `center` function that returns a vector would be nice. Though, I'm not a fan of division; multiplication is always faster, it can't be helped due to the nature of not knowing the underlying type without adding complex template metaprogramming:
```
Vector<BaseType> center() const noexcept {
    return Vector<BaseType>(mX,mY) + Vector<BaseType>(mW,mH) / static_cast<BaseType>(2.0));
}
```

DanRStevens: We should definitely have a method to return the `center`. I suppose it would be a `Point` though.

I'm not too certain of the best way to implement off the top of my head. We should probably defer that to an issue. Actually, since templates are compile time, the compiler should be able to optimize, so just writing `size / 2` should work reasonably well.

Oh, and that reminds me, we still need to implement scalar multiplication and division for `Vector`.

DanRStevens: Opened issues #361 and #360 to track.

cugone: Given we have a `Vector` class now, what's the difference between it and `Point`? It seems to me, if you have a `DrawPoint(Vector point)` function on the renderer, they end up being duplicative.

DanRStevens: They are highly related, though distinct mathematical constructs, with highly similar yet different sets of operations. Conceptually a `Point` represents an a position, while `Vector` represents a direction.

I have been tempted to create a base class to store the common elements, though I'm not sure that would be a good design choice, since one isn't derived from the other.

cugone: Traditionally a vector can be both though. Non-normalized, it represents a point as the distance from the origin; normalized, it represents a direction.

By calling this implementation a vector you will confuse users who expect it to behave like a mathematical vector and not *just* a direction.

DanRStevens: Both a `Point` and a `Vector` have the same components, which makes them easily confused. They represent different things though, and so support different operations. Think about the difference between an odometer and a speedometer.

Point like objects: distance (1D), position (2D, 3D), point in time (1D)
Vector like objects: speed (1D), velocity (2D, 3D), duration of time (1D)

If you want to represent the difference mathematically (though you wouldn't do this in an implementation for efficiency reasons), you can extend both of them to have an extra coordinate. For `Vector`s, that extra coordinate will be `0`. For `Point`s, that extra coordinate will be `1`. If you extend the basic operations to now include that extra coordinate, you can see how identity is transformed:
- Addition:
  -  `(x1, y1, 0) + (x2, y2, 0) = (x1 + x2, y1 + y2, 0)` Vector + Vector = Vector
  -  `(x1, y1, 0) + (x2, y2, 1) = (x1 + x2, y1 + y2, 1)` Vector + Point = Point
  -  `(x1, y1, 1) + (x2, y2, 0) = (x1 + x2, y1 + y2, 1)` Point + Vector = Point
  -  `(x1, y1, 1) + (x2, y2, 1) = (x1 + x2, y1 + y2, 2)` Point + Point = Nonsense
- Subtraction:
  -  `(x1, y1, 0) - (x2, y2, 0) = (x1 - x2, y1 - y2, 0)` Vector - Vector = Vector
  -  `(x1, y1, 0) - (x2, y2, 1) = (x1 - x2, y1 - y2, -1)` Vector - Point = Nonsense
  -  `(x1, y1, 1) - (x2, y2, 0) = (x1 - x2, y1 - y2, 1)` Point - Vector = Point
  -  `(x1, y1, 1) - (x2, y2, 1) = (x1 - x2, y1 - y2, 0)` Point - Point = Vector
- Scalar Multiplication:
  -  `(x1, y1, 0) * a = (x1 * a, y1 * a, 0)` Vector * scalar = Vector
  -  `(x1, y1, 1) * a = (x1 * a, y1 * a, a)` Point * scalar = Nonsense (for a != 0, 1)
- Scalar Division:
  -  `(x1, y1, 0) / a = (x1 / a, y1 / a, 0)` Vector / scalar = Vector (for a != 0)
  -  `(x1, y1, 1) / a = (x1 / a, y1 / a, 1/a)` Point / scalar = Nonsense (for a != 1)
- Average:
  -  `((x1, y1, 0) + (x2, y2, 0)) / 2 = ((x1 + x2) / 2, (y1 + y2) / 2, 0)`
 (Vector + Vector) / 2 = Vector
  -  `((x1, y1, 0) + (x2, y2, 1)) / 2 = ((x1 + x2) / 2, (y1 + y2) / 2, 1/2)`
 (Vector + Point) / 2 = Nonsense
  -  `((x1, y1, 1) + (x2, y2, 0)) / 2 = ((x1 + x2) / 2, (y1 + y2) / 2, 1/2)`
 (Point + Vector) / 2 = Nonsense
  -  `((x1, y1, 1) + (x2, y2, 1)) / 2 = ((x1 + x2) / 2, (y1 + y2) / 2, 1)`
 (Point + Point) / 2 = Point

Of course you wouldn't bother wasting memory to store that extra coordinate in practice. You'd just use the type system to annotate things, and have it all checked at compile time.

cugone: Yes you would need to store that extra coordinate. It's called the homogeneous coordinate and is used in matrix calculations. By using `Vector` in the correct way, you don't need both a `Point` and `Vector` type.

Recommended viewing: https://www.youtube.com/watch?v=o1n02xKP138

PR #363, opened 2020-01-16
Add Vector scalar operations

### LINKED ISSUE #360: Add Vector scalar multiply and divide
The `Vector` class should be given scalar multiply and divide operations:
- `Vector& operator*=(BaseType multiplier)`
- `Vector& operator/=(BaseType divisor)`
- `Vector operator*(BaseType multiplier) const`
- `Vector operator/(BaseType divisor) const`

This would be useful for scaling, direction reversal, linear interpolation, taking averages, or finding center points.

DanRStevens:
Closes #360

### DIFF
--- a/include/NAS2D/Renderer/Vector.h
+++ b/include/NAS2D/Renderer/Vector.h
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <stdexcept>
+
+
 namespace NAS2D {
 
 
@@ -37,6 +40,30 @@ struct Vector {
 			x - vector.x,
 			y - vector.y
 		};
+	}
+
+	Vector& operator*=(BaseType scalar) {
+		x *= scalar;
+		y *= scalar;
+		return *this;
+	}
+	Vector& operator/=(BaseType scalar) {
+		if (scalar == 0) {
+			throw std::domain_error("Cannot divide vector by 0");
+		}
+		x /= scalar;
+		y /= scalar;
+		return *this;
+	}
+
+	Vector operator*(BaseType scalar) const {
+		return {x * scalar, y * scalar};
+	}
+	Vector operator/(BaseType scalar) const {
+		if (scalar == 0) {
+			throw std::domain_error("Cannot divide vector by 0");
+		}
+		return {x / scalar, y / scalar};
 	}
 
 	template <typename NewBaseType>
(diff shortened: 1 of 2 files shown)


include/NAS2D/Renderer/Vector.h
```diff
@@ -39,6 +42,30 @@ struct Vector {
...
 	}
 
+	Vector& operator*=(BaseType scalar) {
+		x *= scalar;
+		y *= scalar;
+		return *this;
+	}
+	Vector& operator/=(BaseType scalar) {
+		if (scalar == 0) {
+			throw std::domain_error("Cannot divide vector by 0");
```
cugone: This `if` check isn't needed and certainly not the custom throw. Throwing on divide by zero is built in.

include/NAS2D/Renderer/Vector.h
```diff
@@ -39,6 +42,30 @@ struct Vector {
...
+		y /= scalar;
+		return *this;
+	}
+
+	Vector operator*(BaseType scalar) const {
+		return {x * scalar, y * scalar};
+	}
+	Vector operator/(BaseType scalar) const {
+		if (scalar == 0) {
+			throw std::domain_error("Cannot divide vector by 0");
```
cugone: Ditto.

include/NAS2D/Renderer/Vector.h
DanRStevens: Actually, no, it's undefined behaviour. For integer operations pretty much anything can happen. For floating point, there is generally some hardware defined behaviour to produce infinity values.

----

Though I realize now in hindsight that I forgot to add checks for divide by 0 to the unit tests.

DanRStevens:
Ok, I fixed the missing test case. If you remove the `throw`, the new test code will crash with a core dump. In particular, it does not trigger C++ exception handling, which would have been reported by GoogleTest, and the tests would have continued.

cugone:
Conveniently, MSVC doesn't throw under any build, Debug or otherwise, it just marks the value as `inf` and continues on its merry way. Great. 😖

DanRStevens:
I assume you were testing with floating point for that result. I'm guessing testing with integers would lead to a crash on MSVC, like it does with Clang.

This is one of those dark corners of the language, where you may find yourself wishing the language behaved differently than it does.

cugone:
> 
> 
> I assume you were testing with floating point for that result. I'm guessing testing with integers would lead to a crash on MSVC, like it does with Clang.

Yes integers cause a dump.

PR #364, opened 2020-01-16
Add graphics libraries to MacOS CircleCI config

DanRStevens:
This is in preparing to adding unit test code for NAS2D features that depend on the graphics libraries.

Currently, no NAS2D features using the graphics libraries are tested. That means the linker will cut out any code that references the graphics libraries, and effectively prunes those dependencies. As we extend unit testing, this will no longer be true. As such, we'll need a proper set of static libraries during the link step to compile the unit test project.

Further, we'll need access to runtime dynamic link libraries for the unit test executable to load and run. This is covered by updates to the `brew install`/`brew link` step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ jobs:
     executor: macos-executor
     steps:
       - brew-install:
-          packages: physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf glew cmake
+          packages: physfs sdl2 sdl2_image sdl2_mixer sdl2_ttf libpng libjpeg libtiff webp libmodplug libvorbis libogg freetype glew cmake
       - build-googletest
       - checkout
       - build-and-test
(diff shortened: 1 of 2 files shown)


DanRStevens:
Well, I don't think anyone else is too enthusiastic about maintaining the MacOS build, or Makefile updates, and the CI builds are passing, so I'm just going to go ahead and merge this.

If anyone wants to take the time to review, they are quite welcome to.

cugone:
Unfortunately I agree.

Game development for non-Windows/non-console/non-mobile platforms is pretty much dead. The defacto default is Windows at 98% and MacOS at a distant 2% and Linux less than 1%.

Ref (expand the OS Version row): https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam
Ref: https://www.statista.com/statistics/265033/proportion-of-operating-systems-used-on-the-online-gaming-platform-steam/

DanRStevens:
Hmm, some interesting stats on there.

Back when I used Windows, I remember my brother using Steam. I never got into it though, and after switching to Linux I rarely ever heard about it. I suppose if I wanted to install something on Ubuntu, I would just use the Ubuntu store. Most things in there are free, though they do have some paid stuff. Integrates well with the system.

... though I'm guessing if the Ubuntu store released any stats, they'd probably claim that 100% of their game downloads are for Linux. :wink:

PR #366, opened 2020-01-19
Cleanup vs project files

DanRStevens:
I noticed some empty element cruft accumulating in the project files. I figure this should be safe to remove, and won't come back automatically. I would like to verify it doesn't affect Windows developers though.

I suspect those elements were added when project settings were modified, saved, then the project settings were deleted, and the elements remained but became empty.

### DIFF
--- a/proj/vs2019/NAS2D.vcxproj
+++ b/proj/vs2019/NAS2D.vcxproj
@@ -51,8 +51,6 @@
     <PlatformToolset>v142</PlatformToolset>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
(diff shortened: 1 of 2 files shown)


DanRStevens:
Heh, from Windows, I never really edited them, but from Linux, there's no choice. I can't exactly run Visual Studio. Every time I need to add a new source file to the project, I have to do it by manually edited the project file with a text editor. That's led me to become more familiar with how they actually work.

Anyway, I've seen a few other things which can be simplified. I'm not sure if they should be, but I'm super curious about what happens to Visual Studio if you simplify them. They work on AppVeyor, and the build still passes, so I may open another PR for more experimental changes, just for the curious. :wink:

PR #369, opened 2020-01-19
Additional cleanup vs project files

DanRStevens:
Here's some more cruft that I believe is safe to remove from the Visual Studio project files. They maybe warrant slightly more attention than the previous set (PR #366).


DanRStevens:
I expect the conditional changes will be overwritten if someone updates project config when it's *not* set to `All Configurations`. I'm uncertain if it will be affected if `All Configurations` is set.

We can always test though. I have a branch I'm about to push with some even crazier edits we can experiment with. :grin:

PR #371, opened 2020-01-19
Use vcpkg to install SDL dependencies

### LINKED ISSUE #365: Switch from NuGet to vcpkg
Currently NuGet is still used to install:
- SDL2
- SDL2_image
- SDL2_mixer
- SDL2_ttf

We should switch these over to `vcpkg`. That will give us a single consistent package manager for all dependencies.

----

The SDL packages can be listed in `InstallVcpkgDeps.bat` as:
```
vcpkg install SDL2:%PLATFORM%-windows
vcpkg install SDL2-mixer:%PLATFORM%-windows
vcpkg install SDL2-image:%PLATFORM%-windows
vcpkg install SDL2-ttf:%PLATFORM%-windows
```

----

Between the two package managers, there is a difference with the automatic `AdditionalIncludeDirectories` (`/I`) settings passed to the compiler. The current settings are effectively (where for brevity `NugetPkg` = `C:\projects\nas2d-core\proj\vs2019\packages\`):
- `/I ../../include`
- `/I "$(NugetPkg)\sdl2_ttf.nuget.2.0.15\build\native\../..//build/native/include/"`
- `/I "$(NugetPkg)\sdl2_mixer.nuget.2.0.4\build\native\../..//build/native/include/"`
- `/I "$(NugetPkg)\sdl2_image.nuget.2.0.5\build\native\../..//build/native/include/"`
- `/I "$(NugetPkg)\sdl2.nuget.2.0.10\build\native\../..//build/native/include/"`
- `/I "C:\Tools\vcpkg\installed\x86-windows\include"`

Note that the SDL2 includes are specific to SDL2, whereas the `vcpkg` include is generic for all installed packages. That leaves a gap in the path for the `SDL2/` folder.

A similar situation arises on Linux:
- `/usr/include/SDL2/` (provided by `sdl2-config --cflags`, among other settings)
- `/usr/include/` (default system include folder)

Thus to access `SDL.h`, you need either:
- `#include <SDL.h>` (using `sdl2-config` for `-I /usr/include/SDL2/`)
- `#include <SDL2/SDL2.h>` (using system default `/usr/include/`)

On MacOS the same `makefile` is used as Linux, and so `sdl2-config` also provides the path to the `SDL2/` folder. As the dependency comes from `brew install` on MacOS, it will likely be outside of the default system include folder, though I suspect `brew` will add to the default include paths.

The other SDL2 modules (`image`, `mixer`, `ttf`) will be in the same install folder as `SDL.h`. That means we either need to have a proper include folder set for all platforms, or the `#include` lines need to account for the `SDL2/` folder.

----

One difficulty with adding a new `/I` option on Windows, is finding the `vcpkg` install folder. The `vcpkg` tool itself doesn't seem to provide for a way to query this folder, nor are there any environment variables holding the location. We could hardcode a path for AppVeyor, but that's an unacceptable solution since it doesn't work well for people developing on Windows, who may have `vcpkg` installed in a different custom folder.

A possible workaround is assuming `vcpkg` is somewhere in the `PATH`, and searching for it that way. The Visual Studio project file supports an `Exec` task that can capture command output to a variable. That could be combined with `where vcpkg` to find the path to the executable. The install folder is then found by stripping off the executable name.

----

Related (issue #294), Visual Studio tasks could also be used to run `vcpkg` commands, much like how NuGet packages are integrated into the project file. We could potentially then add in checks to conditionally install `vcpkg` packages if we detect they are not already installed. This ties the project more closely to `vcpkg`, though as it stands, we are already dependent on it for installing and finding dependencies.

----

The other option, which is likely simpler, and I'm leaning towards, is adjusting all the SDL related include paths:
`#include <SDL2/SDL.h>`

This feels somewhat less proper, as it forces a certain directory layout on developer's systems. If that layout is common between all package managers likely to be used on all platforms, it may be acceptable. It may however complicate switching to other package managers in the future.

DanRStevens:
Closes #365


cugone:
Don't forget to actually delete the `package.config` file from the repo.

cugone:
Just tested with a full `vcpkg remove` then `vcpkg install`. Both the x86 and x64 targets for the `glew` packages fail to build with `BUILD_FAILED` and the output suggests running `vcpkg update`:

vcpkg update/upgrade procedure: https://docs.microsoft.com/en-us/cpp/build/vcpkg?view=vs-2019#updateupgrade-installed-libraries

Similar output occurs with `x64` targets.

>EXEC : error : Building package glew:x86-windows failed with: BUILD_FAILED
>
>Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: glew:x86-windows
  Vcpkg version: 2018.11.23-nohash
>
>Additionally, attach any relevant sections from the log files above.

The relevant output:

>
>The following packages will be built and installed:
    glew[core]:x86-windows
Starting package 1/1: glew:x86-windows
Building package glew[core]:x86-windows...
-- Using cached E:/Git/vcpkg/downloads/glew-2.1.0.tgz
-- Configuring x86-windows-dbg
-- Configuring x86-windows-rel
-- Building x86-windows-dbg
-- Building x86-windows-rel
CMake Error at scripts/cmake/vcpkg_build_cmake.cmake:202 (message):
    Command failed: E:/Git/vcpkg/downloads/tools/cmake-3.14.0-windows/cmake-3.14.0-win32-x86/bin/cmake.exe;--build;.;--config;Release;--target;install;--;/p:VCPkgLocalAppDataDisabled=true;/p:UseIntelMKL=No;/m
    Working Directory: E:/Git/vcpkg/buildtrees/glew/x86-windows-rel
    See logs for more information:
      E:\Git\vcpkg\buildtrees\glew\install-x86-windows-rel-out.log
>
>Call Stack (most recent call first):
  scripts/cmake/vcpkg_install_cmake.cmake:24 (vcpkg_build_cmake)
  ports/glew/portfile.cmake:21 (vcpkg_install_cmake)
  scripts/ports.cmake:71 (include)

---

Unrelated and unavoidable, but still annoying, a full reinstall (`vcpkg install` after `vcpkg remove` and batch building all 8 targets) takes between 15 and 30 seconds per package per target on a 400Mb/s internet speed. I feel sorry for slower internets. Installing after already installed only takes microseconds so that's a plus.

cugone:
Pulling latest and re-running `bootstrap-vcpkg.bat`on the vcpkg repo fixed the issue.

Is there a way to make sure clients have vcpkg up-to-date before trying to build if they already have vcpkg installed (and to either remove/reinstall or update/upgrade all packages)?

DanRStevens:
I'm afraid I don't quite understand what happened here.

It is expected the `InstallVcpkgDeps.bat` will need to be run again to install the SDL packages. The GLEW packages should already be installed from previous builds. I would have expected the installation of that to be skipped, or to run as it did last time. For a fresh clone of the repository, simply running the batch file once should be sufficient.

It's possible `vcpkg` has updated the rule to build GLEW, or that a different version of `vcpkg` was used. It might also be caused by a temporary network error. It's hard to tell from what was pasted. Could you check the log file that was mentioned in the error message?

... I see I'm a little slow replying. Hmm.

DanRStevens:
I have seen some AppVeyor config files that try to update `vcpkg` before installing project specific packages. It does have some kind of update command. That might be a separate PR though. There may also be some pros/cons of doing it, which we might want to discuss. In particular, that could break caching of dependencies, which could really slow down the builds.

I'm tempted to leave things at manual update for `vcpkg` for now.

cugone:
It's evil, but since git is a requirement to build from source, we could just add `git clone/git pull` commands for `vcpkg` before anything else. It'll add upwards of minutes of download time though. 👿

DanRStevens:
Yeah, no, we should probably just rely on manual updates when needed.

DanRStevens:
I'm pretty sure AppVeyor does a reasonable job of keeping their `vcpkg` install up-to-date.

cugone:
I'm not really worried about AppVeyor. I'm more worried about people who can't read and reason about dependencies and will complain that things don't work right out of the box. You know, morons. 😝

DanRStevens:
Yeah, I'm kind of stuck thinking about that now. Thinking we should open an issue to address user friendliness of getting new users up and running. Things like good error messages regarding `vcpkg`, or trying to automate handling for some of it. I think I'll do that.

DanRStevens:
Ok, thank you for the review. Sorry about the difficulty.

@ldicker83, you may want to be aware of this change, since it will require re-running `InstallVcpkgDeps.bat`. If you run into trouble installing dependencies, please check if `vcpkg` needs to be updated.

PR #374, opened 2020-01-19
Resolve windows linker warnings

### LINKED ISSUE #367: Windows linking performance message
When linking `NAS2D.lib`, the following message is shown on AppVeyor:
```
NAS2D.lib(Filesystem.obj) : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance
```

----

References:
Microsoft: [`/LTCG` (Link-time Code Generation)](https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=vs-2019)
StackOverflow: [What causes linker warning “MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG”?](https://stackoverflow.com/questions/26279291/what-causes-linker-warning-msil-netmodule-or-module-compiled-with-gl-found-r)

It sounds like there is a discrepancy between how files are compiled, and how they are linked. Would be nice if someone with Visual Studio could look into this. From the StackOverflow answer, it sounds like the relevant settings are:

> - `/GL` is controlled from Configuration Properties > C/C++ > Optimization > Whole Program Optimization
> - `/LTCG` is controlled from Configuration Properties > Linker > Optimization > Link Time Code Generation / Use Link Time Code Generation (/LTCG)
>

cugone:
Closes #367 

Adds flags `/LTCG` and `/GL` in `test` project `Release` builds.

### DIFF
--- a/test/test.vcxproj
+++ b/test/test.vcxproj
@@ -104,6 +104,7 @@
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <LanguageStandard>stdcpp17</LanguageStandard>
+      <WholeProgramOptimization>true</WholeProgramOptimization>
     </ClCompile>
     <Link>
       <AdditionalDependencies>gtest.lib;gtest_main.lib;opengl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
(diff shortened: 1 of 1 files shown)


cugone:
We should probably add a default "template" downstream project that already has those flags turned on for anyone that wants to make a game with the engine.

DanRStevens:
That's not such a bad idea.

I suppose if we make the template its own repository, it could look something like the [LevelTemplate](https://github.com/OutpostUniverse/LevelTemplate) project. It's listed on GitHub as a template project, so you'll notice the highlighted green button is switched from the `Clone or download` button over to `Use this template`.

Might be worth a new issue.

PR #376, opened 2020-01-19
Simplify VS conditions

DanRStevens:
Ok, this is an experimental change. I'm not sure it's a good idea, but let's have some fun with this one.

I noticed the different configurations have a large amount of similarity. According to the Microsoft documentation, the different sections should inherit and override more global values. I thought I'd try taking advantage of that to see if repetition could be reduced. Indeed, the build passes on AppVeyor.

What I'm actually curious about here, is how well Visual Studio handles these files. Is it glitchy. In particular, if configuration settings are updated, what happens.

Perhaps we can branch off of this, and edit the project configuration in Visual Studio, and see what it does to the project file. Does clear out the new format and revert to the old format. Does it partially retain the new format but update it with old format information. Does it ignore the new format and add back the old format leaving cruft around from the new format which is no longer effective. Or does it do something else entirely.

### DIFF
--- a/test/test.vcxproj
+++ b/test/test.vcxproj
@@ -54,78 +54,63 @@
     </ProjectReference>
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+
+  <ItemDefinitionGroup>
     <ClCompile>
-      <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>..\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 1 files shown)


cugone:
For complicated setups I just provide my own PropertySheets that the user uses to set up the property pages correctly.

See bottom of list here:
https://github.com/cugone/Abrams2019/tree/master/Engine/Code/Engine

All they have to do is add them to their downstream project's proper Property Manager hierarchy in order (hence the naming scheme) and things just work. I could add a template that already adds them but I didn't know that was a thing at the time.

DanRStevens:
Interesting setup you have there for the property sheets.

----

So I'm not sure this should actually be merged. It was more of an experimental idea.

Did you try changing any project settings and re-saving? I'm still curious what Visual Studio does with it.

DanRStevens:
Ok, I'm going to close this now. It was meant more as an experiment to see what was possible. It'll probably cause a lot of churn for changed lines if project settings are ever updated.

It does however provide a good indication of what can be done in the future.

PR #377, opened 2020-01-19
VS wildcard include

DanRStevens:
This is an experimental PR to see how Visual Studio behaves when using wildcard includes:
```
  <ItemGroup>
    <ClCompile Include="..\..\src\**\*.cpp" />
  </ItemGroup>
  <ItemGroup>
    <ClInclude Include="..\..\include\NAS2D\**\*.h" />
  </ItemGroup>
```

Reference: [How to: Select the files to build](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-select-the-files-to-build?view=vs-2019)

One point of interest, is how well Visual Studio handles external modifications to the underlying folders. My guess is, Visual Studio will not see updates without some sort of manual refresh. I remember hearing there was some sort of refresh button for the folder/file view.

One test I want to run, is switching between Git branches that have different files present. I expect the non-wildcard includes would cause Visual Studio to update the file list when it detects a branch change. I am uncertain if Visual Studio will detect such an update and refresh when using the wildcard includes. To make it easy on Visual Studio, I was thinking of testing the branch change with the internal Git support. I suspect Visual Studio may catch and update the folder/file list in that case.

If an external Git tool is used to switch between branches with different files, does Visual Studio notice?

Another test is when the folder is modified externally, outside of Visual Studio. Such as creating a new file inside the folder tree, with an extension that matches the wildcard include. Does Visual Studio notice and display the new file?


cugone:
I added dummy `.cpp` and `.h` files to the monitored folders outside of visual studio and the filters exploded. All file references were moved out of the subfilters and into the base `Source Files` and `Header Files` filters and the custom filters disappeared.

Any attempt to add existing files to the filters throws an exception: `The desired name for [filepath] is invalid` then `The operation could not be completed`.

I thought maybe it had something to do with the inclusion of the `<Extensions>` tag, maybe they were valid extensions lists. Alas, no. The custom filters came back but same error.

Adding new filters is allowed but trying to add files to them only adds them to the base filter.

DanRStevens:
I noticed Brett had completely removed the filters in another project. Sounded like VS just displays the actual folder structure then. I'm not too sure what the experience was like.

I'll dig up the changes he tried, and see if I can apply them here.

cugone:
I prefer filters to organize the solution explorer rather than mucking with the true directory layout.

DanRStevens:
Fair enough. I would prefer sticking to a pretty stock setup, so as not to surprise new users.

I am curious if this setup blows up like the last one though.

cugone:
Still no bueno with the same problems as before except now *every* file in the project folder appears even with Show All Files turned off.

I think leaving it as-is with proper filter use is the way it's intended to be used. Especially since the project files are no where near the actual source files. The filters are needed to correctly organize the project so it knows what to compile.

PR #389, opened 2020-01-22
Prefer in-class initializers

### LINKED ISSUE #387: Prefer in-class initalizers and only initialize class members that are actually passed in as arguments to the Ctor
Related #385 

Ctors that do nothing but initialize class members in an initializer list can be simplified by having sensible defaults declared in the class header and the only members that are listed in the initializer list be the ones actually in the Ctor argument list.

Default Ctors can be completely removed and declared `=default` or `=delete` as needed as well.

cugone:
Closes #387

### DIFF
--- a/include/NAS2D/Resources/FontInfo.h
+++ b/include/NAS2D/Resources/FontInfo.h
@@ -15,13 +15,15 @@
 
 struct GlyphMetrics
 {
-	GlyphMetrics() : uvX(0.0f), uvY(0.0f), uvW(0.0f), uvH(0.0f), minX(0), minY(0), maxX(0), maxY(0), advance(0) {}
-
-	float uvX, uvY;	// Texture coordinates.
-	float uvW, uvH; // Texture coordinates.
-	int minX, minY;
-	int maxX, maxY;
-	int advance;
+	float uvX{0.0f};
+	float uvY{0.0f};
+	float uvW{0.0f};
+	float uvH{0.0f};
+	int minX{0};
+	int minY{0};
+	int maxX{0};
+	int maxY{0};
+	int advance{0};
 };
 
 using GlyphMetricsList = std::vector<GlyphMetrics>;
@@ -32,17 +34,11 @@ using GlyphMetricsList = std::vector<GlyphMetrics>;
  */
 struct FontInfo
 {
-	FontInfo() : texture_id(0), pt_size(0), height(0), ascent(0), ref_count(0)
-	{}
-
-	unsigned int		texture_id;
-	unsigned int		pt_size;
-
-	int					height;
-	int					ascent;
-
-	int					ref_count;
-
-	NAS2D::Point_2d		glyph_size;
-	GlyphMetricsList	metrics;
+	unsigned int texture_id{0u};
+	unsigned int pt_size{0u};
+	int height{0};
+	int ascent{0};
+	int ref_count{0};
+	NAS2D::Point_2d glyph_size{};
+	GlyphMetricsList metrics{};
 };
--- a/include/NAS2D/Resources/Image.h
+++ b/include/NAS2D/Resources/Image.h
@@ -40,10 +40,10 @@ namespace NAS2D {
 class Image: public Resource
 {
 public:
-	Image(const std::string& filePath);
+    Image();
+    Image(const std::string& filePath);
 	Image(void* buffer, int bytesPerPixel, int width, int height);
 	Image(int width, int height);
-	Image();
 
 	Image(const Image &rhs);
 	Image& operator=(const Image& rhs);
@@ -64,8 +64,8 @@ private:
 	void load();
 
 private:
-	std::pair<int, int>				_size;		/**< Width/Height information about the Image. */
-	std::pair<uint32_t, uint32_t>	_center;	/**<  */
+    std::pair<int, int> _size{0,0};		   /**< Width/Height information about the Image. */
+    std::pair<uint32_t, uint32_t> _center{0u,0u}; /**<  */
 };
 
 
--- a/include/NAS2D/Resources/Image.h
+++ b/include/NAS2D/Resources/Image.h
@@ -37,10 +37,10 @@ namespace NAS2D {
  *			and without an alpha channel).
  *
  */
-class Image: public Resource
+class Image : public Resource
 {
 public:
-    Image();
+	Image();
 	explicit Image(const std::string& filePath);
 	Image(void* buffer, int bytesPerPixel, int width, int height);
 	Image(int width, int height);
@@ -64,8 +64,8 @@ private:
 	void load();
 
 private:
-    std::pair<int, int> _size{0,0};		   /**< Width/Height information about the Image. */
-    std::pair<uint32_t, uint32_t> _center{0u,0u}; /**<  */
+	std::pair<int, int> _size{0,0};		   /**< Width/Height information about the Image. */
+	std::pair<uint32_t, uint32_t> _center{0u,0u}; /**<  */
 };
 
 
--- a/include/NAS2D/Resources/Music.h
+++ b/include/NAS2D/Resources/Music.h
@@ -20,16 +20,16 @@ namespace NAS2D {
  *  \class Music
  *  \brief Music Resource.
  */
-class Music: public Resource
+class Music : public Resource
 {
 public:
-	Music();
+	Music() = default;
 	Music(const std::string& filePath);
 
 	Music(const Music& rhs);
 	Music& operator=(const Music& rhs);
 
-	~Music();
+	virtual ~Music();
 
 private:
 	void load();
--- a/include/NAS2D/Resources/Resource.h
+++ b/include/NAS2D/Resources/Resource.h
@@ -24,9 +24,9 @@ namespace NAS2D {
 class Resource
 {
 public:
-	Resource();
+	Resource() = default;
 	Resource(const std::string& filePath);
-	virtual ~Resource();
+	virtual ~Resource() = 0;
 
 	const std::string& name() const;
 
@@ -48,9 +48,8 @@ private:
 	 */
 	virtual void load() = 0;
 
-	std::string mResourceName;		/**< File path and internal identifier. */
-
-	bool mIsLoaded;					/**< Flag indicating whether or not this Resource has loaded properly. */
+	std::string mResourceName{"Default Resource"}; /**< File path and internal identifier. */
+	bool mIsLoaded{false};			 /**< Flag indicating whether or not this Resource has loaded properly. */
 };
 
 } // namespace
--- a/include/NAS2D/Resources/Sound.h
+++ b/include/NAS2D/Resources/Sound.h
@@ -24,10 +24,13 @@ class Sound: public Resource
 {
 public:
 
-	Sound();
+	Sound() = default;
+	Sound(const Sound& other) = default;
+	Sound& operator=(const Sound& rhs) = default;
+	Sound(Sound&& other) = default;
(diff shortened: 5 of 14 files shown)


DanRStevens:
Is it possible to remove the `DEFAULT_ACTION` constant then? It probably shouldn't be defined if we are using hardcoded values instead of the constant. Having it there would give a false sense of security, allowing people to think it only needs to be updated in one place.

include/NAS2D/Resources/Sprite.h
```diff
@@ -136,20 +128,18 @@ class Sprite
 		int frameDelay() const { return mFrameDelay; }
 
 	private:
-		std::string		mSheetId;
-
-		int				mFrameDelay;
-		int				mAnchorX, mAnchorY;
-
-		Rectangle_2d	mRect;
+		std::string mSheetId{};
```
DanRStevens: I'm kind of on the fence about these empty initializers. Seems simpler to not have them there at all. If they weren't being initialized before, they probably don't need to be initialized now.

cugone: Not initializing variables is a source of bugs. "Empty initializers" as you call them are actually zero initializers and are a good thing. They automatically zero the memory for `struct`-like types, or call the default constructor for non-trivial types, instead of initializing to whatever garbage was there already.

cugone:
> 
> Is it possible to remove the `DEFAULT_ACTION` constant then? 

@DanRStevens Done.

include/NAS2D/Resources/Sprite.h
DanRStevens: I'm aware of the difference. I'm suggest zero initializing may not be needed here. If it's not needed why waste the cycles.

As for avoiding bugs, there may be a deeper issue here. Why was no initialization needed before. That may be an indication the variable shouldn't be a class member. Perhaps it would fit better as a local variable in member functions. There were other places in the code where such a change was made.

I would like to look into removing uninitialized member variables before considering default initializing them.

include/NAS2D/Resources/Resource.h
```diff
@@ -24,9 +24,9 @@ namespace NAS2D {
 class Resource
 {
 public:
-	Resource();
+	Resource() = default;
 	Resource(const std::string& filePath);
-	virtual ~Resource();
+	virtual ~Resource() = 0;
```
DanRStevens: Now that I'm looking at this, this class seems kind of odd. It is an ADT. It has a pure virtual member function `load`, however the method is marked `private`. That means this class doesn't define a public interface. Rather it imposes implementation details.

We may need to look at this class more closely. It likely needs a new design.

As for the pure virtual destructor, that's mostly useful when there are no other member functions that can be marked as pure virtual. (Though, having an ADT with no public pure virtual methods is questionable in its own right). Instead, this class could have a defaulted destructor.

include/NAS2D/Resources/Music.h
```diff
@@ -20,16 +20,16 @@ namespace NAS2D {
...
 public:
-	Music();
+	Music() = default;
 	Music(const std::string& filePath);
 
 	Music(const Music& rhs);
 	Music& operator=(const Music& rhs);
 
-	~Music();
+	virtual ~Music();
```
DanRStevens: There are no other virtual member functions. Maybe we don't want to make the destructor virtual.

Do we even want to explicitly declare a destructor here?

include/NAS2D/Resources/Image.h
```diff
@@ -64,8 +64,8 @@ class Image: public Resource
 	void load();
 
 private:
-	std::pair<int, int>				_size;		/**< Width/Height information about the Image. */
-	std::pair<uint32_t, uint32_t>	_center;	/**<  */
+    std::pair<int, int> _size{0,0};		   /**< Width/Height information about the Image. */
+    std::pair<uint32_t, uint32_t> _center{0u,0u}; /**<  */
```
DanRStevens: Kind of off topic, though we should probably avoid using `std::pair` (or tuples) outside of template code. That's a separate issue though. Just noting it down so I don't forget.

include/NAS2D/Resources/Image.h
```diff
@@ -40,10 +40,10 @@ namespace NAS2D {
 class Image: public Resource
 {
 public:
-	Image(const std::string& filePath);
+    Image();
+    Image(const std::string& filePath);
 	Image(void* buffer, int bytesPerPixel, int width, int height);
 	Image(int width, int height);
```
DanRStevens: The indentation in this file looks inconsistent.

include/NAS2D/Resources/FontInfo.h
```diff
@@ -15,13 +15,15 @@
 
 struct GlyphMetrics
 {
-	GlyphMetrics() : uvX(0.0f), uvY(0.0f), uvW(0.0f), uvH(0.0f), minX(0), minY(0), maxX(0), maxY(0), advance(0) {}
```
DanRStevens: Hmm, this one is a struct. Having a constructor makes it non-aggregate. I'm uncertain if having default initialized values does the same. That may depend on the C++ language standard.

We should give some consideration as to whether or not default initialization is even useful here.

include/NAS2D/Resources/Music.h
cugone: Music is derived from Resource which itself has declared its destructor `virtual`. Since the virtual-ness of a function is automatically inherited by derived types it's not *required* that Music's destructor be declared virtual but it makes it more obvious that it is virtual. The `override` keyword not withstanding.

include/NAS2D/Resources/Image.h
cugone: I was refactoring the Image class to #391 specs (I've since decided to move that refactor back a bit behind others) and the first thing I did was replace these with `Vector<int>` and `Vector<uint32_t>` respectively. So I'm in favor of not using `std::pair`. Further off topic, why is center a `uint32_t` but size is an `int`? Maybe that should be normalized too.

include/NAS2D/Resources/Image.h
cugone: Probably a tabs vs spaces and clang-tidy collision.

include/NAS2D/Resources/FontInfo.h
cugone: Could you clarify?

include/NAS2D/Resources/Resource.h
cugone: I whole heartily agree that the Resource class isn't needed at all.

include/NAS2D/Resources/Music.h
DanRStevens: Ahh, that makes more sense.

I suppose if we declare it, we should use `override` then. Though if it's empty, we're not really overriding anything.

include/NAS2D/Resources/Image.h
DanRStevens: I think Leeor was using `uint32_t` as a short hand for `unsigned`. Unfortunately that doesn't play so nice with the differences between 32-bit and 64-bit.

I agree with replacing with `Vector<int>`.

We can probably remove `center`. The `Vector` class supports scalar division, so it's really easy to do without a special method. It might be worth keeping though if we plan to someday support off center images, where the logical center of the object is not the center of the sprite image.

include/NAS2D/Resources/FontInfo.h
DanRStevens: Reference: [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization)

From C++11 until C++14, an aggregate is defined as not having default member initializers.

Since we use C++17, this should be fine.

DanRStevens: To be clear, this one is a bit of an aside. It's not really about the change, but about future potential.

By removing the constructor and replacing it with default member initialization, this type is now an aggregate (in C++17). That's a bonus.

Aggregate types relate to what is safe to `memcpy` around, or serialize directly to/from a file.

include/NAS2D/Resources/Image.h
DanRStevens: Opened issue #399 to track.

include/NAS2D/Resources/Music.h
DanRStevens: Opened issue #400 to track changes to use `override` keyword.

On closer inspection, the destructor is being used here, so does need to be declared. I may have had this confused with another class which had an explicit empty destructor definition.

include/NAS2D/Resources/Resource.h
DanRStevens: Opened issue #401 to track further design review.

include/NAS2D/Resources/FontInfo.h
DanRStevens: Small observation: The `GlyphMetrics` struct is fully initialized in the only method that creates it. It's not all done at once though, so that's not super clear.

Full initialization tends to be typical of structs. I find it makes zero initialization a bit pointless. What I'd really want is a warning if a field is left uninitialized. There are cases where the compiler can warn about use of uninitialized fields, though in general, it can't warn for every possible case. Of course, it can only do that if the field is actually left uninitialized. That may be a consideration for leaving struct fields uninitialized by default.

At any rate, this is a positive change which clearly moves in a good direction. Default initialization is better than constructor list initialization. Whether we want any initialization at all is a separate question.

include/NAS2D/Resources/Sprite.h
DanRStevens: I'm looking back at this now, and realizing part of what bothered me about it. Many of the objects being initialized are not primitive types. That means default initialization using the class constructor. It does not simply zero initialize, as may be done for primitive types.

I think I would prefer omitting the `{}` for object types that have a non-trivial default constructor, such as `std::string` and `std::vector`. Doubly so for `std::vector` since `{}` implies list initialization, which is not really the same as default initialization (just the result is the same in this case for the empty list).

----

For primitive types, if there was no previous initialization, I would assume non-trivial initialization is done in the constructor. If no initialization is done there, it's likely a bug. Using default initialization for those values may hide such bugs.

This reminds me of an article I read a few months back. It mentioned the once common practice of clearing pointer values after calling `delete`. In good code it's an inefficient wasted operation. In bad code, it addresses the symptoms of trying to double `delete`, rather than address the actual cause of the double `delete` itself. The author suggested not bothering with this practice. Just having the code crash may actually be a good thing, since it alerts you to the bug.

I think similar reasoning applies here. A crash, particularly in development, may be preferable. It alerts you to a bug so it can be fixed, rather than papering over the symptoms. A default initialized value might be enough to avoid a spectacular crash, though it still might not work correctly. In such cases, default initialization addresses the symptom, not the cause.

----

For a concrete example, the `Image` copy constructor doesn't set the `_center` value. Default initializing it to `{0, 0}` may make things work smoothly, but its still not correct. (Issue #402).

include/NAS2D/Resources/Sound.h
```diff
@@ -25,11 +25,10 @@ class Sound: public Resource
 public:
 
 	Sound() = default;
-	Sound(const Sound& other) = default;
```
DanRStevens: The merge commit with `master` removes the `Sound` copy constructor. This looks to be accidental. The conversion constructor from `std::string` was moved, so perhaps that's related to how this happened.

DanRStevens:
The majority of these changes have now been broken off into separate PRs and merged individually. It seems there is no plan to pick anymore from this branch. As such I'm going to close this.

I'll leave the branch for a few days, along with the modified `rebase389` branch, in case any more changes need to be picked from it.

Issue #394, opened 2020-01-23
Minimum windows requirements


cugone:
I'd really like to use the modern threading interface API that Microsoft highly recommends using going forward. The problem with using it is it requires Windows 10 version 1607 or later.

Requiring users to have Windows 10 isn't a big deal, I think, because as of last week, Windows 7 is no longer supported and Windows 8 and 8.1 are nearly universally despised by windows users.

Thoughts?

Ref: https://arstechnica.com/information-technology/2020/01/windows-7-im-not-dead-yet/
Ref: https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

Ref: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreaddescription
Ref: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription

DanRStevens:
First question, what would it be used for? I'm not aware of any features for a turn based game that really need much in the way of threading. Music playback maybe?

----

C++ has some recent additions to the standard library for threading. We should probably look there first, before going to something platform specific, especially if it's tied to only the latest version of the Windows platform.

cugone:
To answer your first question, it allows for interfacing with threads as one line of code as seen here:  https://github.com/lairworks/nas2d-core/blob/4ce53bdd3bcfb9182a9f1c783d3c15025fb9ee44/src/ThreadUtils.cpp#L134 which is in line with how `pthreads` does on linux: https://github.com/lairworks/nas2d-core/blob/4ce53bdd3bcfb9182a9f1c783d3c15025fb9ee44/src/ThreadUtils.cpp#L156

Additionally, just because the game being written is turn-based doesn't mean things can't be happening in the background prior to the user clicking next turn:
 - Audio must be handled as a multi-threaded resource or you get stuttering and clicking and deaf users.
 - Real-time File logging
 - A Job System can spin up threads for generic jobs that may be handled asynchronously.
  - Saving the game using a job won't lock up the game.
  - Saving any file, actually...
  - Exporting screenshots is much smoother
 - AI and game state can be updated in the background instead of waiting for it to process all at once.

---
[...]

PR #395, opened 2020-01-23
Tier-0 File logger class

### LINKED ISSUE #393: FileLogger might be helpful
I was looking at OPHD and noticed it used an odd hack to log to a file. A proper multi-threaded FileLogger class might be helpful to include in the engine.

cugone:
Closes #393

Related #394

This is an initial implementation of a multi-threaded file logging class. The bare bones log and logline functions. More can be added later like tags and time stamps.

Also included are classes and functions that support the multi-threadedness of this class and can be used in the future to support a Job System.

### DIFF
--- a/include/NAS2D/FileLogger.h
+++ b/include/NAS2D/FileLogger.h
@@ -24,17 +24,17 @@ public:
 	FileLogger& operator=(FileLogger&& other) = delete;
 	~FileLogger();
 
-	void log(const std::string& msg);
-	void logLine(const std::string& msg);
+	void Log(const std::string& msg);
+	void LogLine(const std::string& msg);
 
 protected:
 private:
-	bool isRunning() const;
-	void setIsRunning(bool value);
-	void initialize(const std::string& logName);
-	void shutdown();
+	bool IsRunning() const;
+	void SetIsRunning(bool value);
+	void Initialize(const std::string& logName);
+	void Shutdown();
 
-	void logWorker();
+	void LogWorker();
 
 	std::streambuf* mOldCout{nullptr};
 	std::ofstream mStreamBuffer{};
--- a/include/NAS2D/FileLogger.h
+++ b/include/NAS2D/FileLogger.h
@@ -1,0 +1,49 @@
+#pragma once
+
+#include "NAS2D/ThreadSafeQueue.h"
+
+#include <atomic>
+#include <condition_variable>
+#include <fstream>
+#include <iosfwd>
+#include <mutex>
+#include <string>
+#include <thread>
+
+namespace NAS2D
+{
+
+class FileLogger
+{
+public:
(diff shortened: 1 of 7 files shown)


cugone:
Can you enumerate those concerns so I can get a better feel for why you have them? What is your definition of complex? It seems to me the code implemented here is pretty straightforward. The gross Unicode stuff is only there because Windows requires its thread names to be in `L""` "Unicode" format. Everything else is literally single-line functions.

As for overkill, no. This is a *dead simple* logger. 

Here's an example of why any logging system should be threaded and an example of a more complex one: https://github.com/tocchan/guildhall/tree/master/sd3/assignments/a03

Throwing 1000+ log messages at it *per frame* and your game doesn't even notice.

You keep thinking in terms of the one game being written right now "because it's turn based, we don't have to worry about it." What if someone wanted to take the engine and make a different game?

Additionally, I can see a situation where each entity (tile, structure, population system, mining, resource manager, etc) in the game is logging many times per frame at least at the beginning of the turn or end for debugging or just general information. With enough active tiles, if the system isn't threaded it will stall your simulation every turn and will make your users complain that the game runs really slow or the time between turns takes "forever" or the UI is unresponsive after a certain number of buildings or colony size.

ldicker83:
I'm not opposed to a multi-threaded logger, I'm just wondering if it's really within the scope of NAS2D? At least for now? I don't want to discourage innovation but I also am unsure about reinventing wheels when other logging libraries are already available and easily included in projects where it's needed?

Most of the logging done by NAS2D itself is really only in debug builds (where performance concerns are effectively non-existent) or in cases of exceptions being thrown so the effect is almost unnoticeable. I can see this being an issue for games that are real-time that use a lot of logging but in those cases I would encourage the end user to roll their own or use an established library.

DanRStevens:
I think this is out of scope.

Current logging produces about a page of text. It's not much. There is little reason to optimize it with threads. I would really like to avoid long term maintenance costs of threaded logging code in this library.

The current logging output often seems like an artifact of the old error handling code. Much of it should be converted to exceptions and/or unit tests. In particular, developers who use the library may want full control over what is output to the user, so we should avoid direct calls to `std::cout` and `std::cerr` from within the library. Instead, an exception based approach should be used, so developers can catch exceptions and relevant messages in their own code and decide what to do with it. That might mean printing it to the console, logging it to a file, or outputting it in some sort of graphical console, or even just in a message box (for early development).

My preference is to reduce and remove logging from this library, and make it an external concern.

Along those lines, if the logging code suggested here is generic, perhaps it can become its own library.

DanRStevens:
As per my comment in the associated issue, I think I'm going to close this as not an area of priority for us right now.

I'll leave the branch for a few days in case you want to do anything with the code.

PR #398, opened 2020-01-25
Makefile cross compile support

DanRStevens:
This adds a bit of `makefile` support so NAS2D can be cross compiled from Linux for Windows, using the Mingw-w64 toolchain.

Mostly this is renaming a few required library includes, and prefixing the unit test running with `wine`.

The real work was in getting a build environment setup. A Dockerfile for the build environment will follow a little later.


DanRStevens:
Makefile only change.

PR #403, opened 2020-01-26
Image class refactor

cugone:
This is massive only because of the pervasive nature of the class.

Feedback and suggestions are highly encouraged.

I need guidance on what to do with the Sprite class' use of image directly. I could go the route of creating a SpriteSheet class (and an AnimatedSprite for completionist-sake)...

Changes include:
 - Image class is now a blob of data
 - Images are decoupled from Texture data so default textures can be supported.
 - Support for stbi image loading in OpenGL and DirectX (removes SDL)
 - Initial support for render-specific textures (OGL and DirectX)

Known Issues
 - Sprite class uses the Image class directly instead of references to a weak_ptr to a sprite sheet that lives on the Renderer. This uses a lot of memory.
 - The Sprite class acts more like an AnimatedSprite than a static Sprite.
 - The above issues also make supporting Batching with a single large spritesheet unavailable.

### DIFF
--- a/include/NAS2D/Renderer/OpenGL/TextureOpenGL.h
+++ b/include/NAS2D/Renderer/OpenGL/TextureOpenGL.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include "NAS2D/Renderer/Texture.h"
+
+namespace NAS2D
+{
+	class TextureOpenGL : public Texture
+	{
+	  public:
+		TextureOpenGL(unsigned int id);
+		virtual ~TextureOpenGL() = default;
+
+		unsigned int Id() const noexcept;
+	  protected:
+	  private:
+		unsigned int _id{0};
+	};
+
+} // namespace NAS2D
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -17,9 +17,14 @@
 #include "../Resources/Image.h"
 #include "../Resources/Font.h"
 
+#include "NAS2D/Renderer/Texture.h"
+
 #include <string>
+#include <memory>
 
 namespace NAS2D {
+
+	class Texture;
 
 /**
  * \class Renderer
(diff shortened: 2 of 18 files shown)


DanRStevens:
This just changes too much at once. There's no reasonable way to review and merge this.

I'd also be very hesitant to incorporate additional external dependencies into the code base. Particularly since that external code does not compile cleanly on GCC and Clang.

ldicker83:
There are a number of problems here.

First, it fails numerous build tests and fails the codacy Code Quality check.

Second, many of the methods don't follow code conventions -- e.g., class methods start with capital letters when the rest of the code starts everything with lower case. Additionally, you have several 'get' and 'set' type methods with Get and Set as part of the method names. This does not follow conventions in the rest of the code base.

A number of methods were added to the Renderer class that pollute the interface with implementation details. This is something I wanted to avoid and went through great lengths to clean up. The interface exposes no methods or variables that indicate implementation details. All of that is hidden in local modules which is the intended and preferred way of going about it.

You mention that Sprite acts as an Animated Sprite... that's precisely what was intended. Sprites can be a static image or an animated image. I would rather not create a static sprite and then a derived animated sprite. For me, they're one and same thing so there's very little value in doing it that way.

Additionally, yes the Sprite class does use an Image internally, but Image only loads itself once while it's in use. When the image loads, it creates a single texture for itself and references said texture once. Batching should be very doable using this method. Image is literally just an interface where the low-level stuff is handled internally. This does make it difficult to support different types of renderers but this is where discussion should come in about how to move forward with it. In 2008 when I first started the code I had originally intended to have an SDL, OpenGL and DirectX renderer but as time progressed it occurred to me that a software renderer was useless and there was little motivation to build and maintain a DirectX renderer so I opted to stick with an OpenGL implementation. This has the unfortunate side effect of being rigid and raises issues should other Renderer implementations be desired.

And here we are -- I'd love to support other renderers but this is something that needs to be discussed and decided upon before tearing things apart and adding code that fails to conform to code styling with the rest of the library and makes changes to established assumptions.

I've said it before -- I don't want to stifle innovation and I don't want to sound ungrateful, on the contrary, I love that OPHD has provided enough interest to garner interest and assistance with its middleware NAS2D, but just going in and doing leads to unmanageable code and unnecessarily complicated code especially when feature creep starts becoming a factor.

DanRStevens:
I'm going to close this due to some of the build problems mentioned above, and because of the difficulty in reviewing such a large change set.

It does look like there were some good ideas in here, so perhaps parts can be picked out into smaller change sets and merged individually.

I'm going to leave the branch around for now.

Issue #415, opened 2020-01-30
Add normalized audio volume getters and setters to Configuration and Mixer


cugone:
In order to facilitate abstractions from different audio implementations, `Configuration` and `Mixer` should provide volume setters that take normalized values and getters that return the volume as a normalized value.

Derived classes can then denormalize or rangemap the values based on their own implementations. E.g. SDL_mixer uses [0,128], while XAudio2 (modern Windows API) uses [-1.0,1.0].

Additionally, SDL uses the same setter with a value of `-1` to mean "don't set it, I just care about the current value.", while XAudio2 uses a dedicated getter.

Ref: https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer.html#SEC27
Ref: https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer.html#SEC61
Ref: https://docs.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2voice-setvolume
Ref: https://docs.microsoft.com/en-us/windows/win32/xaudio2/xaudio2-volume-and-pitch-control
Ref: https://docs.microsoft.com/en-us/windows/win32/xaudio2/how-to--change-voice-volume

DanRStevens:
One caveat is that if we always interface to audio through SDL, then it sort of doesn't matter what the platform implementation is. Using floating point would then be converted to integer for SDL, which may or may not be converted back to floating point to interface to the platform's native system.

Other than that, this sounds like a reasonable change. Would be curious if Leeor wanted to weigh in on this.

PR #418, opened 2020-01-30
Implements convertDomainToRange

### LINKED ISSUE #416: Add MathUtils::RangeMap
Related: #415 

This will make is way easier to map one range of values to another. E.g. Mapping a normalized slider value to SDL_mixer's minimum and maximum volume values.
```
RangeMap(value, 0.0f, 1.0f, 0.0f, 128.0f);
```

cugone:
Closes #416

Related #415 

Adds a function to linearly map an input range to an output range.

### DIFF
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -45,8 +45,8 @@ namespace NAS2D
  *
  * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
  */
-template<typename T>
-T convertDomainToRange(const T& value, const T& minDomain, const T& maxDomain, const T& minRange, const T& maxRange)
+template<typename InputType, typename OutputType>
+OutputType convertDomainToRange(const InputType& value, const InputType& minDomain, const InputType& maxDomain, const OutputType& minRange, const OutputType& maxRange)
 {
 	return (value - minDomain) * (maxRange - minRange) / (maxDomain - minDomain) + minRange;
 }
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -46,9 +46,9 @@ namespace NAS2D
  * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
  */
 template<typename InputType, typename OutputType>
-OutputType convertDomainToRange(const InputType& value, const InputType& minDomain, const InputType& maxDomain, const OutputType& minRange, const OutputType& maxRange)
+OutputType convertDomainToRange(const InputType& value, const InputType& domainPoint1, const InputType& domainPoint2, const OutputType& rangePoint1, const OutputType& rangePoint2)
 {
-	return (value - minDomain) * (maxRange - minRange) / (maxDomain - minDomain) + minRange;
+	return (value - domainPoint1) * (rangePoint2 - rangePoint1) / (domainPoint2 - domainPoint1) + rangePoint1;
 }
 
 } // namespace NAS2D
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -11,6 +11,7 @@
 
 #include "Renderer/Point.h"
 #include "Renderer/Rectangle.h"
+#include "Renderer/Vector.h"
 
 #include <memory>
 #include <string>
@@ -28,6 +29,27 @@ namespace NAS2D
 	bool isRectInRect(const Rectangle_2d& a, const Rectangle_2d& b);
 
 	int divideUp(int to_divide, int divisor);
+
+	
+
+/**
+ * \fn T convertDomainToRange(T a, T b)
+ *
+ * Applies a linear conversion to map an input domain to an output range for built-in types.
+ *
+ * \param	value: The value to be converted from the input domain to the output range.
+ * \param	minDomain: Smallest value of the domain.
+ * \param	maxDomain: Largest value of the domain.
+ * \param	minRange: Smallest value of the range.
+ * \param	maxRange: Largest value of the range.
+ *
+ * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ */
+template<typename T>
+T convertDomainToRange(const T& value, const T& minDomain, const T& maxDomain, const T& minRange, const T& maxRange)
+{
+	return (value - minDomain) * (maxRange - minRange) / (maxDomain - minDomain) + minRange;
+}
 
 } // namespace NAS2D
 
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -38,17 +38,17 @@ namespace NAS2D
  * Applies a linear conversion to map an input domain to an output range.
  *
  * \param	value: The value to be converted from the input domain to the output range.
- * \param	domainPoint1: A known lower fixed point in the domain. Must be less than domainPoint2.
- * \param	domainPoint2: A known upper fixed point in the domain. Must be greater than domainPoint1.
- * \param	rangePoint1: A known lower fixed point in the range. Must be less than or equal to rangePoint2.
- * \param	rangePoint2: A known upper fixed point in the range. Must be greater than or equal to rangePoint1.
+ * \param	domainPoint1: A known fixed point in the domain.
+ * \param	domainPoint2: Another corresponding known fixed point in the domain.
+ * \param	rangePoint1: A known fixed point in the range.
+ * \param	rangePoint2: Another corresponding known fixed point in the range.
  * \return	Returns value as mapped to the output range.
  * \remarks
  * `value` is not required to start within [domainPoint1,domainPoint2]. E.g. temperature conversions:
  *
- *     scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f)
+ *     scaleLinear(98.6f, 32.0f, 212.0f, 0.0f, 100.0f)
  *
- * will convert 0 degees Fahrenheit to its corresponding value `-17.77777` degrees Celsius.
+ * will convert 98.6 degees Fahrenheit (average human body temperature) to its corresponding value `37.0` degrees Celsius.
  * 
  * Avoid inputs such that the difference of domainPoint2 and domainPoint1 is zero. Otherwise, a divide by zero error could occur.
  * 
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -46,7 +46,7 @@ namespace NAS2D
  * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
  */
 template<typename InputType, typename OutputType>
-OutputType convertDomainToRange(const InputType& value, const InputType& domainPoint1, const InputType& domainPoint2, const OutputType& rangePoint1, const OutputType& rangePoint2)
+OutputType scaleLinear(const InputType& value, const InputType& domainPoint1, const InputType& domainPoint2, const OutputType& rangePoint1, const OutputType& rangePoint2)
 {
 	return (value - domainPoint1) * (rangePoint2 - rangePoint1) / (domainPoint2 - domainPoint1) + rangePoint1;
 }
--- a/include/NAS2D/MathUtils.h
+++ b/include/NAS2D/MathUtils.h
@@ -33,17 +33,28 @@ namespace NAS2D
 	
 
 /**
- * \fn T convertDomainToRange(T a, T b)
+ * \fn OutputType scaleLinear(const InputType& a, const OutputType& b)
  *
- * Applies a linear conversion to map an input domain to an output range for built-in types.
+ * Applies a linear conversion to map an input domain to an output range.
  *
  * \param	value: The value to be converted from the input domain to the output range.
- * \param	minDomain: Smallest value of the domain.
- * \param	maxDomain: Largest value of the domain.
- * \param	minRange: Smallest value of the range.
- * \param	maxRange: Largest value of the range.
+ * \param	domainPoint1: A known lower fixed point in the domain. Must be less than domainPoint2.
+ * \param	domainPoint2: A known upper fixed point in the domain. Must be greater than domainPoint1.
+ * \param	rangePoint1: A known lower fixed point in the range. Must be less than or equal to rangePoint2.
+ * \param	rangePoint2: A known upper fixed point in the range. Must be greater than or equal to rangePoint1.
+ * \return	Returns value as mapped to the output range.
+ * \remarks
+ * `value` is not required to start within [domainPoint1,domainPoint2]. E.g. temperature conversions:
  *
- * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ *     scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f)
+ *
+ * will convert 0 degees Fahrenheit to its corresponding value `-17.77777` degrees Celsius.
+ * 
+ * Avoid inputs such that the difference of domainPoint2 and domainPoint1 is zero. Otherwise, a divide by zero error could occur.
+ * 
+ * If the difference of rangePoint2 and rangePoint1 is zero. The result of the function is always rangePoint1.
+ * 
+ * InputType and OutputType must support all arithmetic operators and all arithmetic operators between each other.
  */
 template<typename InputType, typename OutputType>
 OutputType scaleLinear(const InputType& value, const InputType& domainPoint1, const InputType& domainPoint2, const OutputType& rangePoint1, const OutputType& rangePoint2)
--- a/test/MathUtils.test.cpp
+++ b/test/MathUtils.test.cpp
@@ -223,6 +223,8 @@ TEST(MathUtils, mapDomainToRange)
 {
 	// Fahrenheit to Celcius
 	EXPECT_NEAR(-18.33333f, NAS2D::scaleLinear(-1.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
+	EXPECT_NEAR(-18.33333f, NAS2D::scaleLinear(-1.0f, 212.0f, 32.0f, 100.0f, 0.0f), 0.0001f);
+
 	EXPECT_NEAR(0.0f, NAS2D::scaleLinear(32.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
 	EXPECT_NEAR(-17.77777f, NAS2D::scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
 	EXPECT_NEAR(-17.22222f, NAS2D::scaleLinear(1.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
--- a/test/MathUtils.test.cpp
+++ b/test/MathUtils.test.cpp
@@ -218,3 +218,31 @@ TEST(MathUtils, isRectInRectClass)
 	EXPECT_EQ(false, NAS2D::isRectInRect({5, 5, 1, 1}, {3, 3, 1, 1}));
 	EXPECT_EQ(false, NAS2D::isRectInRect({5, 5, 1, 1}, {7, 3, 1, 1}));
 }
+
+TEST(MathUtils, mapDomainToRange)
+{
+	//I know the temperature conversions have established formulas, but that's not what I'm testing.
+	//Given proper input I should get the same value as if I used the derived formulas.
+
+	//(v - minD) * (maxR - minR) / (maxD - minD) + minR;
+	//(F - 32.0f) * (5.0f / 9.0f) + 0.0f;
+	//(C - 0.0f) * (9.0f / 5.0f) + 32.0f;
+
+	//Fahrenheit to Celcius
+	EXPECT_NEAR(0.0f, NAS2D::convertDomainToRange(32.0f, 32.0f, 41.0f, 0.0f, 5.0f), 0.0001f);
+	EXPECT_NEAR(100.0f, NAS2D::convertDomainToRange(212.0f, 32.0f, 41.0f, 0.0f, 5.0f), 0.0001f);
+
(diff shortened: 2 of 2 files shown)


include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,30 @@ namespace NAS2D
...
+ * \param	valueToMap: The value to be mapped from the input range to the output range. Must start in the closed range [minInputRange,maxInputRange]
+ * \param	minInputRange: Smallest value to map from.
+ * \param	maxInputRange: Largest value to map from.
+ * \param	minOutputRange: Smallest value to map to.
+ * \param	maxOutputRange: Largest value to map to.
+ *
+ * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ */
+template<typename T>
+T rangeMap(const T& valueToMap, const T& minInputRange, const T& maxInputRange, const T& minOutputRange, const T& maxOutputRange)
```
DanRStevens: For simple primitive types, which this is likely to be used with, we should drop the `const &`.

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,30 @@ namespace NAS2D
...
+{
+	return (valueToMap - minInputRange) * (maxOutputRange - minOutputRange) / (maxInputRange - minInputRange) + minOutputRange;
+}
+
+Vector<float> rangeMap(const Vector<float>& valueToMap, const Vector<float>& minmaxInputRange, const Vector<float>& minmaxOutputRange);
+Vector<int> rangeMap(const Vector<int>& valueToMap, const Vector<int>& minmaxInputRange, const Vector<int>& minmaxOutputRange);
```
DanRStevens: Hah, I was actually thinking of suggesting something like this. I kind of stopped and realized though that `Vector` is perhaps not the right usage here. A `Vector` has independent `x` and `y` values, which represents different axis. Here we have `min` and `max` values, which are points along the same axis. This is probably an abuse of `Vector` and we should avoid re-using it here. A `Range` is a different concept from a `Vector`.

I would recommend getting rid of these methods for now. A later PR could upgrade the method with a `Range` type to provide such use.

----

Actually, the template method could potentially be applied to 2D spaces, mapping points between each other linearly. That would be different from the above, since you would have 2D points for both `min` and `max`, for both the domain and the range.

include/NAS2D/MathUtils.h
cugone: I'm a little hesitant to drop the `const &` because it denotes at a glance that `valueToMap` will not be changed. As long as the type supports all the basic arithmetic operations it can be used here.

DanRStevens: The exact same thing can be said about passing by value. Plus that would remove the inefficiency of extra indirection.

cugone: A passed-by-value parameter can be accidentally changed. A const value must be deliberately copied then changed. That is, it is indicative of a poor design or a malicious/dumb programmer is messing with something they probably shouldn't.

cugone: *...accidentally changed inside the function...

DanRStevens: Run the code through a compiler and look at the assembly output. See how changing between `const &` and by-value affects the generated code. I recommend:
https://godbolt.org/

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,26 @@ namespace NAS2D
...
+ * \param	value: The value to be converted from the input domain to the output range.
+ * \param	minDomain: Smallest value of the domain.
+ * \param	maxDomain: Largest value of the domain.
+ * \param	minRange: Smallest value of the range.
+ * \param	maxRange: Largest value of the range.
+ *
+ * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ */
+template<typename T>
+T convertDomainToRange(const T& value, const T& minDomain, const T& maxDomain, const T& minRange, const T& maxRange)
```
DanRStevens: I would recommend two template parameters here, one for `InputType` and one for `OutputType`. The first three parameters would be `InputType`, while the last two would be `OutputType`. That way you could convert a scalar value to say a 2D vector.

----

Input and Output types are largely going to be simple primitive types, or simple value-type structs. We should consider passing by value rather than `const &`. For those who are curious what impact that has on the code, I highly recommend checking https://godbolt.org/ to see what assembly code is generated by a few mainstream compilers.

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,26 @@ namespace NAS2D
 
 	int divideUp(int to_divide, int divisor);
 
+	
+
+/**
+ * \fn T convertDomainToRange(T a, T b)
```
DanRStevens: The name `convertDomainToRange` is pretty generic, and could be descriptive of *any* function.

I would recommend `scaleLinear`, or similar. I believe the "linear" aspect is important, and should be part of the name. I might also be willing to entertain something like `mapLinear`. Not to be confused with [map](https://en.wikipedia.org/wiki/Map_(higher-order_function)) found in several popular programming languages.

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,26 @@ namespace NAS2D
...
+ *
+ * Applies a linear conversion to map an input domain to an output range for built-in types.
+ *
+ * \param	value: The value to be converted from the input domain to the output range.
+ * \param	minDomain: Smallest value of the domain.
+ * \param	maxDomain: Largest value of the domain.
+ * \param	minRange: Smallest value of the range.
+ * \param	maxRange: Largest value of the range.
```
DanRStevens: These parameter names are perhaps slightly incorrect. This function actually works quite well outside of the min/max bounds, by simple linear extension. Instead, these parameters mark two corresponding fixed points, which sets the linear scale being used. They do not need to be at the extreme ends of the scale.

Typical values may be the extreme ends, so that's fine, and might be worth a comment. I would however like to document the function works outside of those bounds by linear extension.

Since naming is hard, I suppose I should at least attempt to provide a suggestion:
- `domainPoint1`: Typically the minimum value for domain
- `domainPoint2`: Typically the maximum value for domain
- `rangePoint1`: Typically the minimum value for range
- `rangePoint2`: Typically the maximum value for range
We should perhaps also specify that corresponding points are mapped to each other other, while other values are linearly interpolated.

----

Boundary case:
- I think it's fine if `rangePoint1` coincides with `rangePoint2`. In that case, all inputs produce the same output value.
- We should document what happens if `domainPoint1` coincides with `domainPoint2`. We could to return `rangePoint1` for all values, or return `rangePoint2` for all values, or we could `throw` an exception. Actually, We basically get the division by 0 case here, so probably throw an exception.

test/MathUtils.test.cpp
```diff
@@ -218,3 +218,31 @@ TEST(MathUtils, isRectInRectClass)
...
+TEST(MathUtils, mapDomainToRange)
+{
+	//I know the temperature conversions have established formulas, but that's not what I'm testing.
+	//Given proper input I should get the same value as if I used the derived formulas.
+
+	//(v - minD) * (maxR - minR) / (maxD - minD) + minR;
+	//(F - 32.0f) * (5.0f / 9.0f) + 0.0f;
+	//(C - 0.0f) * (9.0f / 5.0f) + 32.0f;
+
+	//Fahrenheit to Celcius
```
DanRStevens: Very minor point, though most comments in this code base have a space after the `//`
```
// Fahrenheit to Celcius
```

Same applies to the other comments in this file.

test/MathUtils.test.cpp
```diff
@@ -218,3 +218,31 @@ TEST(MathUtils, isRectInRectClass)
...
 }
+
+TEST(MathUtils, mapDomainToRange)
+{
+	//I know the temperature conversions have established formulas, but that's not what I'm testing.
+	//Given proper input I should get the same value as if I used the derived formulas.
+
+	//(v - minD) * (maxR - minR) / (maxD - minD) + minR;
+	//(F - 32.0f) * (5.0f / 9.0f) + 0.0f;
+	//(C - 0.0f) * (9.0f / 5.0f) + 32.0f;
```
DanRStevens: I think this comment could potentially be removed. If we choose two well known points from either scale, I don't think it should be necessary to demonstrate the formula. For example, the melting and boiling points of water: `0C..100C` => `32F..212F`

Btw, nice example for a test case.

test/MathUtils.test.cpp
```diff
@@ -218,3 +218,31 @@ TEST(MathUtils, isRectInRectClass)
...
+	//Fahrenheit to Celcius
+	EXPECT_NEAR(0.0f, NAS2D::convertDomainToRange(32.0f, 32.0f, 41.0f, 0.0f, 5.0f), 0.0001f);
+	EXPECT_NEAR(100.0f, NAS2D::convertDomainToRange(212.0f, 32.0f, 41.0f, 0.0f, 5.0f), 0.0001f);
+
+	//Celcius to Fahrenheit
+	EXPECT_NEAR(32.0f, NAS2D::convertDomainToRange(0.0f, 0.0f, 5.0f, 32.0f, 41.0f), 0.0001f);
+	EXPECT_NEAR(212.0f, NAS2D::convertDomainToRange(100.0f, 0.0f, 5.0f, 32.0f, 41.0f), 0.0001f);
+
+	//char to normalized float
+	EXPECT_NEAR(0.0f, NAS2D::convertDomainToRange(0.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
```
DanRStevens: To be overly pedantic, the comment should read `unsigned char`, since we are assuming a range of `0..255`. The standard makes no guarantee if `char` is `signed` or `unsigned`.

If the template is updated with two type parameters, the inputs could be given as actual `unsigned char` or `int`, rather than as `float`.

----

More generally, for all `EXPECT` calls, usually the expected value is given first on the left, while the method being tested is on the right. Most other unit tests in the project are written in that order. Here the order is swapped.

include/NAS2D/MathUtils.h
cugone: I would disagree on not using `const&`. See: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#fcall-parameter-passing

Specifically the center of the table (emphasis mine):

>
>Don't know (e.g., unfamiliar type, **template**) input parameters: `f(const&)`

image

test/MathUtils.test.cpp
cugone: Good point. The formulas were a note for me as to what the input parameters to the function should be as they corresponded to the internal algorithm.

At the same time, it does demonstrate that you can't just  use `0,100,32,212` as input and get it to work...

test/MathUtils.test.cpp
cugone: Noted on the `unsigned char`.

The expected value *is* on the left and the method is on the right.

`Expect_Near` takes a third parameter: `abs_error` which is the acceptable `epsilon` of the `expected result` vs the `actual result`. Because I knew I was dealing with `float`s, testing exact equality is a logic error.

test/MathUtils.test.cpp
cugone: Disregard Previous. I just tested `EXPECT_NEAR(-17.22222f, NAS2D::convertDomainToRange(1.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);` (1F to C) using the fixed boiling/freezing points of water and it passed. Math is funny some times.

include/NAS2D/MathUtils.h
cugone: Documented expected inputs to the function.

test/MathUtils.test.cpp
DanRStevens: Oh, geeze, I read that wrong. I'm so used to the non-floating point versions of `EXPECT`.

include/NAS2D/MathUtils.h
```diff
@@ -33,17 +33,28 @@ namespace NAS2D
...
- * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ *     scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f)
+ *
+ * will convert 0 degees Fahrenheit to its corresponding value `-17.77777` degrees Celsius.
+ * 
+ * Avoid inputs such that the difference of domainPoint2 and domainPoint1 is zero. Otherwise, a divide by zero error could occur.
+ * 
+ * If the difference of rangePoint2 and rangePoint1 is zero. The result of the function is always rangePoint1.
+ * 
+ * InputType and OutputType must support all arithmetic operators and all arithmetic operators between each other.
```
DanRStevens: This line is slightly incorrect, since not *all* arithmetic operations need to be supported, just the ones used by the implementation. It's also kind of hand wavy about what those operations actually are. As such, it doesn't really add much value.

I would be fine if we left this unspecified for now. We might need more examples of mixed input and output types to really drive this point home anyway.

include/NAS2D/MathUtils.h
```diff
@@ -33,17 +33,28 @@ namespace NAS2D
...
+ * \param	rangePoint1: A known lower fixed point in the range. Must be less than or equal to rangePoint2.
+ * \param	rangePoint2: A known upper fixed point in the range. Must be greater than or equal to rangePoint1.
+ * \return	Returns value as mapped to the output range.
+ * \remarks
+ * `value` is not required to start within [domainPoint1,domainPoint2]. E.g. temperature conversions:
  *
- * \return	Returns the mapped value in the closed range [minOutputRange,maxOutputRange]
+ *     scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f)
+ *
+ * will convert 0 degees Fahrenheit to its corresponding value `-17.77777` degrees Celsius.
```
DanRStevens: An interesting example might be body temperature (37C). There isn't really anything special about 0F, so this example seems a bit arbitrary.

test/MathUtils.test.cpp
```diff
@@ -222,11 +222,16 @@ TEST(MathUtils, isRectInRectClass)
 TEST(MathUtils, mapDomainToRange)
 {
 	// Fahrenheit to Celcius
+	EXPECT_NEAR(-18.33333f, NAS2D::convertDomainToRange(-1.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
 	EXPECT_NEAR(0.0f, NAS2D::convertDomainToRange(32.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
+	EXPECT_NEAR(-17.77777f, NAS2D::convertDomainToRange(0.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
+	EXPECT_NEAR(-17.22222f, NAS2D::convertDomainToRange(1.0f, 32.0f, 212.0f, 0.0f, 100.0f), 0.0001f);
```
DanRStevens: The "around zero" argument would make more sense for this method for an `unsigned` data type. For `signed` data types like these temperature conversions, particularly from degrees F, there doesn't seem to be anything particularly special about 0.

If it's a well know conversion point, that's fine. In particular, 0 degrees C is the melting/freezing point of water. If not, and I'm not aware of anything special happening at 0F, other than it's a nice round number and maybe people have memorized this, it maybe starts to approach the realm of [Cargo Cult Programming](https://en.wikipedia.org/wiki/Cargo_cult_programming).

With that said, we might perhaps still want to add testing where `InputType` is an `unsigned` quantity, and include a zero point there.

test/MathUtils.test.cpp
```diff
@@ -218,3 +218,29 @@ TEST(MathUtils, isRectInRectClass)
...
+	EXPECT_NEAR(212.0f, NAS2D::scaleLinear(100.0f, 0.0f, 100.0f, 32.0f, 212.0f), 0.0001f);
+
+	// unsigned char to normalized float
+	EXPECT_NEAR(0.0f, NAS2D::scaleLinear(0.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(0.5f, NAS2D::scaleLinear(128.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(1.0f, NAS2D::scaleLinear(255.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
+
+	EXPECT_NEAR(-1.0f, NAS2D::scaleLinear(0.0f, 0.0f, 255.0f, -1.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(0.0f, NAS2D::scaleLinear(128.0f, 0.0f, 255.0f, -1.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(1.0f, NAS2D::scaleLinear(255.0f, 0.0f, 255.0f, -1.0f, 1.0f), 0.01f);
```
DanRStevens: Now that the template has both `InputType` and `OutputType`, we might consider updating this to use integers for the first 3 parameters.

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,37 @@ namespace NAS2D
...
+ * \remarks
+ * `value` is not required to start within [domainPoint1,domainPoint2]. E.g. temperature conversions:
+ *
+ *     scaleLinear(0.0f, 32.0f, 212.0f, 0.0f, 100.0f)
+ *
+ * will convert 0 degees Fahrenheit to its corresponding value `-17.77777` degrees Celsius.
+ * 
+ * Avoid inputs such that the difference of domainPoint2 and domainPoint1 is zero. Otherwise, a divide by zero error could occur.
+ * 
+ * If the difference of rangePoint2 and rangePoint1 is zero. The result of the function is always rangePoint1.
```
DanRStevens: I was kind of thinking this might be simpler to talk about points between equal, rather than having a difference of zero. I suppose this is technically more accurate though, since you can create an exotic type where they have a distance of zero but are not equal. Just wanted to draw attention to that technicality.

include/NAS2D/MathUtils.h
```diff
@@ -29,5 +30,37 @@ namespace NAS2D
...
+/**
+ * \fn OutputType scaleLinear(const InputType& a, const OutputType& b)
+ *
+ * Applies a linear conversion to map an input domain to an output range.
+ *
+ * \param	value: The value to be converted from the input domain to the output range.
+ * \param	domainPoint1: A known lower fixed point in the domain. Must be less than domainPoint2.
+ * \param	domainPoint2: A known upper fixed point in the domain. Must be greater than domainPoint1.
+ * \param	rangePoint1: A known lower fixed point in the range. Must be less than or equal to rangePoint2.
+ * \param	rangePoint2: A known upper fixed point in the range. Must be greater than or equal to rangePoint1.
```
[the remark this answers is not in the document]
cugone: You are correct. I had forgotten to swap the output range when testing it locally.

Added tests to address this.

include/NAS2D/MathUtils.h
cugone: I initially wrote out the requirements to support the specific operations but found it confusing. It's easier to just require all of them. It's only four functions anyway.

test/MathUtils.test.cpp
```diff
@@ -235,12 +237,12 @@ TEST(MathUtils, mapDomainToRange)
 	EXPECT_NEAR(212.0f, NAS2D::scaleLinear(100.0f, 0.0f, 100.0f, 32.0f, 212.0f), 0.0001f);
 
 	// unsigned char to normalized float
-	EXPECT_NEAR(0.0f, NAS2D::scaleLinear(0.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
-	EXPECT_NEAR(0.5f, NAS2D::scaleLinear(128.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
-	EXPECT_NEAR(1.0f, NAS2D::scaleLinear(255.0f, 0.0f, 255.0f, 0.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(0.0f, NAS2D::scaleLinear(static_cast<unsigned char>(0), static_cast<unsigned char>(0), static_cast<unsigned char>(255), 0.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(0.5f, NAS2D::scaleLinear(static_cast<unsigned char>(128), static_cast<unsigned char>(0), static_cast<unsigned char>(255), 0.0f, 1.0f), 0.01f);
+	EXPECT_NEAR(1.0f, NAS2D::scaleLinear(static_cast<unsigned char>(255), static_cast<unsigned char>(0), static_cast<unsigned char>(255), 0.0f, 1.0f), 0.01f);
```
DanRStevens: Hmm, the use of `static_cast` seems a bit unfortunate. I looked, but it seems there is no suffix to specify `unsigned char`. We could approximate with `unsigned int`:
```cpp
EXPECT_NEAR(0.0f, NAS2D::scaleLinear(0u, 0u, 255u, 0.0f, 1.0f), 0.01f);
```

We could also instantiate the template with explicit types:
```cpp
EXPECT_NEAR(0.0f, NAS2D::scaleLinear<unsigned char, float>(0u, 0u, 255u, 0.0f, 1.0f), 0.01f);
```

include/NAS2D/MathUtils.h
```diff
@@ -38,17 +38,17 @@ namespace NAS2D
...
  *
  * \param	value: The value to be converted from the input domain to the output range.
- * \param	domainPoint1: A known lower fixed point in the domain. Must be less than domainPoint2.
- * \param	domainPoint2: A known upper fixed point in the domain. Must be greater than domainPoint1.
- * \param	rangePoint1: A known lower fixed point in the range. Must be less than or equal to rangePoint2.
- * \param	rangePoint2: A known upper fixed point in the range. Must be greater than or equal to rangePoint1.
+ * \param	domainPoint1: A known fixed point in the domain.
+ * \param	domainPoint2: Another known fixed point in the domain.
+ * \param	rangePoint1: A known fixed point in the range.
+ * \param	rangePoint2: Another known fixed point in the range.
```
DanRStevens: What if we throw in the word "corresponding" here:
"corresponding known fixed point"

test/MathUtils.test.cpp
cugone: I like the explicit types version. It's cleaner.

PR #419, opened 2020-01-30
Enable warnings as errors for Linux and MacOS

DanRStevens:
Looks like the NAS2D builds for Linux and MacOS are all clean. Should try to keep them that way.


cugone:
Hmm...should I turn Visual Studio up to Level 4 Warnings and enable warnings as errors to match parity?

DanRStevens:
The Windows build already has a warning about `SDL_bool`. We can't enable warnings as errors until that is dealt with. We should experiment with a higher warning level in a branch though.

PR #429, opened 2020-02-05
Make Configuration use explicit types for private member function parameters

DanRStevens:
Using explicit types eliminates the need for type casts, and being more specific about them means error checks can be removed. A forward declaration is enough to use an explicit type, without needing to `#include` implementation details from the header.

In hindsight, I probably should have swapped the order of the commit that removes the checks and the commit that changes the declared parameter types. Small point though, since all calls to the private method only ever pass objects of the correct type.

### DIFF
--- a/include/NAS2D/Configuration.h
+++ b/include/NAS2D/Configuration.h
@@ -9,8 +9,6 @@
 // ==================================================================================
 #pragma once
 
-#include "Exception.h"
-
 #include <string>
 #include <map>
 
--- a/include/NAS2D/Configuration.h
+++ b/include/NAS2D/Configuration.h
@@ -11,6 +11,7 @@
 
 #include "Exception.h"
 
+#include <string>
 #include <map>
 
 namespace NAS2D {
--- a/include/NAS2D/Configuration.h
+++ b/include/NAS2D/Configuration.h
@@ -89,9 +89,9 @@ private:
 
 	bool readConfig(const std::string& filePath);
 
-	void parseGraphics(NAS2D::Xml::XmlElement *node);
-	void parseAudio(NAS2D::Xml::XmlElement *node);
-	void parseOptions(NAS2D::Xml::XmlElement *node);
+	void parseGraphics(NAS2D::Xml::XmlElement* node);
+	void parseAudio(NAS2D::Xml::XmlElement* node);
+	void parseOptions(NAS2D::Xml::XmlElement* node);
 
 	Options mOptions{};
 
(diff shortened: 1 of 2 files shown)


include/NAS2D/Configuration.h
```diff
@@ -85,9 +89,9 @@ class Configuration
 
 	bool readConfig(const std::string& filePath);
 
-	void parseGraphics(void *node);
-	void parseAudio(void *node);
-	void parseOptions(void *node);
+	void parseGraphics(NAS2D::Xml::XmlElement *node);
+	void parseAudio(NAS2D::Xml::XmlElement *node);
+	void parseOptions(NAS2D::Xml::XmlElement *node);
```
cugone: Consistency and convention.
Lines 74-76 have their reference ampersands on the left as well.
Move the asterisk to the left. These are types, not values that need to be dereferenced.

src/Configuration.cpp
```diff
@@ -333,25 +311,15 @@ void Configuration::parseAudio(void* _n)
...
-		std::cout << "Unexpected XML tag '" << static_cast<XmlNode*>(_n)->value() << "' found in configuration file while processing the '<graphics>' element." << std::endl;
-		return;
-	}
-
-	for (auto xmlNode = element->iterateChildren(nullptr);
-		 xmlNode != nullptr;
-		 xmlNode = element->iterateChildren(xmlNode))
+	for (auto setting = element->firstChildElement();
+		 setting != nullptr;
+		 setting = setting->nextSiblingElement())
```
cugone: That reminds me, need to add ForEachElement/Attribute functions. This type of code is just gross. :P

include/NAS2D/Configuration.h
cugone: Additionally, the implementation of these functions also has them on the left.

src/Configuration.cpp
DanRStevens: I know, right. It would be so much nicer.

include/NAS2D/Configuration.h
DanRStevens: I suppose technically that was pre-existing, and such a style change is outside the scope of this PR, though it may indeed be a good time to change it. A quick grep shows very few occurrences of this style.

DanRStevens: Actually, considering I changed the implementation file and put the `*` on the left there, I should probably be consistent about it in the header file too. Anyway, change has already been made.

Issue #390, opened 2020-01-22
Remove Filesystem's argv_0 parameter


cugone:
The recent changes to Filesystem Update use OS-specific calls to get the path to the exe. As such, the argv_0 argument is no longer needed.

DanRStevens:
I'm not convinced it is no longer needed. Use of this parameter depended on the platform. On Windows it was not needed. It may still be needed on Linux. I'm less certain about MacOS, though it may (optionally?) function similar to Linux.

cugone:
The system now uses OS calls exclusively for all platforms the variable literally isn't used at all. Therefore, it's not needed.

DanRStevens:
Came across this, which right at the bottom has code for finding the executable path:
https://www.doxsey.net/blog/a-response-to-hello-world

Interesting how it uses `/proc/self` on Linux. That may be a good indication it is a reliable way to get the path. Would love to find some documentation that addresses how portable it is. Curiously the code uses a different path for BSD derived operating systems.

As a side note, this makes me curious how many and what operating systems are supported by PhysFS.

Related, which do we care about. We currently target 3 main OSes. Though the decision is largely because only those 3 ever came to mind.

cugone:
I could have it try to use the current implementation and if it doesn't exist, fall back to `/proc/self/exe` and follow its symlinks all the way to a real file.

DanRStevens:
This is currently blocked by:
- Issue #212
- Replacing `Filesystem`'s implementation so it doesn't use PhysFS (What PR #243 is attempting)

PR #435, opened 2020-02-06
Add Filesystem special folders

### LINKED ISSUE #317: Filesystem special folders
Currently the `Filesystem` constructor default mounts several special real filesystem folders. These include the executable's folder (read access), and a user specific write folder (read/write access).

These folders currently all come as a bundle. We should unpack this bundle, so code can have greater flexibility and control over which folders are mounted. Initially, this may be more of an internal change, where the default set of mounted folders is not changed, but rather flexibility is expanded for future changes.

To accomplish this, we should have methods to return certain special real filesystem folders. At a minimum, this should include the executable's folder, and a user specific write folder. We might also include the user's home folder. These should return real filesystem paths, suitable to be passed to the `mount` method (or an `unmount` method).

As a convenience, we should perhaps still provide a helper method that can mount the default set of folders as a one liner. The `Filesystem` constructor could delegate to this method to setup the default set of mount folders. In the future, if the `Filesystem` constructor were to remove default mounts, this method could be used by programs to setup mount points compatible with the old default.

Related: Issue #212
This issue is concerned with creating public methods to determine special real filesystem folders. It is expected the implementations of these methods will defer to PhysFS for the implementation. This provides the framework for issue #212, which is concerned with replacing the PhysFS implementation of those methods with a custom implementation.

DanRStevens:
Closes #317

Add `Filesystem` special folders `basePath` and `prefPath`.

This allows for delayed mounting of special filesystem folders, or unmounting of default mounted folders.

This intentionally uses PhysFS functions. Doing this without the use of PhysFS is to be handled by later issues, such as #212. This is intended to create the structure for later changes to follow.

This should help unblock work on: #212, #320

### DIFF
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -34,6 +34,7 @@ public:
 	~Filesystem();
 
 
+	std::string basePath() const;
 	std::string dataPath() const;
 	std::string workingPath(const std::string& filename) const;
 	StringList searchPath() const;
--- a/include/NAS2D/Filesystem.h
+++ b/include/NAS2D/Filesystem.h
@@ -35,6 +35,8 @@ public:
 
 
 	std::string basePath() const;
+	std::string prefPath() const;
+
 	std::string dataPath() const;
 	std::string workingPath(const std::string& filename) const;
 	StringList searchPath() const;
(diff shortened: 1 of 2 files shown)


cugone:
Also closes #390

DanRStevens:
Err, this doesn't close #390. The `argv_0` parameter is still there, and still being used by `PHYSFS_init`. That issue is unrelated to this change.

PR #436, opened 2020-02-06
Add `Filesystem::dirSeparator()` method

DanRStevens:
Add `Filesystem::dirSeparator()` method
- Windows: `\\`
- Linux: `/`
- Mac: `:`

Reference:
[`PHYSFS_getDirSeparator()`](https://icculus.org/physfs/docs/html/physfs_8h.html#ad1dc17d387c45e9155d1f0c348a7cbcc)

This may not find much direct use, though is perhaps useful internally for path manipulations, and for unit test code. In particular, unit tests which rely on platform specific folders can potentially use this to abstract away platform differences in unit test values.

### DIFF
--- a/test/Filesystem.test.cpp
+++ b/test/Filesystem.test.cpp
@@ -111,3 +111,10 @@ TEST_F(FilesystemTest, mountUnmount) {
 
 	EXPECT_THROW(fs.mount("nonExistentPath/"), std::runtime_error);
 }
+
+TEST_F(FilesystemTest, dirSeparator) {
+	// Varies by platform, so we can't know the exact value ("/", "\", ":")
+	// New platforms may choose a new unique value
+	// Some platforms may not even have a hierarchal filesystem ("")
+	EXPECT_NO_THROW(fs.dirSeparator());
+}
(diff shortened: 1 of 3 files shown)


test/Filesystem.test.cpp
```diff
@@ -17,6 +17,16 @@ class FilesystemTest : public ::testing::Test {
 };
 
 
+TEST_F(FilesystemTest, basePath) {
+	// Result is a path, so should end with a directory separator
```
cugone: Result is a *directory* not a filepath. Not a big deal since this is a comment but should be consistent.

test/Filesystem.test.cpp
```diff
@@ -17,6 +17,16 @@ class FilesystemTest : public ::testing::Test {
...
+	EXPECT_THAT(fs.basePath(), testing::EndsWith(fs.dirSeparator()));
+}
+
+TEST_F(FilesystemTest, prefPath) {
+	// Result is a path, so should end with a directory separator
```
cugone: Result is a *directory* not a filepath. Not a big deal since this is a comment but should be consistent.

test/Filesystem.test.cpp
DanRStevens: Hmm, might have been leftover from when I did a bit of renaming.

PR #439, opened 2020-02-06
XML Validation

### LINKED ISSUE #434: Add XML Validation
Add a validation function that verifies an XML element for correctness.
```
void ValidateXmlElement(const XMLElement& element,
                        const std::string& name,
                        const std::string& requiredChildElements,
                        const std::string& requiredAttributes,
                        const std::string& optionalChildElements = std::string{},
                        const std::string& optionalAttributes = std::string{}) noexcept;
```
 The function should:
- [x] Verify the name is correct. Pop an error dialog then exit the program if the names do not match.
- [x] Ensure all required child elements exist. Pop an error dialog then exit the program if an element is missing.
- [x] Ensure all required attributes exist. Pop an error dialog then exit the program if an attribute is missing.
- [x] Verify any extra child elements are in the optional list. Pop a warning dialog and/or log in debug mode, but otherwise ignore extra elements.
- [x] Verify any extra attributes are in the optional list. Pop a warning dialog and/or log in debug mode, but otherwise ignore extra elements.

cugone:
Closes #434

Adds a function that validates XML elements for correctness.


include/NAS2D/DataUtils.h
```diff
@@ -0,0 +1,23 @@
...
+	{
+		class XmlElement;
+
+		void validateXMLElement(
+			const XmlElement& element,
+			const std::string& name,
+			const std::string& requiredChildElements,
+			const std::string& requiredAttributes,
+			const std::string& optionalChildElements = std::string{},
+			const std::string& optionalAttributes = std::string{});
```
DanRStevens: I suppose a matter of style, though I might suggest spilling the closing parenthesis onto it's own line:
```cpp
void validateXMLElement(
	const XmlElement& element,
	const std::string& name,
	const std::string& requiredChildElements,
	const std::string& requiredAttributes,
	const std::string& optionalChildElements = std::string{},
	const std::string& optionalAttributes = std::string{}
);
```
That way the indentation is a little more consistent with the parsed structure.

include/NAS2D/DataUtils.h
```diff
@@ -0,0 +1,23 @@
...
+
+	} // namespace Xml
+
+} // namespace DataUtils
```
DanRStevens: Missing newline at end of file.

src/DataUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+	{
+		// TODO: Make this a function with [[noreturn]] attribute
+#if defined(_WIN32)
+		std::ostringstream err_ss;
+		err_ss << "Xml Validation failed. Element name is required.";
+		::MessageBoxA(nullptr, err_ss.str().c_str(), "Fatal Error", MB_ICONERROR | MB_OK);
+#else
+		// TODO: Add non-windows version.
+#endif
+		exit(0);
```
DanRStevens: Ehh, such code is of very questionable use in a library. People generally don't want their calls to some library to suddenly terminate their app.

The error handling on this really needs to be reworked. I would suggest throwing an exception (`std::runtime_error` or derived) with the desired error message.

src/DataUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include "NAS2D/Xml/XmlAttribute.h"
+#include "NAS2D/Xml/XmlElement.h"
+
+#include <algorithm>
+#include <sstream>
+
+#if defined(_WIN32)
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
```
DanRStevens: With programmatic error reporting (such as exceptions), platform specific code such as this could be removed.

src/DataUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	const std::string elemName = element.value();
+	if (elemName != name)
+	{
+		// TODO: Make this a function with [[noreturn]] attribute
+#if defined(_WIN32)
+		std::ostringstream err_ss;
+		err_ss << "Xml Validation failed. Element name " << elemName << " does not match valid name " << name;
+		::MessageBoxA(nullptr, err_ss.str().c_str(), "Fatal Error", MB_ICONERROR | MB_OK);
```
DanRStevens: Simple string concatenation would probably be easier than using `ostringstream`. One caveat is there is no `operator+` that takes two `const char*`. At least one of the two operands must be `std::string` or similar. If that's not that case, you can always wrap a `const char *` with a `std::string` constructor call.

src/DataUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+#endif
+		exit(0);
+	}
+
+	// Get list of required/optional attributes/children
+	// Sort
+	// Remove duplicates
+	// Rational for not using std:set:
+	// Profiled code takes average of 10 microseconds to complete.
+	auto requiredChildNames = NAS2D::split(requiredChildElements);
```
DanRStevens: I'm actually wondering why this wasn't passed in as an array (`std::vector<std::string>`).

src/DataUtils.cpp
```diff
@@ -0,0 +1,211 @@
...
+	}(); // Immediately Invoked Initializing Lambda (IIIL)
+	std::sort(std::begin(actualAttributeNames), std::end(actualAttributeNames));
+	actualAttributeNames.erase(std::unique(std::begin(actualAttributeNames), std::end(actualAttributeNames)), std::end(actualAttributeNames));
+
+	StringList extraChildElementNames{};
+	std::set_difference(
+		std::begin(actualChildNames), std::end(actualChildNames),
+		std::begin(requiredChildNames), std::end(requiredChildNames),
+		std::back_inserter(extraChildElementNames));
+	std::sort(std::begin(extraChildElementNames), std::end(extraChildElementNames));
```
DanRStevens: I may need to look at this section again later.

My thought here is the code is effectively doing a tree diff. There is an expected tree, and an actual tree. When doing a tree diff, if any required nodes are missing, it's an error. If any extra (optional) nodes are present, it (may be) a warning.

src/DataUtils.cpp
cugone: So you don't have to build a a vector beforehand and the list can just be a CSV list C-string.

See: https://github.com/lairworks/nas2d-core/issues/434#issuecomment-583219252

src/DataUtils.cpp
cugone: There's no tree involved. There are two flat lists of values. A list of the unique names that actually appear in the file, and a list of unique names that are required *or* a list of names that are optional. The result `set_difference` returns a list of missing or extra names, depending on the order of the arguments (set_difference is not commutative). If an item is missing from the required set, that's an error. If an extra item is found that's not in the optional set, that's a warning.

src/DataUtils.cpp
cugone: These are workarounds for not having Leeor's `doNonFatalErrorMessage` and `doAlertMessage` functions available in the engine. I'd really like to add those instead of the above.

src/DataUtils.cpp
cugone: It's also faster by a whole order of magnitude. Just tested `1'000'000` iterations of equivalent code using `std::ostringstream` vs `std::to_string`. `ostringstream` took around 3 seconds, `to_string` took around a third of a second.

DanRStevens: Oh hey, sweet. I didn't realize there would be that kind of performance gain.

src/DataUtils.cpp
DanRStevens: I was just thinking. On another project, I recently discovered that `std::string_view` is `constexpr` compatible, even though `std::string` is not. In that project I was able to create an array of string_views at compile time. That might be one way to set a static list of parameters without incurring runtime cost.

I may have been using `std::array` rather than `std::vector` though, so that may impact the usefulness of the technique.

src/DataUtils.cpp
DanRStevens: I meant conceptually, in the general case, of a deeply nested tree. I believe the function parameters here limits the number of levels it will actually test.

cugone: Yes, it intentionally only tests the current level. This allows for different classes handling their own XML serialization by just passing in the current element.

src/DataUtils.cpp
cugone: You might get away with `std::string_view` parameters into the validation function but they are still being split up into separate strings. So at most you're saving 5 string constructions where SSO in most typical cases would handle 3 of them.

include/NAS2D/DataUtils.h
cugone: I blame Visual Studio. I just spent more than a few seconds trying to get it to not do that in a different file.

cugone:
I [commented on it](https://github.com/lairworks/nas2d-core/issues/434#issuecomment-583790180) 2 days ago; you've yet to reply. ;)

PR #443, opened 2020-02-08
Add `Filesystem::mountWrite` method

DanRStevens:
This addresses: https://github.com/lairworks/nas2d-core/issues/320#issuecomment-583734421

Add ability to set a write folder, other than the initial default.

----

Would appreciate some feedback on the naming, and if read/write access should be tied together.

### DIFF
--- a/src/Filesystem.cpp
+++ b/src/Filesystem.cpp
@@ -109,6 +109,24 @@ void Filesystem::mount(const std::string& path) const
 
 
 /**
+ * Mount a folder with write access (includes read access)
+ *
+ * \param path	File path to add.
+ */
+void Filesystem::mountWrite(const std::string& path) const
+{
+	// Mount for read access
+	mount(path);
+
+	// Mount for write access
+	if (PHYSFS_setWriteDir(path.c_str()) == 0)
+	{
+		throw std::runtime_error(std::string("Couldn't add write folder '") + path + "': " + PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode()));
+	}
+}
+
+
+/**
  * Removes a directory or supported archive from the Search Path.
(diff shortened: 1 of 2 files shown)


src/Filesystem.cpp
```diff
@@ -108,6 +108,24 @@ void Filesystem::mount(const std::string& path) const
 }
 
 
+/**
+ * Mount a folder with write access (includes read access)
+ *
+ * \param path	File path to add.
+ */
+void Filesystem::mountWrite(const std::string& path) const
```
cugone: This could be better named as `mountReadWrite` and another function could be created: `setWriteDirectory` that just calls `PHYSFS_setWriteDir`.

DanRStevens: Hmm, this makes me think the existing method should be renamed for symmetry. Perhaps:
- `mountRead`
- `mountReadWrite`

I want to avoid `setWriteDirectory`, since it implies a single write directory. I think we might go more general than that in the future.

cugone: Yeah, adding a `mountRead` that delegates to `mount` would be useful for understandability. You don't *have* to rename `mount` itself. It's up to you if you want to break the API.

DanRStevens: Hmm, good point. I sort of want to avoid breaking compatibility in this PR. I'd prefer it to be focused on adding a new feature.

I'm hesitant to add alias method names at this point. Maybe I should just focus on the new method, with the intention of eventually renaming the old method.

One potential future design path is to cleanup the `Game` class, so it is usable by OPHD. The `Game` class initializes the `Filesystem`, so that could bring the `Filesystem` calls back into the library, and eliminate further downstream changes.

PR #447, opened 2020-02-09
Replace include of `NAS2D` with components used

DanRStevens:
Reduce the transitive dependencies of `Sprite.h`, which results in fewer recompiles during an incremental build.

This was noticed while looking at downstream issue: https://github.com/OutpostUniverse/OPHD/issues/134

The change has a fairly significant impact on the dependency graph for the downstream project. It has also uncovered numerous places where downstream files were not including all the headers they were actually using.

### DIFF
--- a/include/NAS2D/Resources/Sprite.h
+++ b/include/NAS2D/Resources/Sprite.h
@@ -10,7 +10,14 @@
 
 #pragma once
 
-#include "../NAS2D.h"
+#include "Image.h"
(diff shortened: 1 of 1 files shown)


DanRStevens:
Just a note that downstream updates should be applied first before this is merged. As it stands now, downstream projects may break due to missing includes. Testing against this branch is one way of finding the places that are broken.

PR #450, opened 2020-02-10
Add Vector::to<NewBaseType>()

DanRStevens:
Reference: #448 for corresponding `Rectangle` update
Reference: #449 for corresponding `Point` update

Add `Vector::to<NewBaseType>()` method, as a new shorter easier to use way of converting a `Vector` to a new underlying base type.

- No need for static_cast in client code
- No need to specify the full struct type (in addition to the underlying NewBaseType)
- No need to specify the namespace

### DIFF
--- a/include/NAS2D/Renderer/Vector.h
+++ b/include/NAS2D/Renderer/Vector.h
@@ -74,6 +74,11 @@ struct Vector {
 		};
 	}
 
+	template <typename NewBaseType>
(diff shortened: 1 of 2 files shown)


DanRStevens:
Whoops. Merged and deleted the branch before the unit tests had a chance to run. Naturally they failed when they couldn't find the now deleted branch.

Issue #451, opened 2020-02-10
Move message box functions from OPHD to NAS2D


cugone:
Ref: https://github.com/OutpostUniverse/OPHD/issues/208

PR #452, opened 2020-02-10
Implements MessageBox functions engine-side

cugone:
Implements various message box functions engine-side.

Doesn't completely close the related issue (#451) yet. A separate PR will need to be created to refactor the uses in OPHD.

### DIFF
--- a/include/NAS2D/MessageBox.h
+++ b/include/NAS2D/MessageBox.h
@@ -15,6 +15,7 @@ namespace NAS2D
 
 	void MessageBox_Okay(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 	bool MessageBox_OkayCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
+	
 	bool MessageBox_YesNo(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 	int MessageBox_YesNoCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 
--- a/include/NAS2D/MessageBox.h
+++ b/include/NAS2D/MessageBox.h
@@ -16,5 +16,6 @@ namespace NAS2D
 	void MessageBox_Okay(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 	bool MessageBox_OkayCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 	bool MessageBox_YesNo(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
+	int MessageBox_YesNoCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 
 } // namespace NAS2D
--- a/include/NAS2D/MessageBox.h
+++ b/include/NAS2D/MessageBox.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <string_view>
+
+namespace NAS2D
+{
+
+	enum class MessageBoxSeverityLevel
+	{
+		Information,
+		Question,
+		Warning,
+		Fatal
+	};
+
+	void MessageBox_Okay(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
+
+} // namespace NAS2D
--- a/include/NAS2D/MessageBox.h
+++ b/include/NAS2D/MessageBox.h
@@ -22,4 +22,6 @@ namespace NAS2D
 	bool MessageBox_RetryCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 	int MessageBox_AbortRetryIgnore(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 
+	int MessageBox_CancelTryContinue(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
+
 } // namespace NAS2D
--- a/include/NAS2D/MessageBox.h
+++ b/include/NAS2D/MessageBox.h
@@ -14,5 +14,6 @@ namespace NAS2D
 	};
 
 	void MessageBox_Okay(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
+	bool MessageBox_OkayCancel(std::string_view messageTitle, std::string_view messageText, MessageBoxSeverityLevel severity);
 
 } // namespace NAS2D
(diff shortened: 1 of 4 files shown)


cugone:
> First off, these functions should **never** be called from within the NAS2D library itself. Ever.

Why?

> Additionally, these methods don't test easily though unit tests.

Not everything will be. Testing the graphics code and shaders, for example, will need to be tested manually.

> 
> Further, having these functions available can encourage the use of modal popup boxes, which I really don't want to do. I would much prefer non-modal dismissable popups, much like the Windows tray messages.

@ldicker83 wants modal message boxes.

Toast messages are for when your application is not in use.
See: https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/send-local-toast
> A toast notification is a message that an app can construct and deliver to the user while they are not currently inside your app.

Toast messages *do not work* in fullscreen, exclusive or non-exclusive. Windows disables them.

Users can disable them. This is not ideal for error messages:
https://www.askvg.com/windows-10-tip-disable-toast-notifications-and-enable-classic-balloon-tips/


> With that said, such messages are commonly used in early prototypes or while debugging, and we don't currently have a non-modal alternative to offer.

There are several times where if something fails and the program literally cannot continue without user input (or at all!) a message box popping up is preferred rather than a hard crash with the obscure "Unhandled exception" error.

An example is compiling shaders at runtime. A library feature of the engine. If the shader fails to compile, I want a message saying it failed and a Retry option to recompile after I fix the shader. These solve that problem.

> I don't much like the use of platform detects. Since SDL has support for popups, and the code is already there for other platforms, I'm kind of wondering why it isn't just used consistently for all platforms. I think it would be better to get rid of Windows specific code.
> 

We're trying to move away from SDL handling *everything*. In this case, it's only here for Linux. MacOSX has their own too.

SDL's implementation of message boxing is horrible. Windows has more options with less code. I'd prefer Windows' one-liners than SDL's multi-function call and multi-struct declarations. Failing that, these functions turn the god-awful SDL code into a sensible one-liner like Windows' versions.

> I think I would prefer "Ok", rather than "Okay". The use of "Ok" is just too widespread.

This was an attempt to keep inline with Windows' `MB_*` arguments that denote what type of message box you are getting.
See: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messagebox

> There are some minor code formatting issues I see, such as how multiline structure initializers are formatted, in particular with line leading `,` rather than trailing `,`, and line trailing `};`.

Noted.

DanRStevens:
The error popup methods should never be called from within the library because it's a major pain in the ass to library clients when their programs start displaying popups or crashing because of code deep inside a library method they called.

The client app should always be in control of error reporting and termination. Any error code must be written such that the error can be delivered to client code so the client code can decide how to act upon it there.

Your shader example sounds like a higher app level concern, not a library concern.

----

To clarify, I don't think Leeor *wants* modal popup boxes. I think he's resigned to use them due to other difficulties. Should those difficulties be worked around through some other means, we may very well want to get rid of modal popup boxes.

----

I had mentioned Windows tray messages as an example of non-modal dismissable messages. Not for their use when the app is not active. There are other examples of in-app non-modal dismissable messages. Browsers sometimes use such messages, such as when the page slides down so the browser can insert a short message about some warning across the top of the page.

----

I'm not aware of any plan to move away from SDL. It's not even feasible to move away from SDL. Particularly not if we want things to be cross platform, which I certainly do.

As for one liners, that's what the methods are for. If the code inside them was only one line, there wouldn't be much need for the method.

cugone:
Okay. That sounds like a good reason. As for my shader example, it was just an example. The same can easily be accomplished as Unity does it: No messages at all, your shader just outputs an ugly magenta (255, 0, 255) solid color on whatever you're trying to attach the shader to and you get to figure out why.

I was also going to mention the XMLValidation where loud failures are a must. But again, throwing accomplishes the same thing.

---

Yeah, makes sense. I've been moving in that direction with the UI. It's slow-going. I'm not a UI dev.

---

Ohhh....

---

The OpenGL refactor will be moving away from SDL in favor of up-to-date GL 4.x programmable pipelines with a material/shader system and away from SDL's GL 2.0 implementation. Keeping SDL for cross-platform application window management is required though. Windows vs SDL implementations can easily be encapsulated inside a Window class and an RHI layer.

---

DanRStevens:
@ldicker83, it might be good if you could weigh in on direction here. I don't think you need to do a line by line check of the code, just a general check if this is something we are interested in adding to the library, or leaving in the game, and perhaps if the public API is acceptable.

cugone:
The nas2d-tests repo uses OS-specific calls and defines to use Windows' `MessageBox` function for reporting errors. This code would get rid of that and make it cross-platform. Realistically, any downstream project that wraps `Game` in a try-catch block can use them now.

```
//...
try
{
	NAS2D::Game game("NAS2D Graphics Test", "NAS2D_GraphicsTest", "LairWorks", argv[0]);
	game.go(new TestGraphics());
}
catch(std::exception& e)
{
	std::cout << e.what() << std::endl;
	NAS2D::MessageBox_Okay("Application Error", e.what(), MessageBoxSeverityLevel::Error);
        return 1;
}
catch(...)
{
	std::cout << "Unknown error occurred." << std::endl;
	NAS2D::MessageBox_Okay("Unexpected Error", "Unknown error occurred", MessageBoxSeverityLevel::Error);
	return 2;
}
//...
```

cugone:
OPHD uses a MessageBox function with a confusing name at the very end of `main.cpp`:

```
            cf.save(); // force configuration to save any changes.
        }
        catch(const std::exception& e)
        {
            doNonFatalErrorMessage("Application Error", e.what());
        }

        #ifdef NDEBUG
        filestr.close();
        #endif

        SDL_Quit();
```

It reads like: Catch exception, Pop a non-fatal message box: "Application Error"....uhhhh.

Clearly the name of the function is misleading.

DanRStevens:
I believe it means non-fatal in the sense that the message popup routine doesn't try to terminate the application. It just displays a message box.

Though yes, we could stand to improve the naming and increase consistency between the various message functions.

cugone:
> Do the SDL methods produce ugly or non-native looking popup boxes on Windows?

As written they do (using the `SDL_ShowSimpleMessageBox` function). Part of the refactor will be to use the fully-featured `SDL_ShowMessageBox`, but that requires filling out SDL structs beforehand. They aren't needed for Windows so I'd rather they be compiled out on Windows, hence, the platform detects.

The platform detects are safely hidden behind function calls so if you don't like them, don't look at them. :P

DanRStevens:
That kind of sounds like premature optimization of code that isn't performance critical.

I'd rather have less code to maintain in this case.

----

I took a look at the structs. They do seem a bit unfortunate with how they are filled out. We could use predefined `constexpr` values for the `buttons` and `colorScheme` fields, though I'm not sure of an easy convenient inline way to set all of `numbuttons`, `buttons`, and `colorScheme`, while also being able to freely set `flags`, `window`, `title`, and `message`. I kind of wish they'd used sub-structures or something there.

PR #458, opened 2020-02-12
Re-adds now-supported clang-format options

cugone:
There is an option in Visual Studio to only apply clang-format changes on Keyboard combo `Ctrl+K, Ctrl+D` to the whole document or `Ctrl+K,Ctrl+F` for the current selection only.

`Tools > Options > Text Editor > C/C++ > Formatting > General > Clang Format execution > check "Run ClangFormat only for manually invoked formatting commands"`

This would have been good to know earlier... Oh well.

Here's a [reference guide](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) to the commands.

There are no anchor links, sorry. :(

You'll have to Ctrl+F search for the command....or just read through the entire document like I did. :/

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -68,7 +68,7 @@ SpaceBeforeCpp11BracedList: false
 SpaceBeforeCtorInitializerColon: true
 SpaceBeforeInheritanceColon: true
 SpaceBeforeParens: ControlStatements
-SpaceBeforeRangeBasedForLoopColon: false
+SpaceBeforeRangeBasedForLoopColon: true
 SpaceInEmptyParentheses: false
 SpacesInAngles: false
 SpacesInCStyleCastParentheses: false
--- a/.clang-format
+++ b/.clang-format
@@ -4,9 +4,7 @@ AllowShortBlocksOnASingleLine: true
 AllowShortCaseLabelsOnASingleLine: true
 AllowShortIfStatementsOnASingleLine: true
 
-#Unsupported in VS2017-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#AllowShortLambdasOnASingleLine: All
-
+AllowShortLambdasOnASingleLine: All
 AlwaysBreakTemplateDeclarations: true
 BinPackArguments: true
 BinPackParameters: false
@@ -56,8 +54,7 @@ IncludeCategories:
     - Priority: 1
 
 IndentCaseLabels: false
-#Unsupported in VS-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#IndentPPDirectives: BeforeHash
+IndentPPDirectives: BeforeHash
 IndentWidth: 4
 IndentWrappedFunctionNames: false
 NamespaceIndentation: All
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -67,21 +64,11 @@ SortIncludes : true
...
-
+SpaceBeforeCpp11BracedList: false
+SpaceBeforeCtorInitializerColon: true
+SpaceBeforeInheritanceColon: true
 SpaceBeforeParens: ControlStatements
-
-#Unsupported in VS2017-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#SpaceBeforeRangeBasedForLoopColon: false
-
+SpaceBeforeRangeBasedForLoopColon: false
```
DanRStevens: I suspect we want this to be `true`. Existing uses of range for put a space before the colon.

.clang-format
```diff
@@ -67,21 +64,11 @@ SortIncludes : true
...
-#Unsupported in VS2017-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#SpaceBeforeCpp11BracedList: false
-
-#Unsupported in VS2017-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#SpaceBeforeCtorInitializerColon: true
-
-#Unsupported in VS2017-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#SpaceBeforeInheritanceColon: true
-
+SpaceBeforeCpp11BracedList: false
```
DanRStevens: Would be nice to know what exactly this means. Maybe a reference link would help.

In particular, does this setting describe existing practice?

.clang-format
```diff
@@ -56,8 +54,7 @@ IncludeCategories:
     - Priority: 1
 
 IndentCaseLabels: false
-#Unsupported in VS-included ClangFormat 6.0 we'll see about VS2019's ClangFormat 7.0
-#IndentPPDirectives: BeforeHash
+IndentPPDirectives: BeforeHash
```
DanRStevens: I have no idea what this one does.

.clang-format
cugone: Ref: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
There are no anchor links, sorry. :(
You'll have to Ctrl+F search for the command.

> If true, a space will be inserted before a C++11 braced list used to initialize an object (after the preceding identifier or type).
```
    true:                                  false:
    Foo foo { bar };               vs.     Foo foo{ bar };
    Foo {};                                Foo{};
    vector<int> { 1, 2, 3 };               vector<int>{ 1, 2, 3 };
    new int[3] { 1, 2, 3 };                new int[3]{ 1, 2, 3 };
```

cugone: The false option does.

.clang-format
cugone: Ref: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
>IndentPPDirectives (PPDirectiveIndentStyle)
>
>    The preprocessor directive indenting style to use.
>
> Possible values:
>
- `PPDIS_None` (in configuration: `None`) Does not indent any directives.
```
        #if FOO
        #if BAR
        #include <foo>
        #endif
        #endif
```
- `PPDIS_AfterHash` (in configuration: `AfterHash`) Indents directives after the hash.
```
        #if FOO
        #  if BAR
        #    include <foo>
        #  endif
        #endif
```
- `PPDIS_BeforeHash` (in configuration: `BeforeHash`) Indents directives before the hash.
```
        #if FOO
          #if BAR
            #include <foo>
          #endif
        #endif
```

.clang-format
cugone: Point. Oversight on my part.

PR #460, opened 2020-02-13
warpMouse emits a mouseMotion event

### LINKED ISSUE #41: EventHandler::warpMouse() doesn't raise a mouse motion event
This is a basic oversight but it results in incorrect behavior. When the ```warpMouse()``` function is called, no event for a mouse position (mouse motion) event is raised.

To get correct behavior, when this function is called the ```EventHandler``` should raise a Mouse Motion event with the current coordinates and a ```{ 0, 0 }``` relative coordinate change.

cugone:
Closes #41

### DIFF
--- a/src/EventHandler.cpp
+++ b/src/EventHandler.cpp
@@ -516,7 +516,11 @@ void EventHandler::releaseMouse()
  */
 void EventHandler::warpMouse(int x, int y)
 {
-	if (underlyingWindow) { SDL_WarpMouseInWindow(underlyingWindow, x, y); }
+	if (underlyingWindow)
+	{
+		mMouseMotionEvent.emit(x, y, 0, 0);
+		SDL_WarpMouseInWindow(underlyingWindow, x, y);
+	}
 }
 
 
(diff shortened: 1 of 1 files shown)


src/EventHandler.cpp
```diff
@@ -516,7 +516,11 @@ void EventHandler::releaseMouse()
  */
 void EventHandler::warpMouse(int x, int y)
 {
-	if (underlyingWindow) { SDL_WarpMouseInWindow(underlyingWindow, x, y); }
+	if (underlyingWindow)
+	{
+		mMouseMotionEvent.emit(x, y, 0, 0);
+		SDL_WarpMouseInWindow(underlyingWindow, x, y);
```
DanRStevens: I'm wondering if the order of these two lines should be swapped. It probably doesn't matter, though conceptually I would have raised the event after the action took place. That's what I would have expected anyway.

DanRStevens:
It appears the `warpMouse` method isn't currently used anywhere. I guess that means it had incorrect behaviour on a feature that wasn't fully developed and merged.

Issue #399, opened 2020-01-25
Review uses of `std::pair`


DanRStevens:
We should review uses of `std::pair` to see if `Point` or `Vector` may be a better alternative.

A quick `grep` shows these areas of interest:
```
include/NAS2D/Resources/Image.h:67:	std::pair<int, int> _size{0,0};		   /**< Width/Height information about the Image. */
include/NAS2D/Resources/Image.h:68:	std::pair<uint32_t, uint32_t> _center{0u,0u}; /**<  */
include/NAS2D/StringUtils.h:25:	std::pair<std::string, std::string> splitOnFirst(const std::string& str, char delim);
include/NAS2D/StringUtils.h:26:	std::pair<std::string, std::string> splitOnLast(const std::string& str, char delim);
include/NAS2D/Xml/XmlBase.h:118:	std::pair<int, int> location;
include/NAS2D/Xml/XmlDocument.h:58:	std::pair<int, int>	_errorLocation;
src/StringUtils.cpp:85:std::pair<std::string, std::string> NAS2D::splitOnFirst(const std::string& str, char delim)
src/StringUtils.cpp:98:std::pair<std::string, std::string> NAS2D::splitOnLast(const std::string& str, char delim)
src/Xml/XmlParser.cpp:105:	const std::pair<int, int>& cursor() const { return _cursor; }
src/Xml/XmlParser.cpp:115:	std::pair<int, int>	_cursor;
```

cugone:
Closed via #461 and #462

DanRStevens:
Small point, but you should generally wait until after changes have been merged before closing.

PR #462, opened 2020-02-13
Refactors xml uses of`std::pair`

cugone:
Ref #399

### DIFF
--- a/include/NAS2D/Xml/XmlBase.h
+++ b/include/NAS2D/Xml/XmlBase.h
@@ -94,7 +94,14 @@ public:
 		XML_ERROR_STRING_COUNT
 	};
 
+	struct ParseLocation
+	{
+		int row{0};
+		int column{0};
+	};
+
 protected:
+
 	static const char* skipWhiteSpace(const char*);
 
 	inline static bool white_space(char c) { return (isspace(static_cast<unsigned char>(c)) || c == '\n' || c == '\r'); }
@@ -116,7 +123,7 @@ protected:
 
 	inline static int toLower(int v) { return tolower(v); }
 
-	std::pair<int, int> location;
+	ParseLocation location;
 
 private:
 	struct Entity
--- a/src/Xml/XmlParser.cpp
+++ b/src/Xml/XmlParser.cpp
@@ -608,8 +608,7 @@ const char* XmlDocument::parse(const char* p, void* prevData)
 	// so that parsing starts from the pointer we are given.
 	if (prevData)
 	{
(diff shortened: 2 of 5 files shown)


src/Xml/XmlParser.cpp
```diff
@@ -604,20 +603,16 @@ const char* XmlDocument::parse(const char* p, void* prevData)
...
 	{
-		location.first = static_cast<XmlParsingData*>(prevData)->_cursor.first;
-		location.second = static_cast<XmlParsingData*>(prevData)->_cursor.second;
-	}
-	else
-	{
-		location.first = 0;
-		location.second = 0;
+		location.row = static_cast<XmlParsingData*>(prevData)->_cursor.row;
+		location.column = static_cast<XmlParsingData*>(prevData)->_cursor.column;
```
DanRStevens: Could probably modify this further to:
```cpp
location = static_cast<XmlParsingData*>(prevData)->_cursor;
```

include/NAS2D/Xml/XmlBase.h
```diff
@@ -94,7 +94,14 @@ class XmlBase
 		XML_ERROR_STRING_COUNT
 	};
 
+	struct ParseLocation
+	{
+		int row{0};
+		int column{0};
```
DanRStevens: In case you were going to submit this to upstream, which I would encourage, this might be using a C+17 feature here. I believe the original author wants to be more general about what standards the code will compile for.

src/Xml/XmlParser.cpp
```diff
@@ -604,20 +603,16 @@ const char* XmlDocument::parse(const char* p, void* prevData)
...
 	// so that parsing starts from the pointer we are given.
 	if (prevData)
 	{
-		location.first = static_cast<XmlParsingData*>(prevData)->_cursor.first;
-		location.second = static_cast<XmlParsingData*>(prevData)->_cursor.second;
-	}
-	else
-	{
-		location.first = 0;
-		location.second = 0;
```
DanRStevens: If the default initial values are removed (C++17), this would still need initialization code here.

include/NAS2D/Xml/XmlBase.h
cugone: TinyXML1 is no longer in development and TinyXML2 explicitly does not use the STL.

src/Xml/XmlParser.cpp
DanRStevens: It's fine. The comment was more about potentially contributing changes upstream.

cugone: Ah.

PR #463, opened 2020-02-13
Fix Image bounds check bug

DanRStevens:
For 0-based indexes, a value at the `width` or `height` is out of bounds.

### DIFF
--- a/src/Resources/Image.cpp
+++ b/src/Resources/Image.cpp
@@ -265,7 +265,7 @@ int Image::center_y() const
  */
 Color Image::pixelColor(int x, int y) const
 {
-	if (x < 0 || x > width() || y < 0 || y > height())
+	if (x < 0 || x >= width() || y < 0 || y >= height())
 	{
 		return Color(0, 0, 0, 255);
 	}


src/Resources/Image.cpp
```diff
@@ -265,7 +265,7 @@ int Image::center_y() const
  */
 Color Image::pixelColor(int x, int y) const
 {
-	if (x < 0 || x > width() || y < 0 || y > height())
+	if (x < 0 || x >= width() || y < 0 || y >= height())
 	{
 		return Color(0, 0, 0, 255);
```
cugone: This should assert or throw instead. Silently returning a color makes me think it's valid.

DanRStevens: That's a perfectly valid argument. Would need to assess impact on calling code. The calling code may be designed to expect silent failures to a transparent color.

I recommend deferring to a separate PR.

cugone: The color *isn't* transparent though. It's solid and visible black.

DanRStevens: Ok, I may have misread the color. Still the same argument though.

Changing the error behaviour is an orthogonal change.

cugone: I make the assumption the user knows what they're doing and just try to access the index regardless. `std::array` will handle an out-of-bounds access for me by hard crashing...

```
Rgba Image::GetTexel(const IntVector2& texelPos) const noexcept {
    int index = texelPos.x + texelPos.y * m_dimensions.x;
    int byteOffset = index * m_bytesPerTexel;
    Rgba color;
    color.r = m_texelBytes[byteOffset + 0];
    color.g = m_texelBytes[byteOffset + 1];
    color.b = m_texelBytes[byteOffset + 2];
    if(m_bytesPerTexel == 4) {
        color.a = m_texelBytes[byteOffset + 3];
    } else {
        color.a = 255;
    }
    return color;
}
void Image::SetTexel(const IntVector2& texelPos, const Rgba& color) noexcept {
    Rgba oldColor = GetTexel(texelPos);
    int index = texelPos.x + texelPos.y * m_dimensions.x;
    int byteOffset = index * m_bytesPerTexel;
    m_texelBytes[byteOffset + 0] = color.r;
    m_texelBytes[byteOffset + 1] = color.g;
    m_texelBytes[byteOffset + 2] = color.b;
    if(m_bytesPerTexel == 4) {
        m_texelBytes[byteOffset + 3] = color.a;
    }
}
```

cugone: But, that said. Your objection is noted for a future change.

PR #464, opened 2020-02-13
Add Image Point Vector methods

DanRStevens:
Add `Image` methods taking `Point` and `Vector` types.

### DIFF
--- a/include/NAS2D/Resources/Image.h
+++ b/include/NAS2D/Resources/Image.h
@@ -11,6 +11,7 @@
 
 #include "Resource.h"
 #include "../Renderer/Color.h"
+#include "../Renderer/Point.h"
 #include "../Renderer/Vector.h"
 
 #include <vector>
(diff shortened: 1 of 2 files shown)


include/NAS2D/Resources/Image.h
```diff
@@ -51,13 +52,15 @@ class Image : public Resource
...
+	Vector<int> center() const;
 
 	int width() const;
 	int height() const;
 
 	int center_x() const;
 	int center_y() const;
 
+	Color pixelColor(Point<int> point) const;
 	Color pixelColor(int x, int y) const;
```
cugone: The split apart function overload is redundant.

DanRStevens: The split apart function still contains the actual implementation. The new overload just defers to the existing method. Would need more refactoring before the old method can be removed. Would also need to update all calls to it to use the new overload instead.

I recommend deferring until more refactoring can be done.

PR #467, opened 2020-02-13
minimum_size function does not recalculate the viewport

cugone:
Found this using the `graphics-test` project.

Calling minimum_size will not recalculate the drawable area (the viewport) or move the window to the center relative to the new size.

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -579,6 +579,7 @@ void RendererOpenGL::size(int w, int h)
 void RendererOpenGL::minimum_size(int w, int h)
 {
 	SDL_SetWindowMinimumSize(underlyingWindow, w, h);
+	_resize(w, h);
 }
 
 


src/Renderer/RendererOpenGL.cpp
```diff
@@ -579,6 +579,8 @@ void RendererOpenGL::size(int w, int h)
 void RendererOpenGL::minimum_size(int w, int h)
 {
 	SDL_SetWindowMinimumSize(underlyingWindow, w, h);
+	_resize(w, h);
+	SDL_SetWindowPosition(underlyingWindow, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
 }
```
DanRStevens: Could you explain this one a bit more? My understanding is that `minimum_size` doesn't actually change the window size or position, but rather sets a lock point when the user tries to resize it too small.

cugone: Easier to show an example:

main.cpp
```
NAS2D::Game game("NAS2D Graphics Test", "NAS2D_GraphicsTest", "LairWorks", argv[0]);
game.go(new TestGraphics());
```

TestGraphics.cpp
```
//...

void TestGraphics::initialize()
{
    //...
    NAS2D::Utility<NAS2D::Renderer>::get().showSystemPointer(true);
    //This gets called *after* the Renderer has already initialized with some arbitrary value.
    NAS2D::Utility<NAS2D::Renderer>::get().minimum_size(1600, 900);
}
```

By using `minimum_size` to ensure a minimum window size, the drawable area does not fill the screen and instead is whatever it was set to previously. So drawing things at `(0,0)` will be on the left-center of the screen instead of the top-left.

Additionally, the window will not be re-centered so it will spill off the desktop.

DanRStevens: But `minimum_size` is not supposed to change the window size. It's supposed to set a limit for when the user resizes the window by dragging the borders.

Are you saying it's a problem if the minimum size is increased beyond the current size?

cugone: Well it is.

---

Yes.

Ref: https://wiki.libsdl.org/SDL_SetWindowMinimumSize
>Use this function to set the minimum size of a window's client area. 

This by definition means it changes the drawable area/viewport. The window should change size to match.

cugone: This is what it looks like on startup:
offscreen

And after I've manually moved the window, you can see that the viewport is not the correct size. I'm drawing everything starting at [0,0].
badViewport

cugone: ...And here's what it looks like with the fix applied. No user action is required and things draw in the correct place:
fixed

DanRStevens: I think we have a fundamental misunderstanding about what the method being edited is supposed to do.

DanRStevens: Ok, to summarize an outside discussion, the `_resize` method does not actually resize the window. Rather it recalculates the viewport dimensions based on the (possibly updated) window size.

The `SDL_SetWindowMinimumSize` method can resize the window when not running fullscreen, if either of the new minimum dimensions are larger than the current dimensions. In that case, the viewport dimensions need to be updated.

If the window is resized such that it spills off the screen, it should ideally be repositioned so it fits within the screen. (This might not be possible if the new dimensions are larger than the screen). If the window does not spill off the screen, it should not be moved.

DanRStevens: Ok, looking back at this now, the call to `_resize` is fine. (It's just horribly named, and actually recalculates viewport dimensions to match window size after a resize event).

The unconditional use of `SDL_SetWindowPosition` still doesn't look right to me. It does makes sense to move a window if it overflows the screen. If it does move, it should be moved just enough to not clip, not necessarily to recenter it. If the window doesn't overflow the screen, it shouldn't be moved at all.

I would like to see this updated so it doesn't recenter the window on each resize. Alternatively, the window movement update can be deferred to a different PR, leaving this one to focus on just recalculating the viewport dimensions.

PR #470, opened 2020-02-14
Simplify signal declarations

DanRStevens:
Simplify `Signal` declarations by removing explicit `void` for no parameter signals, and removing the explicit count of the number of signal parameters.


cugone:
Small nit, it might have been easier to remove the aliases first and let the compiler tell you where the uses were located. I can't tell if you might have missed any.

DanRStevens:
That's basically what I did, but then committed stuff in reverse order so nothing would break along the way. Plus I wanted to stash the breaking changes to make sure the simplifications still worked without them.

PR #476, opened 2020-02-14
Adds getClosestMatchingDisplayMode

### LINKED ISSUE #471: Add getClosestMatchingDisplayMode
Wraps the `SDL_GetClosestDisplayMode` as below. It removes the need to declare the implementation-specific `extern SDL_Window* underlyingWindow;` in game code.
```
NAS2D::DisplayDesc getClosestMatchingDisplayMode(const DisplayDesc& preferred)
{
    const auto display_index = SDL_GetWindowDisplayIndex(underlyingWindow);
    SDL_DisplayMode desired{preferred.width, preferred.height}; //Filter by resolution only. Default other members to desktop.
    SDL_DisplayMode closest{};
    if(SDL_GetClosestDisplayMode(display_index, &desired, &closest))
    {
        return NAS2D::DisplayDesc{closest.w, closest.h, closest.refreshRate};
    }
    const auto err_str = std::string{__FUNCTION__} + " no closest matching display mode available."
    throw std::runtime_error(err_str);
}
```

cugone:
Closes #471

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -777,7 +777,9 @@ NAS2D::DisplayDesc NAS2D::RendererOpenGL::getClosestMatchingDisplayMode(const Di
 	{
 		return {closest.w, closest.h, closest.refresh_rate};
 	}
-	return {};
+	const auto display_str = std::to_string(preferredDisplayDesc.width) + 'x' + std::to_string(preferredDisplayDesc.height) + 'x' + std::to_string(preferredDisplayDesc.refreshHz);
+	auto err_str = "No matching display mode for " + display_str;
+	throw std::runtime_error(err_str);
 }
 // ==================================================================================
 // = NON PUBLIC IMPLEMENTATION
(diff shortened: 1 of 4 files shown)


src/Renderer/RendererOpenGL.cpp
```diff
@@ -764,6 +764,21 @@ std::vector<NAS2D::DisplayDesc> NAS2D::RendererOpenGL::getDisplayModes() const
...
+	preferred.w = preferredDisplayDesc.width;
+	preferred.h = preferredDisplayDesc.height;
+	preferred.refresh_rate = preferredDisplayDesc.refreshHz;
+
+	SDL_DisplayMode closest{};
+	if (SDL_GetClosestDisplayMode(display_index, &preferred, &closest))
+	{
+		return {closest.w, closest.h, closest.refresh_rate};
+	}
+	return {};
```
DanRStevens: Might be better to `throw` rather than silently return an unusable result.

cugone: I was following what the underlying implementations do: SDL just returns a `nullptr` and DirectX returns an `HRESULT` denoting an error code. Neither throw.

Could document that if one isn't found a null description (all zeros) is returned.

DanRStevens: Neither of those can really throw. Exception handling with `throw` is C++ specific. Those libraries are fairly language agnostic.

It's up to the wrapper code to translate library error signalling to language specific error signalling.

cugone: Makes sense.

PR #477, opened 2020-02-14
Adds getWindowDisplayDesc

### LINKED ISSUE #474: Add GetWindowDisplayDesc
This is separate from calling `size` on Renderer as that is for rendering relative to the window size.

This would also include the refresh rate of the display as well as the dimensions in order to interrogate the current window settings:

```
//...
{
	extern SDL_Window* underlyingWindow;
	const auto display_index = SDL_GetWindowDisplayIndex(underlyingWindow);
	auto& cf = NAS2D::Utility<NAS2D::Configuration>::get();
	const auto gfx_width = cf.graphicsWidth();
	const auto gfx_height = cf.graphicsHeight();
        int resolution_index = 0;
        SDL_DisplayMode cur_mode{};
        SDL_GetWindowDisplayMode(underlyingWindow, &cur_mode);
        NAS2D::DisplayDesc current_dimensions{cur_mode.w, cur_mode.h, cur_mode.refresh_rate};
        for(const auto& resolution : resolutions)
	{
		SDL_DisplayMode closest_mode{};
		SDL_DisplayMode desired_mode{};
		desired_mode.w = resolution.width;
		desired_mode.h = resolution.height;
		if(SDL_GetClosestDisplayMode(display_index, &desired_mode, &closest_mode))
		{
                    NAS2D::DisplayDesc closest{closest_mode.w, closest_mode.h, closest_mode.refresh_rate};
                    if(current_dimensions == closest)
                    {
                        //Set combobox to current dimensions of window, not desktop
                        currentResolutionSelection = resolution_index;
                    }
                }
                std::string resolution_str = std::to_string(resolution.width) + "x" + std::to_string(resolution.height) + "x" + std::to_string(resolution.refreshHz);
                cmbResolution.addItem(resolution_str, resolution_index);
                ++resolution_index;
    }
    cmbResolution.currentSelection(currentResolutionSelection);
}
cmbResolution.selectionChanged().connect(this, &MainMenuOptions::onVideoOptionsChanged);
```

cugone:
Closes #474

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -764,6 +764,13 @@ std::vector<NAS2D::DisplayDesc> NAS2D::RendererOpenGL::getDisplayModes() const
 	return result;
 }
 
+NAS2D::DisplayDesc NAS2D::RendererOpenGL::getWindowDisplayDesc() const
+{
+	SDL_DisplayMode cur_mode{};
+	SDL_GetWindowDisplayMode(underlyingWindow, &cur_mode);
+	return {cur_mode.w, cur_mode.h, cur_mode.refresh_rate};
+}
+
 // ==================================================================================
 // = NON PUBLIC IMPLEMENTATION
 // ==================================================================================
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -42,6 +42,9 @@ public:
 	
 	virtual std::vector<DisplayDesc> getDisplayModes() const = 0;
 
+
+	virtual DisplayDesc getWindowDisplayDesc() const = 0;
+
 	const std::string& driverName();
 
 	const std::string& title();
--- a/include/NAS2D/Renderer/RendererNull.h
+++ b/include/NAS2D/Renderer/RendererNull.h
@@ -22,6 +22,9 @@ public:
 
 	std::vector<DisplayDesc> getDisplayModes() const override { return {}; }
 
+
+	DisplayDesc getWindowDisplayDesc() const override { return{}; };
+
 	void drawImage(Image&, float, float, float, uint8_t, uint8_t, uint8_t, uint8_t) override {}
 
 	void drawSubImage(Image&, float, float, float, float, float, float, uint8_t, uint8_t, uint8_t, uint8_t) override {}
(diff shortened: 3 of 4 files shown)


src/Renderer/RendererOpenGL.cpp
```diff
@@ -764,6 +764,13 @@ std::vector<NAS2D::DisplayDesc> NAS2D::RendererOpenGL::getDisplayModes() const
 	return result;
 }
 
+NAS2D::DisplayDesc NAS2D::RendererOpenGL::getWindowDisplayDesc() const
```
DanRStevens: This could maybe use a bit of documentation. A screen has display mode settings. A window itself doesn't. It's maybe a little confusing to call this `getWindowDisplayDesc`. To confuse matters further, a window could be split across two different screens, and those screens might have different resolution settings. A bit more clarity on what this represents would be appreciated. Are these properties of the window, or of the display the window is on, and if split across multiple displays, which one?

cugone: Properties of the *majority* of the window the screen is on...at least that's how it works in DirectX. SDL documentation is lacking when the window is across several monitors.

Will add documentation.

DanRStevens:
Merge conflict needs to be resolved here. Additionally a bit of documentation is needed.

cugone:
The function is being used incorrectly in `optionsMenuBug`. The documentation of the function as provided in `SDL 2.0.10` (The wiki is out of date, easier to get directly from source) states:

```
/**
*  \brief Fill in information about the display mode used when a fullscreen
*         window is visible.
*
*  \sa SDL_SetWindowDisplayMode()
*   \sa SDL_SetWindowFullscreen()
*/
```

I should be using `SDL_GetWindowSize` (which should be named `GetWindowClientArea`!) to interrogate the window for its drawable area.

This function is for getting/setting the description of a display for use with a fullscreen (I'm guessing exclusive) window.

DirectX has no equivalent. You must declare the exclusive fullscreen window description during the swap buffer creation phase. Since no one uses exclusive fullscreen anymore, it's useless to do so.

I'm contemplating closing this in favor of A) fixing `optionsMenuBug` to use the correct call and B) creating a new PR that wraps the correct call in a new function `getClientArea` or even better: encapsulates a `Window` class entirely.

DanRStevens:
The name `getClientArea` sounds much more clear about what it actually does.

I'll hold off until you decide what you want to do with this.

cugone:
This is the wrong function to use. Will remake as `getWindowClientArea`

PR #479, opened 2020-02-14
Adds DisplayDesc inequality comparison operators

### LINKED ISSUE #472: Add DisplayDesc comparison operators
Simplifies comparing descriptions.

cugone:
Closes #472

### DIFF
--- a/src/Renderer/DisplayDesc.cpp
+++ b/src/Renderer/DisplayDesc.cpp
@@ -1,6 +1,6 @@
 #include "NAS2D/Renderer/DisplayDesc.h"
 
-#include <utility>
+#include <tuple>
 
 namespace NAS2D
 {
@@ -18,11 +18,9 @@ namespace NAS2D
 
 	bool operator<(const DisplayDesc& a, const DisplayDesc& b)
 	{
-		const auto adims = std::make_pair(a.width, a.height);
-		const auto ap = std::make_pair(adims, a.refreshHz);
-		const auto bdims = std::make_pair(b.width, b.height);
-		const auto bp = std::make_pair(bdims, b.refreshHz);
-		return ap < bp;
+		const auto a_tuple = std::make_tuple(a.width, a.height, a.refreshHz);
+		const auto b_tuple = std::make_tuple(b.width, b.height, b.refreshHz);
+		return a_tuple < b_tuple;
 	}
 
 	bool operator>(const DisplayDesc& a, const DisplayDesc& b)
--- a/src/Renderer/DisplayDesc.cpp
+++ b/src/Renderer/DisplayDesc.cpp
@@ -1,5 +1,7 @@
 #include "NAS2D/Renderer/DisplayDesc.h"
 
+#include <utility>
+
 namespace NAS2D
 {
 
@@ -11,6 +13,31 @@ namespace NAS2D
 	bool operator!=(const DisplayDesc& a, const DisplayDesc& b)
 	{
 		return !(a == b);
+	}
+
+
+	bool operator<(const DisplayDesc& a, const DisplayDesc& b)
+	{
+		const auto adims = std::make_pair(a.width, a.height);
+		const auto ap = std::make_pair(adims, a.refreshHz);
+		const auto bdims = std::make_pair(b.width, b.height);
+		const auto bp = std::make_pair(bdims, b.refreshHz);
+		return ap < bp;
+	}
+
+	bool operator>(const DisplayDesc& a, const DisplayDesc& b)
+	{
+		return b < a;
+	}
+
+	bool operator<=(const DisplayDesc& a, const DisplayDesc& b)
+	{
+		return !(a > b);
+	}
+
+	bool operator>=(const DisplayDesc& a, const DisplayDesc& b)
+	{
+		return !(a < b);
 	}
 
 } // namespace NAS2D
--- a/include/NAS2D/Renderer/DisplayDesc.h
+++ b/include/NAS2D/Renderer/DisplayDesc.h
@@ -13,4 +13,9 @@ namespace NAS2D
 	bool operator==(const DisplayDesc& a, const DisplayDesc& b);
 	bool operator!=(const DisplayDesc& a, const DisplayDesc& b);
 
+	bool operator<(const DisplayDesc& a, const DisplayDesc& b);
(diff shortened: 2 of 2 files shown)


src/Renderer/DisplayDesc.cpp
```diff
@@ -13,4 +15,29 @@ namespace NAS2D
...
 	}
 
+
+	bool operator<(const DisplayDesc& a, const DisplayDesc& b)
+	{
+		const auto adims = std::make_pair(a.width, a.height);
+		const auto ap = std::make_pair(adims, a.refreshHz);
+		const auto bdims = std::make_pair(b.width, b.height);
+		const auto bp = std::make_pair(bdims, b.refreshHz);
+		return ap < bp;
```
DanRStevens: Might be easier to use [`std::make_tuple`](https://en.cppreference.com/w/cpp/utility/tuple/make_tuple) for something like this. Tuples support [comparison operators](https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp) as well.

cugone:
> 
> 
> Do we want a [strong ordering](https://en.cppreference.com/w/cpp/utility/compare/strong_ordering) here, or would a [partial ordering](https://en.cppreference.com/w/cpp/utility/compare/partial_ordering) make more sense?
> 
> I was thinking:
> 
>* `a <= b` implies all component values compare `<=`
> 
>* `a < b` implies `a <= b` and `a != b` (at least one component `<`, others `<=`)
> 
>* elements are incomparable if `(a.x < b.x) && (a.y > b.y)` for some fields `x` and `y`

You make a good point, however `std::strong_ordering` and `std::partial_ordering` are C++20 features that are unavailable and will have to be manually implemented for now. :(

The underlying implementations differ on ordering. Neither define what happens if the width is the same as the height on Resolution ordering.

Ref: https://wiki.libsdl.org/SDL_GetClosestDisplayMode
>The modes are scanned with size being first priority, format being second priority, and finally checking the refresh rate. If all the available modes are too small, then NULL is returned. 

As usual, SDL's documentation is a single sentence while DirectX is an entire page.

Ref: https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgioutput1-findclosestmatchingmode1
>FindClosestMatchingMode1 resolves similarly ranked members of display modes (that is, all specified, or all unspecified, and so on) in the following order:

>- ScanlineOrdering
>- Scaling
>- Format
>- Resolution
>- RefreshRate

>When FindClosestMatchingMode1 determines the closest value for a particular member, it uses previously matched members to filter the display mode list choices, and ignores other members. For example, when FindClosestMatchingMode1 matches Resolution, it already filtered the display mode list by a certain ScanlineOrdering, Scaling, and Format, while it ignores RefreshRate. This ordering doesn't define the absolute ordering for every usage scenario of FindClosestMatchingMode1, because the application can choose some values initially, which effectively changes the order of resolving members.

>FindClosestMatchingMode1 matches members of the display mode one at a time, generally in a specified order.

>If a member is unspecified, FindClosestMatchingMode1 gravitates toward the values for the desktop related to this output. If this output is not part of the desktop, FindClosestMatchingMode1 uses the default desktop output to find values. If an application uses a fully unspecified display mode, FindClosestMatchingMode1 typically returns a display mode that matches the desktop settings for this output.
>Because unspecified members are lower priority than specified members, FindClosestMatchingMode1 resolves unspecified members later than specified members.

DanRStevens:
The you for the documentation links/quotes.

Since the method is a new virtual method, you could document its behaviour as implementation defined. Leave it to the SDL/DirectX implementation how to order the matching.

Or, you could develop your own matching algorithm, so behaviour is consistent across implementations, in which case the ordering chosen here could dictate how matching works.

----

In terms of strong order or partial ordering, those concepts have been in C++ for many many years, and references to that could be found in various STL algorithms from early on. They weren't yet formalized they way they have been for C++20 though.

At any rate, I don't believe a partial ordering would be any more work that what you've already implemented. It's done by defining each of the individual comparison operators.

cugone:
After reading the descriptions of strong ordering vs partial ordering more closely. I think strong ordering applies best.

If the width and height are equal between two resolutions, the tie-breaker is the refresh-rate in most cases. Given that, partial ordering is not a good fit because it claims that any two description comparisons can return false for all non-compound comparisons. This can't happen if we are only checking for width, height, and refresh rate. By leaving out the formats and directx's scanline ordering and scaling the comparisons are between values that all implementations have.

Other than deciding on ordering, was there any specific changes to be made?

DanRStevens:
I think that's where the partial ordering actually shines.

If resolutions match, and only the refresh rates are different, then the two are comparable, with the ordering set by refresh rate.

```cpp
DisplayDesc{x, y, 50} < DisplayDesc{x, y, 60}
```

----

What would make two modes incomparable is if one had a larger width, while the other had a larger height. There is no obvious ordering in such a case.

```cpp
DisplayDesc{1280, 720, 60} ? DisplayDesc{1024, 768, 60}
```

Another possible way a display mode might be incomparable, is if one has lower resolution, but higher refresh rate.

```cpp
DisplayDesc{1280, 720, 60} ? DisplayDesc{1920, 1024, 50}
```


----

My only real concern here is if this is the right code. No concerns about the specifics of how it was written.

cugone:
> What would make two modes incomparable is if one had a larger width, while the other had a larger height. There is no obvious ordering in such a case.
> 
> ```c++
> DisplayDesc{1280, 720, 60} ? DisplayDesc{1024, 768, 60}
> ```
> 
> Another possible way a display mode might be incomparable, is if one has lower resolution, but higher refresh rate.
> 
> ```c++
> DisplayDesc{1280, 720, 60} ? DisplayDesc{1920, 1024, 50}
> ```

In the implementations of "Give me all available Display Description" functions that I've seen, they are ordered by width, then height, then refresh rate. In the case of `1280x720x60 vs 1024x768x60` The `1280x720x60 ` would be listed first and in the case of `1280x720x60` vs `1920x1024x50` the `1920x1024x50` would be listed first.

DanRStevens:
Should we close this, or did we want to extract the code for some kind of lexicographical comparison function?

cugone:
Yeah, go ahead and close it. If I need a separate type of comparison I can always use a lambda.

PR #481, opened 2020-02-17
Adds Dictionary class

cugone:
This PR is a refactor of the XMLValidation system to support changes requested in https://github.com/lairworks/nas2d-core/issues/434#issuecomment-586780930

See [this portion of Walter Brown's talk](https://www.youtube.com/watch?v=NIDEjY5ywqU&t=39m35s) on How Templates Work for rationale in using regular function overloads rather than function template specialization for template customization.


cugone:
These are just the getters. I have to interrogate the underlying `unordered_map` using `value` to get the value out. Since everything is stored as a string I'm constantly getting the value from the map and converting from a string to the requested type. I could easily use `std::to_string` for setting values in the string, but those aren't implemented yet.

In regards to `std::istringstream` vs `std::stoi` there's a conversation [here on SO](https://stackoverflow.com/questions/33441350/efficiency-of-using-stringstream-to-convert-string-to-int) regarding `std::stringstream` vs the `std::stoi` family of functions performance. The `std::stoi` family of functions are faster.

Just to satisfy my curiosity, I tested it:

```
#include <chrono>
#include <iostream>
#include <sstream>
#include <string>

int main()
{
    {
        //Fill stream
        std::stringstream ss;
        for(int i = 0; i < 1000000; ++i)
        {
            ss << i;
        }
        int d = 0;
        auto start = std::chrono::steady_clock::now();
        for(int i = 0; i < 1000000; ++i)
        {
            ss >> d;
        }
        auto end = std::chrono::steady_clock::now();
        //Don't optimize me away!
        std::cout << "D: " << d << " iss Time: " << std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count() << " ms." << std::endl;
    }
    {
        int d = 0;
        std::string s{"100"};
        auto start = std::chrono::steady_clock::now();
        for(int i = 0; i < 1000000; ++i)
        {
            d = std::stoi(s);
        }
        auto end = std::chrono::steady_clock::now();
        //Don't optimize me away!
        std::cout << "D: " << d << " stoi Time: " << std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count() << " ms." << std::endl;
    }

    return 0;
}
```

For 1'000'000 iterations in Release:
`std::stringstream`: ~180 ms
`std::stoi`: 39 ms

The code repetition is used for performance to initialize `const` values. (Except for `bool`, I need to test if the value is a word or a number, so some modification of the returned value is required).

See [here](https://www.youtube.com/watch?v=9i2ZirXdrVA&t=7m30s) about `IILE` and specifically using in-line lambdas vs storing the lambda off.

src/DataUtils.cpp
```diff
@@ -0,0 +1,778 @@
...
+		}
+	}(); // IILE
+	valueAsStr = NAS2D::trimWhitespace(NAS2D::toLowercase(valueAsStr));
+	if (valueAsStr == "true" || valueAsStr == "false")
+	{
+		return valueAsStr == "true";
+	}
+	else
+	{
+		const auto valueAsBool = [&valueAsStr]() {
```
DanRStevens: I suppose similar advice can be given concerning the bool format about being strict in terms of what you accept. Perhaps we only want to allow either `"true"` or `"false"` for boolean values, and not `0`, or `1`.

DanRStevens:
To reduce duplication, the get/set methods only need to be written once, with the format conversion part abstracted out. Only the format conversion part needs to be written to handle each type, not the rest of the get/set.

With some cleverness, even the format conversion code could potentially be handled in bulk sets, rather than specifically for every type. Maybe one method that handles all integer conversions using `stol` and `numeric_limits<T>::max()`. Another one for all floating point. A dummy null operation for `std::string`.

----

In regards to `std::istringstream`, that suggestion was about the syntax used there, which could allow this to work quite simply. Yes, you could actually implement with `std::istringstream`, it would be quick to do, and the performance really wouldn't matter considering we're loading ballpark 20 configuration settings.

For the more performance conscious, you can of course implement without `std::istringstram`, either by using a similar syntax with out parameters and an internal implementation based on the `std::stoi` family of functions, or by using an alternate format mentioned above that makes use of template specialization (but which doesn't mix it with function overloading) and gives custom return types with no out parameter.

----

(Btw, the performance test you did, doesn't quite do the same for both tests. The `std::stringstream` code uses way more memory, which will impact performance. But then, so would excessive object construction within the timing section. At any rate, this is getting off topic and a bit irrelevant).

----

The use of Immediately Invoked Lambda Expressions is mainly used for complex object initialization. That might be initializing a global variable outside of a function, or wrapping the initialization code in a new scope so you can use variables for intermediate values that don't pollute the parent namespace. The use in this PR doesn't seem to make use of those properties.

DanRStevens:
To add to the above, the `Dictionary` class could be a header only class implemented something like this:
```cpp
#pragma once

#include <string>
#include <unordered_map>

namespace NAS2D
{

	template <typename T>
	void stringTo(const std::string& value);


	class Dictionary
	{
	public:

		template <typename T = std::string>
		void set(const std::string& key, const T& value)
		{
			mDictionary[key] = std::to_string(value);
		}

		template <typename T = std::string>
		T get(const std::string& key) const
		{
			return stringTo<T>(mDictionary.at(key));
		}

		template <typename T = std::string>
		T get(const std::string& key, const T& defaultValue) const
		{
			if (mDictionary.find(key) == std::end(mDictionary)) {
				return defaultValue;
			}
			return stringTo<T>(mDictionary.at(key));
		}

	private:
		std::unordered_map<std::string, std::string> mDictionary;
	};


}
```

The details of the template `stringTo` method have been omitted. It will be needed for the `get` methods. The `set` method makes use of `std::to_string`.

DanRStevens:
As the string to specific type is a fairly fundamental operation, it can be extracted to it's own header. It would be quite easy to implement using template specialization, such as:
```
#pragma once

#include <string>

namespace NAS2D
{
	template <typename T>
	T stringTo(const std::string& value);


	template <>
	std::string stringTo(const std::string& value) {
		return value;
	}

	template <>
	int stringTo(const std::string& value) {
		return stoi(value);
	}

	template <>
	long stringTo(const std::string& value) {
		return stol(value);
	}
}
```

If you insist on using overloads, you could replace the template specialization with a set of overloaded methods that have an out parameter instead of a return value. However, that forces the caller to declare a variable first, and pass it in, and makes the use of `auto` impossible.

Using template specialization:
```
auto value = stringTo<int>("500");
```

Using function overloading and an out parameter:
```
int value;
stringTo("500", &value);
```

I would recommend template specialization as it provides more expressive syntax for library users.

src/DataUtils.cpp
```diff
@@ -0,0 +1,778 @@
...
+	auto valueAsStr = [this, &key]() -> std::string {
+		try
+		{
+			return value<std::string>(key);
+		}
+		catch (...)
+		{
+			throw;
+		}
+	}(); // IILE
```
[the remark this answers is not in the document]
cugone: Yeah, that re-throw was an oversight. I was going to convert it to a `std::runtime_error` like the bool call does. If we don't need the re-throw I can make the change like you said to make the code cleaner.

DanRStevens:
... and if you don't want to write out all the specializations, you can quickly get up and running by using a single `std::istringstream` implementation. :wink:

```cpp
	template <typename T>
	T stringTo(const std::string& value) {
		std::istringstream ss(value);
		T typedValue;
		ss >> typedValue;
		return typedValue;
	}
```

Assuming the type is both default constructible, and supports `operator>>` with the basic stream types.

DanRStevens:
Here's some quick test code, which works with the suggested code above:
```cpp
#include "NAS2D/DataUtils.h"
#include <gtest/gtest.h>


TEST(DataUtils, setGet) {
	NAS2D::Dictionary dict;

	// Set some test values
	dict.set("Key2", 500);
	dict.set("Key3", true);

	// Read back typed values
	EXPECT_EQ(500, dict.get<int>("Key2"));
	EXPECT_EQ(true, dict.get<bool>("Key3"));

	// Set values override defaults
	EXPECT_EQ(500, dict.get<int>("Key2", 1000));
	EXPECT_EQ(true, dict.get<bool>("Key3", false));

	// Default values returned when key does not exist
	EXPECT_EQ(1000, dict.get<int>("Key4", 1000));
	EXPECT_EQ(false, dict.get<bool>("Key4", false));
}
```

----

Notable the code I suggested does not work with the `std::string` type, since `std::to_string` does not have an overload for the null operation of converting a `std::string` to a `std::string`. That case would need to be worked around. Perhaps by providing the overload, or detecting this case and avoiding the call to `std::to_string`.

src/DataUtils.cpp
cugone: I prefer being robust with boolean values. Especially since it's really only two checks, "If the value is convertible to a string, is the lowercased value equal to "true"? If not, is it not zero (i.e., "true")?

Trimming white space is me being paranoid because some strings returned from functions include surrounding white space and would return false for an otherwise true value. TinyXml2 will happily do this if you ask for the Element Text or Value:

Compare this:
```
<foo>true</foo>
```
which returns `"true"`.

While this:
```
<foo>
true
</foo>
```
returns
`"\ntrue\n"`

Annoyances like above have made me use attributes exclusively when intending for strict text values.

DanRStevens: True... (pun?), though keep in mind these files will be machine written. The format should always be consistent. That's true of both spacing and casing.

----

I had to go looking to dig up the terms for what we're discussing. It's known as the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). You're proposing we follow the Robustness principle. I'm bringing up the objections from the Criticism section, which suggests the principle may actually harm robustness.

Another source on the topic is:
[The Harmful Consequences of the Robustness Principle](https://tools.ietf.org/id/draft-thomson-postel-was-wrong-03.html)

----

Oh, and we are exclusively using attributes for the values.

include/NAS2D/StringUtils.h
```diff
@@ -17,6 +17,44 @@
 
 namespace NAS2D
 {
+	template<typename T>
+	T from_string(const std::string& value)
+	{
+		return T(value);
+	}
```
DanRStevens: This method could probably use an `std::enable_if` guard, to make sure it is only considered when `T` is constructible from a `std::string`. I'd have to think about the exact syntax needed to do that. Maybe a tomorrow thing.

include/NAS2D/StringUtils.h
```diff
@@ -17,6 +17,44 @@
...
 namespace NAS2D
 {
+	template<typename T>
+	T from_string(const std::string& value)
+	{
+		return T(value);
+	}
+
+	template<>
+	std::string from_string(const std::string& value);
```
DanRStevens: Curious why this wasn't implemented inline. Usually template methods have the body in the header file. I guess with a specialization like this, it is possible to implement in a source file. That would prevent inlining though, and these are pretty small and easily inlined methods.

src/StringUtils.cpp
```diff
@@ -14,6 +14,74 @@
...
+template<>
+int NAS2D::from_string(const std::string& value)
+{
+	return std::stoi(value);
+}
+
+template<>
+unsigned int NAS2D::from_string(const std::string& value)
+{
+	return static_cast<unsigned int>(std::stoul(value));
```
DanRStevens: This may do silent wrapping of values that are out of range.

cugone: The STL doesn't have an overload for `std::stoui`, unfortunately. This was the only way to safely  coerce the value into an unsigned integer. Using `std::stoi` drops half the values in the range of `unsigned int`. Getting the value as an `unsigned long` at least covers all of `unsigned int` and the value can be downcast.

`Dictionary::set(..., unsigned int)` will force the input to only be within the acceptable range of `unsigned int`. Converting an `unsigned int` to `unsigned long` and back is safe to do.

include/NAS2D/StringUtils.h
cugone: I don't have much (re: any) experience with using `std::enable_if`, definitely a tomorrow thing.

cugone: At least it's documented that construction from `std::string` is required.

src/StringUtils.cpp
DanRStevens: Yes, though we have no guarantee the `set` and `get` calls will be properly paired on type.

It might be wise to check the results with `numeric_limits<unsigned int>::max()`, and `throw` if it is out of range. Once the check is complete, the `static_cast` should be safe to do.

include/NAS2D/StringUtils.h
cugone: Putting explicit template specializations in the header causes VS to throw an "already defined" error:
```
1>E:\Git\nas2d-core\src\StringUtils.cpp(22,1): error C2766: explicit specialization; 'std::string NAS2D::from_string<std::string>(const std::string &)' has already been defined
1>E:\Git\nas2d-core\include\NAS2D\StringUtils.h(27): message : see previous definition of 'from_string'
```

cugone: Defining explicit template specializations in the header causes the VS linker to freak out:

```
2>Filesystem.test.obj : error LNK2005: "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl NAS2D::from_string<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (??$from_string@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@NAS2D@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABV12@@Z) already defined in DataUtils.test.obj
2>StringUtils.test.obj : error LNK2005: "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl NAS2D::from_string<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (??$from_string@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@NAS2D@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABV12@@Z) already defined in DataUtils.test.obj
2>NAS2D.lib(StringUtils.obj) : warning LNK4075: ignoring '/EDITANDCONTINUE' due to '/SAFESEH' specification
2>NAS2D.lib(StringUtils.obj) : error LNK2005: "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl NAS2D::from_string<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (??$from_string@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@NAS2D@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABV12@@Z) already defined in DataUtils.test.obj
2>NAS2D.lib(Image.obj) : error LNK2005: "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl NAS2D::from_string<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (??$from_string@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@NAS2D@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABV12@@Z) already defined in DataUtils.test.obj
2>NAS2D.lib(Filesystem.obj) : error LNK2005: "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl NAS2D::from_string<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (??$from_string@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@NAS2D@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABV12@@Z) already defined in DataUtils.test.obj
2>E:\Git\nas2d-core\test\x86\Debug\test.exe : fatal error LNK1169: one or more multiply defined symbols found

```

ldicker83:
So XML is going to be pulled, I fail to see the point in refactoring any of it as a markup language is up to the user to decide on and to decide which library/implementation to use.

cugone:
Having Key-Value parsing is useful in other aspects.

The Dictionary class should be kept around for that at least.

DanRStevens:
Reference: #739

PR #484, opened 2020-02-18
Refactors _resize

cugone:
Refactors `_resize`'s implementation by splitting it into three component functions that are also useful individually:

- `setViewport`
- `setOrthoProjection`
- `setResolution`

 and renames it to a more meaningful function name `onResize`.

Two of the component functions are implementation-defined between `DirectX` and `OpenGL` so their declarations were declared as pure virtual in the base `Renderer` class.

`setResolution` just sets the private member variable on `Renderer` so it made more sense to just declare it there.

### DIFF
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -136,6 +136,8 @@ public:
 
 	virtual void setViewport(const Rectangle<int>& viewport) = 0;
 	virtual void setOrthoProjection(const Rectangle<float>& orthoBounds) = 0;
+	void setResolution(const Point<float> newResolution);
+
 protected:
 	Renderer(const std::string& appTitle);
 
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -134,6 +134,8 @@ public:
 
 	virtual void update();
 
+	virtual void setViewport(const Rectangle<int>& viewport) = 0;
+
 protected:
 	Renderer(const std::string& appTitle);
 
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -136,7 +136,7 @@ public:
 
 	virtual void setViewport(const Rectangle<int>& viewport) = 0;
 	virtual void setOrthoProjection(const Rectangle<float>& orthoBounds) = 0;
-	void setResolution(const Point<float> newResolution);
+	void setResolution(const Vector<float>& newResolution);
 
 protected:
 	Renderer(const std::string& appTitle);
--- a/include/NAS2D/Renderer/Renderer.h
+++ b/include/NAS2D/Renderer/Renderer.h
@@ -135,7 +135,7 @@ public:
 	virtual void update();
 
 	virtual void setViewport(const Rectangle<int>& viewport) = 0;
-
+	virtual void setOrthoProjection(const Rectangle<float>& orthoBounds) = 0;
 protected:
 	Renderer(const std::string& appTitle);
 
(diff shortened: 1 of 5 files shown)


include/NAS2D/Renderer/RendererOpenGL.h
```diff
@@ -11,6 +11,9 @@
 
 #include "Renderer.h"
 
+#include "Point.h"
+#include "Vector.h"
```
DanRStevens: No problem with this, just wanted to point out these are included by the base class. That is necessarily true, since the base class has method signatures that use them. I'm uncertain if it's more proper to include these or leave them out for derived classes. Might be a point worth discussing.

include/NAS2D/Renderer/Renderer.h
```diff
@@ -136,6 +136,8 @@ class Renderer
 
 	virtual void setViewport(const Rectangle<int>& viewport) = 0;
 	virtual void setOrthoProjection(const Rectangle<float>& orthoBounds) = 0;
+	void setResolution(const Point<float> newResolution);
```
DanRStevens: Resolution is a size quantity. It may make more sense as a `Vector`.

The use of `float` also stands out, though that is pre-existing practice, so probably best to leave as is.

src/Renderer/RendererOpenGL.cpp
```diff
@@ -625,16 +625,12 @@ bool RendererOpenGL::resizeable()
...
-	glViewport(0, 0, w, h);
-	glMatrixMode(GL_PROJECTION);
-	glLoadIdentity();
-	glOrtho(0.0, (GLdouble)w, (GLdouble)h, 0.0, -1.0, 1.0);
-	glMatrixMode(GL_MODELVIEW);
+	const auto wAsf = static_cast<float>(w);
+	const auto hAsf = static_cast<float>(h);
+	setViewport(Rectangle{0, 0, w, h});
+	setOrthoProjection(Rectangle<float>::Create(Point{0.0f, 0.0f}, Vector<float>{wAsf, hAsf}));
+	setResolution(Point{wAsf, hAsf});
```
DanRStevens: If `setResolution` is changed to accept a `Vector`, it can share the same parameter as `setOrthoProjection`.

src/Renderer/RendererOpenGL.cpp
```diff
@@ -625,16 +625,12 @@ bool RendererOpenGL::resizeable()
 
 void RendererOpenGL::_resize(int w, int h)
 {
-	glViewport(0, 0, w, h);
-	glMatrixMode(GL_PROJECTION);
-	glLoadIdentity();
-	glOrtho(0.0, (GLdouble)w, (GLdouble)h, 0.0, -1.0, 1.0);
-	glMatrixMode(GL_MODELVIEW);
+	const auto wAsf = static_cast<float>(w);
+	const auto hAsf = static_cast<float>(h);
```
DanRStevens: This could potentially be simplified as:
```cpp
const auto size = Vector<int>{w, h}.to<float>();
```

cugone:
@DanRStevens 

The AppVeyor/pr CI build isn't pulling latest or is not properly updating its cache.

The error it claims does not exist. `_resize` is not being called. at `RendererOpenGL.cpp(528,2)`

cugone:
@DanRStevens 

Correction. Line 582

src/Renderer/Renderer.cpp
```diff
@@ -534,10 +534,10 @@ void Renderer::update()
 	}
 }
 
-void NAS2D::Renderer::setResolution(const Point<float> newResolution)
+void NAS2D::Renderer::setResolution(const Vector<float>& newResolution)
 {
 	if (!fullscreen())
 	{
-		mResolution = newResolution;
+        mResolution = {newResolution.x, newResolution.y};
```
DanRStevens: Indentation

PR #491, opened 2020-02-24
Extra uses of std::endl in Game are unnecessary.

cugone:
Calling flush multiple times in a row for a stream that hasn't changed just because a newline is needed is incorrect. Prefer `\n` where you can and call `std::cout::flush()` manually.

`std::endl` should only be used if you need the stream to be flushed ASAP and certainly *never* multiple times in a row.


cugone:
The relevant guideline: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rio-endl

PR #499, opened 2020-03-10
DisplayDesc partial order comparison

DanRStevens:
Add partial order comparison for `DisplayDesc`.

The partial order comparison encapsulates how people might view comparisons between different screen modes. This can be used for filtering out screen modes that don't meet some minimum set of requirements.


DanRStevens:
Forgot to add references:
Issue #472
PR #479

PR #502, opened 2020-03-11
Use Vector for size() in Renderer

DanRStevens:
A `size()` is more accurately represented by a `Vector` than by a `Point`.


DanRStevens:
Huh, looks like I requested review after you already reviewed it.

Well, thanks for the quick review.

PR #504, opened 2020-03-12
Fix zero-as-null-pointer-constant warnings

DanRStevens:
Fix zero-as-null-pointer-constant warnings, and enable warning flag:
`-Wzero-as-null-pointer-constant`

### DIFF
--- a/include/NAS2D/Delegate.h
+++ b/include/NAS2D/Delegate.h
@@ -108,7 +108,7 @@ struct SimplifyMemFunc
 	inline static GenericClass* Convert(X* /*pthis*/, XFuncType /*function_to_bind*/, GenericMemFuncType& /*bound_func*/)
 	{
 		static_assert(N < 100, "Unsupported member function pointer on this compiler");
-		return 0;
+		return nullptr;
 	}
 };
 
@@ -236,11 +236,11 @@ protected:
 
 public:
 #if !defined(FASTDELEGATE_USESTATICFUNCTIONHACK)
-	DelegateMemento() : m_pthis(0), m_pFunction(0), m_pStaticFunction(0) {};
-	void clear() { m_pthis = 0; m_pFunction = 0; m_pStaticFunction = 0; }
+	DelegateMemento() : m_pthis(nullptr), m_pFunction(nullptr), m_pStaticFunction(nullptr) {};
+	void clear() { m_pthis = nullptr; m_pFunction = nullptr; m_pStaticFunction = nullptr; }
 #else
-	DelegateMemento() : m_pthis(0), m_pFunction(0) {};
-	void clear() { m_pthis = 0; m_pFunction = 0; }
+	DelegateMemento() : m_pthis(nullptr), m_pFunction(nullptr) {};
+	void clear() { m_pthis = nullptr; m_pFunction = nullptr; }
 #endif
 public:
 #if !defined(FASTDELEGATE_USESTATICFUNCTIONHACK)
@@ -248,7 +248,7 @@ public:
 	{
 		if (m_pFunction != x.m_pFunction) return false;
 		if (m_pStaticFunction != x.m_pStaticFunction) return false;
-		if (m_pStaticFunction != 0) return m_pthis == x.m_pthis;
+		if (m_pStaticFunction != nullptr) return m_pthis == x.m_pthis;
 		else return true;
 	}
 #else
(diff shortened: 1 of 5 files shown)


include/NAS2D/Delegate.h
```diff
@@ -236,19 +236,19 @@ class DelegateMemento
...
+	void clear() { m_pthis = nullptr; m_pFunction = nullptr; }
 #endif
 public:
 #if !defined(FASTDELEGATE_USESTATICFUNCTIONHACK)
 	inline bool IsEqual(const DelegateMemento& x) const
 	{
 		if (m_pFunction != x.m_pFunction) return false;
 		if (m_pStaticFunction != x.m_pStaticFunction) return false;
-		if (m_pStaticFunction != 0) return m_pthis == x.m_pthis;
+		if (m_pStaticFunction != nullptr) return m_pthis == x.m_pthis;
```
cugone: It's a common C++ idiom to compare pointers as conversions to bool, there's even a rule about it: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions . The second **Note** references using negation instead of `!=`

include/NAS2D/Delegate.h
```diff
@@ -257,15 +257,15 @@ class DelegateMemento
 	inline bool IsLess(const DelegateMemento& right) const
 	{
 		#if !defined(FASTDELEGATE_USESTATICFUNCTIONHACK)
-		if (m_pStaticFunction != 0 || right.m_pStaticFunction != 0) return m_pStaticFunction < right.m_pStaticFunction;
+		if (m_pStaticFunction != nullptr || right.m_pStaticFunction != nullptr) return m_pStaticFunction < right.m_pStaticFunction;
```
cugone: Ditto

include/NAS2D/Delegate.h
```diff
@@ -257,15 +257,15 @@ class DelegateMemento
...
 		#endif
 		if (m_pthis != right.m_pthis) return m_pthis < right.m_pthis;
 
 		return memcmp(&m_pFunction, &right.m_pFunction, sizeof(m_pFunction)) < 0;
 	}
 
-	inline bool operator ! () const { return m_pthis == 0 && m_pFunction == 0; }
-	inline bool empty() const { return m_pthis == 0 && m_pFunction == 0; }
+	inline bool operator ! () const { return m_pthis == nullptr && m_pFunction == nullptr; }
+	inline bool empty() const { return m_pthis == nullptr && m_pFunction == nullptr; }
```
cugone: Ditto

include/NAS2D/Delegate.h
```diff
@@ -334,13 +334,13 @@ class ClosurePtr : public DelegateMemento
 	inline void CopyFrom(DerivedClass* pParent, const DelegateMemento& x)
 	{
 		SetMementoFrom(x);
-		if (m_pStaticFunction != 0) m_pthis = reinterpret_cast<GenericClass*>(pParent);
+		if (m_pStaticFunction != nullptr) m_pthis = reinterpret_cast<GenericClass*>(pParent);
```
cugone: Ditto

include/NAS2D/Delegate.h
```diff
@@ -334,13 +334,13 @@ class ClosurePtr : public DelegateMemento
...
 	}
 
 	template <class DerivedClass, class ParentInvokerSig>
 	inline void bindstaticfunc(DerivedClass* pParent, ParentInvokerSig static_function_invoker, StaticFuncPtr function_to_bind)
 	{
-		if (function_to_bind == 0) m_pFunction = 0;
+		if (function_to_bind == nullptr) m_pFunction = nullptr;
```
cugone: Ditto

include/NAS2D/Delegate.h
```diff
@@ -353,7 +353,7 @@ class ClosurePtr : public DelegateMemento
...
```
cugone: Ditto

include/NAS2D/Delegate.h
```diff
@@ -368,7 +368,7 @@ class ClosurePtr : public DelegateMemento
 
 	inline bool IsEqualToStaticFuncPtr(StaticFuncPtr funcptr)
 	{
-		if (funcptr == 0) return empty();
+		if (funcptr == nullptr) return empty();
```
cugone: Ditto

src/Xml/XmlParser.cpp
```diff
@@ -570,7 +570,7 @@ void XmlDocument::streamIn(std::istream& in, std::string& tag)
 			if (node)
 			{
 				node->streamIn(in, tag);
-				bool isElement = node->toElement() != 0;
+				bool isElement = node->toElement() != nullptr;
```
cugone: Ditto

include/NAS2D/Delegate.h
DanRStevens: I agree we shouldn't be overly verbose with such comparisons.

Maybe slightly out of scope for the change I intended, though it may be suitable to include here.

src/Xml/XmlParser.cpp
DanRStevens: This one seems slightly odd to remove, as that leaves an assignment operator with an implicit type conversion. I suppose it's not much different than the others, though it kind of feels different here.

DanRStevens: Actually, come to think of it, [ES.87](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions) is worded to apply specifically to conditions. This is not a conditional per say, but rather an assignment expression. I'm not sure that ES.87 really applies in spirit here.

The other thought I had, was for custom types, such code would fail if the type conversion operator was made `explicit`. I suppose we're using the built in type conversion operator for pointers though, which is not `explicit`.

I'm kind of on the fence about this one.

cugone: Yeah, I'm going to agree with you on this one. It's okay to leave it.

PR #505, opened 2020-03-12
Fix extra semicolons

DanRStevens:
Found using the flag `-Wextra-semi`.

The warning flag is supported by Clang and GCC-8. It is not supported by GCC-7.5 (which is stock with Ubuntu 18.04), nor by Mingw-w64. Enabling the flag globally would cause CI build failures for the linux (stock GCC) and Mingw builds. As such, it is not being added to the `makefile`.


DanRStevens:
I suppose because I anticipated you would request changes if I touched that code and didn't fix something obvious with it. Though yes, those changes probably should have been separated out into a followup PR.

Issue #510, opened 2020-03-12
Missing declarations warnings


DanRStevens:
When compiled with `-Wmissing-declarations`, GCC-8 produces the following warnings:
```
src/Resources/Sprite.cpp:27:8: warning: no previous declaration for ‘std::__cxx11::string endTag(int, const string&)’ [-Wmissing-declarations]
 string endTag(int row, const std::string& name)
        ^~~~~~

src/Resources/Font.cpp:58:10: warning: no previous declaration for ‘unsigned int nextPowerOf2(unsigned int)’ [-Wmissing-declarations]
 unsigned nextPowerOf2(unsigned n)
          ^~~~~~~~~~~~

src/Mixer/MixerSDL.cpp:34:6: warning: no previous declaration for ‘void MIXER_HOOK()’ [-Wmissing-declarations]
 void MIXER_HOOK() { MIXER_HOOK_CALLBACK_SIGNAL(); }
      ^~~~~~~~~~

src/Xml/XmlBase.cpp:20:6: warning: no previous declaration for ‘void fillErrorTable()’ [-Wmissing-declarations]
 void fillErrorTable()
      ^~~~~~~~~~~~~~
```

I've found wrapping a method in an anonymous namespace can silence the warning. Perhaps these should be investigated to see if they are file level functions that should be protected from outside access.

DanRStevens:
The `-Wmissing-declarations` flag, and a number of other flags were found from various articles suggesting warning flags to compile with.

References:
[GCC: 3.8 Options to Request or Suppress Warnings](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)
[StackOverflow: How to turn on (literally) ALL of GCC's warnings?](https://stackoverflow.com/questions/11714827/how-to-turn-on-literally-all-of-gccs-warnings)
[Best GCC warning flags for compiling C++](https://www.artificialworlds.net/blog/2014/07/18/best-gcc-warning-flags-for-compiling-c/)

cugone:
Oddly, this is compiler specific. MSVC doesn't, and has never, warned for this. The only warning I get is `Function definition for 'function' not be found.` from live code analysis, not intellisense or the compiler, for missing *definitions* not declarations. The code compiles just fine, no warnings, even with Warning Level All and Warnings As Errors. If I'm missing a declaration at class scope it's declared as an error: `<Foo> Is not a member of <Bar>`, but file-level functions work just fine.

[...]

PR #514, opened 2020-03-13
Xml declare data where used

DanRStevens:
Fixes one of the warnings from #510.

It seems the XML code was using a string data table from the parser, but for some odd reason the table was created in the `XmlBase` file, where it wasn't used.

Doing early initialization of the string table using the constructor seems more straightforward. One small downside is the table will always be initialized, whether it is used or not.

With C++20, it may be possible to do the initialization at compile time. Of course, by then, we may remove the XML code from direct inclusion in this library, so none of this will matter.

### DIFF
--- a/src/Xml/XmlBase.cpp
+++ b/src/Xml/XmlBase.cpp
@@ -15,24 +15,6 @@ using namespace NAS2D::Xml;
 
 bool XmlBase::condenseWhiteSpace = true;
 
-const std::vector<std::string> XML_ERROR_TABLE = {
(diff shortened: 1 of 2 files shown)


cugone:
This was probably in XmlBase so a user could reference the strings for error checking/reporting. I dunno...

DanRStevens:
Yeah, I don't know. The error strings appear to relate to parse errors. Maybe just an oversight when things were split. I think the original code was all in one file.

PR #515, opened 2020-03-13
Wrap private file level method in anonymous namespace

DanRStevens:
Fix one of the warnings from #510.

The `endTag` method is a private file level helper function (non-class member). It is not meant to be shared with other translation units. As such, it can be wrapped in an anonymous namespace to protect it from being accessed from other translation units.

From the [GCC Warnings](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) page under `-Wmissing-declarations`:

> Warn if a global function is defined without a previous declaration. Do so even if the definition itself provides a prototype. Use this option to detect global functions that are not declared in header files. In C, no warnings are issued for functions with previous non-prototype declarations; use -Wmissing-prototypes to detect missing prototypes. In C++, no warnings are issued for function templates, or for inline functions, or for functions in anonymous namespaces.

### DIFF
--- a/src/Resources/Sprite.cpp
+++ b/src/Resources/Sprite.cpp
@@ -23,10 +23,12 @@ const string NAS2D::SPRITE_VERSION("0.99");
 const int FRAME_PAUSE = -1;
 
 
-// Adds a row/name tag to the end of messages.
-string endTag(int row, const std::string& name)
-{
-	return " (Row: " + std::to_string(row) + ", " + name + ")";
+namespace {
+	// Adds a row/name tag to the end of messages.
+	string endTag(int row, const std::string& name)
+	{
+		return " (Row: " + std::to_string(row) + ", " + name + ")";
+	}
 }
 
 


DanRStevens:
That doesn't seem to be fully in the spirit of the warning flag. The flag was meant to find methods that have no prototype in a header file.

The help text for the flag even suggests using an anonymous namespace.

In terms of protection, you have less protection by declaring a prototype in the source file. There's nothing stopping anyone from declaring an equivalent `extern` prototype in their own source file, and thereby gain access to such a method. Using an anonymous namespace makes that impossible. They would have to modify the original file to gain access to the method.

cugone:
> In terms of protection, you have less protection by declaring a prototype in the source file. There's nothing stopping anyone from declaring an equivalent `extern` prototype in their own source file, and thereby gain access to such a method. Using an anonymous namespace makes that impossible. They would have to modify the original file to gain access to the method.

If that's the case, why isn't this being done for every non-public, implementation-only function?

DanRStevens:
Well, in terms of the projects here at Outpost Universe, because we only recently realized the benefits of wrapping functions in anonymous namespaces.

I sort of felt this realization only happened within the last few months, thought searching GitHub, it seems [we started making use of anonymous namespaces](https://github.com/search?q=org%3AOutpostUniverse+anonymous+namespace&type=Issues) back about June 2018.

Granted, most of the work done since then has been refactoring old code, rather than developing or extending new API code. That may explain why we haven't been using the technique much yet. Though in the case of the op2ext, we did expand library capabilities using that technique in the last year. The op2ext entries from the above search should provide the most detail about the technique. In particular: https://github.com/OutpostUniverse/OP2Utility/pull/177#issuecomment-419167719

I should point out that I've since become more used to wrapping functions in an anonymous namespace, and would actually recommend that as a good design, rather than dumping a bunch of private methods in the header file, where they affect a much greater compilation area. The distinction of how to implement the methods is largely dependent on whether or not they need direct access to object state.

For a more recent example, op2ext used the technique back in November to implement some file level variables:
https://github.com/OutpostUniverse/op2ext/blob/bb6b6a224561afd79dccafd7a187f44809e5db45/srcStatic/Log.cpp

And of course, as a workaround for the random number generator problem:
https://github.com/OutpostUniverse/OPHD/pull/167

In that case we do actually want to consolidate uses into a single instance, rather than keep instances separate. However, the compile problem you encountered there shows what happens when you have a name collision in two separate compilation units for a global variable. Once translation units are linked together, re-use of the same name in the same project can be a problem. Wrapping methods in anonymous namespaces helps protect against that.

So yeah, the idea looked odd to me at first. Though once I started using it and understood it better, I've come to find it may actually be a good practice. I've since noticed anonymous namespaces being used in a number of libraries where I felt the quality of implementation was pretty good. I'm kind of getting the feeling wrapping private top level helper functions into anonymous namespaces may be a best practice.

cugone:
Well, I really meant "Why isn't everybody doing it?" or "why isn't this taught in academia?", but nice write-up none-the-less. 😛 

Turns out it's compiler specific, disabled by default, and is mostly a C concern, not a C++ one. (See my comments on #510)

DanRStevens:
My guess is it's not more wide spread because many people aren't aware of the issue, and they aren't aware of the issue because there are so many other potentially more important issues that they need to learn first.

----

The C-concern listed in that issue is rather different than the current warning being addressed. The issue addressed here is very much a C++ concern. It relates to the default linkage of file level globals, and potential name collisions at link time. As the matter is concerned with visibility of symbols (at link time), there is no particular difference between functions and variables, they are all treated the same.

Not including a prototype (or extern variable declaration) in a header impacts compile time name visibility issue. This is separate from the link time name visibility. The linkage of a symbol is concerned with the link time name visibility.

DanRStevens:
The following article contains a nice summary of linkage:
[Internal and External Linkage in C++](http://www.goldsborough.me/c/c++/linker/2016/03/30/19-34-25-internal_and_external_linkage_in_c++/)

DanRStevens:
For a slightly silly example to illustrate linkage issues:

Moral.h
```cpp
float doGood(float extent); // extent: 0.0 (minimal) to 1.0 (very good)
float sin(float extent); // extent: 0.0 (minimal) to 1.0 (way bad)
```

Math.h
```cpp
float sin(float radians);
float cos(float radians);
```

Main.cpp
```cpp
// Let's simplify things by not adding header confusion
// Forward declaration
float sin(float);

int main() {
  sin(1.0); // Which function is called?

  return 0;
}
```

:wink:

----

The above example is a little different from what we're dealing with, since both all functions are probably assumed to be `public`, and are being called from a third file. In that case you have a real name collision problem. If instead, one of those functions was a `private` support function, which just happened to conflict with a public method name, that's the case we're really dealing with. This may be different yet again from the case of two `private` support functions, and no sharing from a third file.

cugone:
Given the silly example and my immediate thought was: "Well, duh, that's why namespaces exist!" I now get what you're talking about. I'd just never used [unnamed namespaces](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces) in such a way before. I'd used them in order to [hide a global variable](https://github.com/cugone/Abrams2019/blob/7f70c5bf2dfe0d09faf099fd6912d8daf9ec8d69/Engine/Code/Engine/Math/MathUtils.cpp#L25) that I didn't want users to be able to access; but never thought about using them for functions. (Every example I've seen only uses them for variables!)

PR #516, opened 2020-03-15
Add angleToDirection method

DanRStevens:
Needed to adjust a method's return type from `Point` to `Vector`:
- Add `angleToDirection` method (returns a `Vector`).
- Deprecate `getDirectionVector` method (oddly returns `Point`, as it predates the introduction of `Vector`).

I chose to deprecate rather than remove the old method so old code wouldn't immediately break. The old method is used in nas2d-tests.

Edit: I also chose to rename the method so it would describe the concept rather than the type being used. Type names should normally be kept out of identifier names (except maybe for a generic type conversion method).

### DIFF
--- a/include/NAS2D/Trig.h
+++ b/include/NAS2D/Trig.h
@@ -24,8 +24,6 @@ float degToRad(float degree);
 float radToDeg(float rad);
 float angleFromPoints(float x, float y, float x2, float y2);
 Vector<float> angleToDirection(float angle);
-[[deprecated("Please use angleToDirection instead")]]
-Point_2df getDirectionVector(float angle);
 
 bool lineIntersectsCircle(const Point_2d& p, const Point_2d& q, const Point_2d& c, float r);
 
--- a/include/NAS2D/Trig.h
+++ b/include/NAS2D/Trig.h
@@ -24,6 +24,7 @@ float degToRad(float degree);
 float radToDeg(float rad);
 float angleFromPoints(float x, float y, float x2, float y2);
 Vector<float> angleToDirection(float angle);
+[[deprecated("Please use angleToDirection instead")]]
 Point_2df getDirectionVector(float angle);
 
 bool lineIntersectsCircle(const Point_2d& p, const Point_2d& q, const Point_2d& c, float r);
--- a/include/NAS2D/Trig.h
+++ b/include/NAS2D/Trig.h
@@ -23,6 +23,7 @@ extern const float RAD2DEG;
 float degToRad(float degree);
 float radToDeg(float rad);
 float angleFromPoints(float x, float y, float x2, float y2);
+Vector<float> angleToDirection(float angle);
 Point_2df getDirectionVector(float angle);
 
 bool lineIntersectsCircle(const Point_2d& p, const Point_2d& q, const Point_2d& c, float r);
--- a/src/Trig.cpp
+++ b/src/Trig.cpp
@@ -54,6 +54,17 @@ float NAS2D::angleFromPoints(float x, float y, float x2, float y2)
 /**
  * Gets a directional vector from an angle in degrees.
  */
+Vector<float> angleToDirection(float angle)
+{
+	// static_cast<float> used to suppress warning at possible loss of data. Intentionally
+	// surpressed as we don't need that level of precision.
+	return {static_cast<float>(std::sin(NAS2D::degToRad(angle))), static_cast<float>(-std::cos(NAS2D::degToRad(angle)))};
+}
+
+
+/**
+ * Gets a directional vector from an angle in degrees.
+ */
 Point_2df NAS2D::getDirectionVector(float angle)
 {
 	// static_cast<float> used to suppress warning at possible loss of data. Intentionally
--- a/src/Trig.cpp
+++ b/src/Trig.cpp
@@ -54,7 +54,7 @@ float NAS2D::angleFromPoints(float x, float y, float x2, float y2)
 /**
  * Gets a directional vector from an angle in degrees.
  */
-Vector<float> angleToDirection(float angle)
+Vector<float> NAS2D::angleToDirection(float angle)
 {
 	// static_cast<float> used to suppress warning at possible loss of data. Intentionally
 	// surpressed as we don't need that level of precision.
--- a/src/Trig.cpp
+++ b/src/Trig.cpp
@@ -59,15 +59,6 @@ Vector<float> NAS2D::angleToDirection(float angle)
 
 
 /**
- * Gets a directional vector from an angle in degrees.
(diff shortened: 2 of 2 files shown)


include/NAS2D/Trig.h
```diff
@@ -23,6 +23,8 @@ extern const float RAD2DEG;
 float degToRad(float degree);
 float radToDeg(float rad);
 float angleFromPoints(float x, float y, float x2, float y2);
+Vector<float> angleToDirection(float angle);
+[[deprecated("Please use angleToDirection instead")]]
```
cugone: As much as I'd like to approve this, I can't. Microsoft has decided that using the `deprecated` attribute should always result in an error, not a warning as the specification says it should.

Consider changing it to `#pragma message` to get around the problem.

src/Trig.cpp
```diff
@@ -51,6 +51,17 @@ float NAS2D::angleFromPoints(float x, float y, float x2, float y2)
...
 
 
+/**
+ * Gets a directional vector from an angle in degrees.
+ */
+Vector<float> NAS2D::angleToDirection(float angle)
+{
+	// static_cast<float> used to suppress warning at possible loss of data. Intentionally
+	// surpressed as we don't need that level of precision.
+	return {static_cast<float>(std::sin(NAS2D::degToRad(angle))), static_cast<float>(-std::cos(NAS2D::degToRad(angle)))};
```
cugone: This is incorrect. The formula is `x = std::cos(degToRad(angle)); y = std::sin(degToRad(angle))`

The cast to float is not needed either, there are overloads that return a `float`.

include/NAS2D/Trig.h
DanRStevens: I think that depends on the warning level and having warnings as errors enabled.

I suppose I could just remove the old function. As far as I'm aware, it's only used one place in the test code, so should be easy enough to update the call while updating the NAS2D submodule. It was probably overly conservative of me to keep the old method and deprecate it.

src/Trig.cpp
DanRStevens: Oh yeah, the `float` overloads were added in C++11. For some reason I thought that was an MSVC thing. I did actually try removing the casts locally and saw there was no conversion warning.

Reference: [`std::sin`](https://en.cppreference.com/w/cpp/numeric/math/sin)

----

I'll have to give the rotation formula some consideration. It will depend on things like orientation of the coordinate system, and orientation of the angle. Perhaps we should open an Issue and change that as a separate PR.

DanRStevens: Hmm, seems there is similar `static_cast` for other trig functions too, such as [`std::atan2`](https://en.cppreference.com/w/cpp/numeric/math/atan2).

DanRStevens: I've opened issue #519 to track the formula used to convert from an `angle` to a `Vector`.

cugone: The above referenced Rotation Matrix only applies if you are rotating a point in space. `angleToDirection` isn't doing that. It's calculating a unit vector given an angle:
https://en.wikipedia.org/wiki/Trigonometric_functions#Unit-circle_definitions

>The trigonometric functions cos and sin are defined, respectively, as the x- and y-coordinate values of point A, i.e., cos [theta] = Ax and sin [theta] = Ay

cugone: Additionally, this function should really be moved into `Vector` itself as it has no equivalent for three- and four-dimensional vectors.

DanRStevens: The exact formula used is perhaps out of scope for this PR, and so better addressed as a separate issue. This change is more concerned with the return type.

----

Adding the method to `Vector` is an interesting idea. Though in this case it's constructing a `Vector` from a scalar value. That would imply a factory function, rather than a member function.

DanRStevens: I was thinking, the reverse direction, which calculates the angle from a `Vector` would make for a good member function.

Probably best to name it something like `angle`, rather than the more mathematical name `arg` (argument). Though there may need to be a distinction for radians versus degrees. Or we could just default to always using radians.

----

If I add an `angle` method, I would probably add the reverse factory method at the same time.

----

I would probably want to use strict mathematical conventions for something implemented directly in `Vector`.

For other conventions, such as screen coordinates, we'd probably want to keep those helper methods outside of `Vector`.

cugone: https://github.com/cugone/Abrams2019/blob/master/Engine/Code/Engine/Math/Vector2.hpp

- `CalcHeadingRadians/Degrees` for the getting the angle of a vector (`atan2(y/x)`)
- `SetHeadingRadians/Degrees` for setting the angle via `R * cos and sin` without changing the length.
- `SetUnitLengthAndHeadingRadians/Degrees` for setting x/y directly with `cos(angle)`/`sin(angle)`

include/NAS2D/Trig.h
cugone: > I think that depends on the warning level and having warnings as errors enabled.

Nope. Regardless of warning level and if WAE is on or off, still emits an error. Moot now, since you just removed the old function.

DanRStevens:
I'm going to close this one for some rework.

src/Trig.cpp
DanRStevens: Just had a thought about moving this stuff over to `Vector`. The result `angle` will always be a floating point value, even if `BaseType` is `int`. Similarly, you might have a `Vector<double>`. Should that return a `float` angle, or a `double` angle?

Should it be possible to construct a `Vector<int>` from an angle?

I'm kind of leaning away from including this directly in `Vector` for the moment. It seems these methods are more useful for `Vector<float>`, rather than generic `Vector` types.

cugone: > 
> 
> Just had a thought about moving this stuff over to `Vector`. The result `angle` will always be a floating point value, even if `BaseType` is `int`. Similarly, you might have a `Vector<double>`. Should that return a `float` angle, or a `double` angle?
> 
> Should it be possible to construct a `Vector<int>` from an angle?
> 
> I'm kind of leaning away from including this directly in `Vector` for the moment. It seems these methods are more useful for `Vector<float>`, rather than generic `Vector` types.

This is where trying to be generic all the time is going to hurt and make life more complicated for no reason. It'd be much simpler to just have `Vector2/3/4` and `IntVector2/3/4`. `double` and `long double`s are abhorred by both GPUs and the gaming industry. They just don't use them due to space, alignment, and padding requirements.

DanRStevens: I suppose if you wanted to get specific with types, you could create a `VectorFloat` class that derived from `Vector<float>` and implemented extension methods.

I'm not sure it's really worth it though.

I'm tempted to leave the conversion as free standing methods.

Issue #523, opened 2020-03-17
Sprite::skip versus Sprite::setFrame


DanRStevens:
The current `Sprite::skip` implementation behaves more like a `Sprite::setFrame` method. We should consider providing both methods.

Currently we can't just change the implementation of `skip`, since downstream projects such as OPHD use `skip` and depend on the current behaviour. I recommend adding a new `setFrame` method, using the current implementation of `skip`. Downstream projects can switch to using `setFrame`. We can then update the `skip` method implementation to use `+=` instead of `=`. That will make `skip` match the current documentation, and behaviour implied by its name.

DanRStevens:
I was looking back at this. I think maybe just renaming the existing method is the way to go.

We're not actually using any kind of `skip` functionality, so let's not bother with it for now. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle. Having both methods would probably just be picked up as code duplication in Codacy anyway.

PR #524, opened 2020-03-17
Refactor Sprite::skip

DanRStevens:
Refactor `Sprite::skip`.

----

I noticed #523 while working on this. I wanted to keep refactoring separate from implementation change, hence the separate issue.

----

There's probably more that can be done with this method, such as avoiding the double lookup (both `find` and `operator[]`). The result is surprisingly ugly though, using currently available standard library methods. The `find` method returns an iterator, which dereferences to a `std::pair`. The result is just too ugly. I thought I would defer on changing that part for now.

More long term, I wonder if we should use a `std::vector` rather than a `std::map`, and move the name lookup outside of this code. Rather than specify actions by named string, we should perhaps be specifying them by numerical index. The string to index lookup can be done ahead of time during file loading, caching the results. Might need a separate issue to track this idea.

### DIFF
--- a/include/NAS2D/Resources/Sprite.h
+++ b/include/NAS2D/Resources/Sprite.h
@@ -49,7 +49,7 @@ public:
 	void pause();
 	void resume();
 
-	void skip(int frames);
+	void skip(int frameCount);
 
 	void update(Point<float> position);
 	void update(float x, float y);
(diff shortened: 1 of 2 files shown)


DanRStevens:
Huh, that's a good idea. I didn't think to use structured bindings.

Issue #528, opened 2020-03-18
Increase compiler warnings


DanRStevens:
We should strive to increase compiler warning options.

In #510, there were links to some useful external resources for warning flags (GCC):
> References:
[GCC: 3.8 Options to Request or Suppress Warnings](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)
[StackOverflow: How to turn on (literally) ALL of GCC's warnings?](https://stackoverflow.com/questions/11714827/how-to-turn-on-literally-all-of-gccs-warnings)
[Best GCC warning flags for compiling C++](https://www.artificialworlds.net/blog/2014/07/18/best-gcc-warning-flags-for-compiling-c/)

----

As noted in the links, you very likely do not want to enable all possible warning flags, as some may not make sense for modern code, and some may even be contradictory. Nevertheless, it's hard to know what warnings are available, or applicable to your code, without some way to turn them on. Conveniently Clang provides the `-Weverything` flag for such a purpose.

I experimented with Clang by enabling the `-Weverything` option, and then trying to disable warnings until the project was able to build again without warnings or errors. For each warning seen, I disabled it by adding the `no-` prefix to the flag name. The resulting command that got the project compiling cleanly again was:
```
make CXXFLAGS_EXTRA="-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-documentation-unknown-command -Wno-weak-vtables -Wno-comma -Wno-sign-conversion -Wno-double-promotion -Wno-ctad-maybe-unsupported -Wno-cast-qual -Wno-old-style-cast -Wno-padded -Wno-missing-variable-declarations -Wno-float-equal -Wno-exit-time-destructors -Wno-inconsistent-missing-destructor-override -Wno-global-constructors -Wno-date-time -Wno-newline-eof -Wno-unreachable-code-break -Wno-shadow-field"
```

Some of those are not of interest. In particular, we explicitly use C++17 features, and so we don't care about code that breaks compatibility with C++98. Other flags however do seem to suggest there are things we could improve upon. In particular, unknown commands in Doxygen comments, abuse of the comma operator, unexpected conversions, missing override declarations, no newlines at end of file, unreachable code, and old style casts.

----

The last link recommended the following flags:
```
-Wredundant-decls
-Wcast-align
-Wmissing-declarations
-Wmissing-include-dirs
-Wswitch-enum
-Wswitch-default
-Wextra
-Wall
-Werror
[...]

PR #531, opened 2020-03-18
Replace C-style casts

DanRStevens:
Reference: #528 (`-Wold-style-cast`)

Replaces old C-style casts with `static_cast` or `reinterpret_cast`.


DanRStevens:
Agreed. At least using the newer `X_cast` syntax allows you to search for casts, so they can be found and fixed easier.

cugone:
In particular, any non-pointer `reinterpret_cast` is suspect.

PR #532, opened 2020-03-18
Fix missing extern warnings

DanRStevens:
Reference: #528 (`-Wmissing-variable-declarations`)

Wraps file level variables in an anonymous namespace to limit their scope. This prevents warnings when Clang is passed the `-Wmissing-variable-declarations` flag.

More long term, we'll probably want to move some of these into function level locals.


DanRStevens:
Closing to rework.

Issue #493, opened 2020-02-26
Separate updating from rendering


cugone:
In nearly all classes, updating the state of an object is done *in* the draw function (or vice versa). This violates [the rule that functions should do one thing](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f2-a-function-should-perform-a-single-logical-operation).

Highly consider splitting the various `draw/update` functions into a separate `void update(float deltaSeconds)` and `void draw() const` or `void draw(Renderer& r) const` where appropriate.

PR #536, opened 2020-03-18
Renderer const-correctness

cugone:
Precursor to #493 and [OPHD#253](https://github.com/OutpostUniverse/OPHD/issues/253)


cugone:
Deleting to rework. The draw functions on a Renderer shouldn't be `const`, just other classes that need to be drawn.

PR #542, opened 2020-03-21
Adds stringTo and stringFrom conversion functions

cugone:


cugone:
@DanRStevens  You have easier access to Linux in order to figure out what they want in terms of `extern template` functions. It's been a feature since c++11 so I don't understand why Linux, GCC, GCC-8, MinGW, and MacOS all give different errors.

include/NAS2D/StringUtils.h
```diff
@@ -54,5 +97,23 @@ namespace NAS2D
...
+				if (std::numeric_limits<T>::max() < valueAsInteger)
+				{
+					throw std::out_of_range("rangeCheckHelper argument out of range");
+				}
+			}
+			return static_cast<T>(valueAsInteger);
+		}
+		extern template char rangeCheckHelper(const std::string& value);
+	}
+
```
DanRStevens: This is a private detail which is not used in the header file. It would be better off in an unnamed namespace in the implementation file.

src/StringUtils.cpp
```diff
@@ -14,6 +14,259 @@
...
 #include <sstream>
 
+template<>
+char NAS2D::detail::rangeCheckHelper(const std::string& value)
+{
+	if (value.empty())
+	{
+		throw std::invalid_argument("rangeCheckHelper string must be non-empty.");
+	}
+	return value[0];
```
DanRStevens: I'm still on the fence about using quoted string values for this, rather than small range integers.

If we're going to use quoted string values, this check should at least be changed so it checks for exactly one letter strings, rather than not 0 length strings.

cugone: Non-signed and non-unsigned `char`s are understood to be single character values. If you want small range integers, use `signed char` or `unsigned char`.

include/NAS2D/StringUtils.h
cugone: Is that even possible with templates? It was my understanding base templates are [typically defined in the header](https://www.youtube.com/watch?v=NIDEjY5ywqU&t=6m55s).

DanRStevens: Things (including templates) that are used by multiple translation units need to be in a header. This code is only used internally by one translation unit. It does not need to be in a header.

This is not to be confused with why template **definitions** are typically placed in headers, as opposed to only a **declaration**. That video is talking about definitions and usability (from multiple translation units). In that case, yes, the definition needs to be in the header, since templates need a full definition to be usable.

Aside: An exception to that, are templates with a specific set of type arguments that are declared to be `extern`. In that case, the compiler can assume the function will be defined elsewhere for that exact set of type parameters, and can rely on the linker to later find the right references. Note that a template with a full set of type parameters specified is basically equivalent to a regular function.

src/StringUtils.cpp
DanRStevens: I understand where you're coming from. Though I would like to point out that `'a'` and `"a"` are different types in C++. To specify a single `char`, you use the `'a'` syntax. If you use `"a"` you get a `char[]` (which decays to `char*`). Syntactically, the `char` type is compatible with single character literals of the form `'a'` and with integer literals of the form `97`.

Additionally it's fairly common practice to use `char` to hold byte arrays for binary data. It's maybe not the best practice, but it is quite common.

----

Regardless of syntax used for literals, a `char` type can only hold 1 character, so the check should be for a string of `size` 1.

src/StringUtils.cpp
```diff
@@ -14,6 +14,223 @@
...
+std::string NAS2D::stringFrom(double value)
+{
+	return std::to_string(value);
+}
+
+template<>
+std::string NAS2D::stringFrom(long double value)
+{
+	return std::to_string(value);
+}
```
DanRStevens: These all have the exact same implementation, and can be collapsed to a single definition.

src/StringUtils.cpp
```diff
@@ -14,6 +14,223 @@
...
+std::string NAS2D::stringFrom(short value)
+{
+	return std::to_string(static_cast<int>(value));
+}
+
+template<>
+std::string NAS2D::stringFrom(unsigned short value)
+{
+	return std::to_string(static_cast<unsigned int>(value));
+}
```
DanRStevens: I'm not sure the cast adds much value here. Automatic [numeric promotion](https://en.cppreference.com/w/cpp/language/implicit_conversion) should allow an appropriate function overload to be called.

Removing the cast makes the implementation match the others, and can allow further consolidation of the function definitions.

cugone: `std::to_string` doesn't have overloads for types smaller than `int`, hence, the casts.

test/StringUtils.test.cpp
```diff
@@ -1,8 +1,352 @@
...
+	EXPECT_EQ("5"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("5"s)));
+	EXPECT_EQ("6"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("6"s)));
+	EXPECT_EQ("7"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("7"s)));
+	EXPECT_EQ("8"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("8"s)));
+	EXPECT_EQ("9"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("9"s)));
+	EXPECT_EQ("\n"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("\n"s)));
+	EXPECT_EQ("a"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("a"s)));
+	EXPECT_EQ("b"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("b"s)));
+	EXPECT_EQ("t"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("true"s)));
+	EXPECT_EQ("f"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("false"s)));
```
DanRStevens: For string of length 1 requirement, these should fail.

test/StringUtils.test.cpp
```diff
@@ -1,8 +1,352 @@
...
+	EXPECT_EQ("1"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("1"s)));
+	EXPECT_EQ("2"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("2"s)));
+	EXPECT_EQ("3"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("3"s)));
+	EXPECT_EQ("4"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("4"s)));
+	EXPECT_EQ("5"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("5"s)));
+	EXPECT_EQ("6"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("6"s)));
+	EXPECT_EQ("7"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("7"s)));
+	EXPECT_EQ("8"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("8"s)));
+	EXPECT_EQ("9"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("9"s)));
+	EXPECT_EQ("\n"s, NAS2D::stringFrom<char>(NAS2D::stringTo<char>("\n"s)));
```
DanRStevens: This check is also questionable, as `"\n"` may be multiple characters on some platforms.

src/StringUtils.cpp
cugone: Could you elaborate?

src/StringUtils.cpp
DanRStevens: And I'm saying it doesn't need a cast. If you try to pass a `short`, it will be automatically promoted to `int`, and then call the `int` overload.

src/StringUtils.cpp
DanRStevens:
```cpp
template<typename T>
std::string NAS2D::stringFrom(T value)
{
	return std::to_string(value);
}
```

DanRStevens: If you want to get fancy, you can use SFINAE to limit the definition to types for which `std::to_string` is callable with that type. The SFINAE expression might look something like:
```cpp
decltype(std::to_string(declval<T>()))
```

I'm not quite certain on the exact syntax needed, though it might look something like:
```cpp
template<typename T>
decltype(std::to_string(declval<T>()))
NAS2D::stringFrom(T value)
{
	return std::to_string(value);
}
```

That should result in a return type of `std::string`, provided `std::to_string` is callable with the template type.

test/StringUtils.test.cpp
```diff
@@ -1,8 +1,387 @@
...
+	const std::string bigNum(40, '9');
+
+	EXPECT_EQ(""s, NAS2D::stringTo<std::string>(NAS2D::stringFrom<std::string>("")));
+	EXPECT_EQ("Hello World"s, NAS2D::stringTo<std::string>(NAS2D::stringFrom<std::string>("Hello World")));
+
+	EXPECT_EQ(true, NAS2D::stringTo<bool>(NAS2D::stringFrom<bool>(true)));
+	EXPECT_EQ(false, NAS2D::stringTo<bool>(NAS2D::stringFrom<bool>(false)));
+
+	EXPECT_EQ(65, NAS2D::stringTo<char>(NAS2D::stringFrom<char>(65)));
+	EXPECT_EQ(66, NAS2D::stringTo<char>(NAS2D::stringFrom<char>(66)));
```
DanRStevens: This could perhaps be written as:
```cpp
EXPECT_EQ('A', NAS2D::stringTo<char>(NAS2D::stringFrom<char>('A')));
EXPECT_EQ('B', NAS2D::stringTo<char>(NAS2D::stringFrom<char>('B')));
```

test/StringUtils.test.cpp
```diff
@@ -116,60 +129,83 @@ TEST(String, stringFrom)
...
-	EXPECT_EQ("1.100000"s, NAS2D::stringFrom<long double>(1.1L));
+}
+
+
+template<typename T>
+class StringFromSignedIntegerTest : public ::testing::Test
+{
+  public:
+	  StringFromSignedIntegerTest<T>()
+	{}
```
DanRStevens: This probably don't need to be explicit. If it does, it can probably be `= default;`

test/StringUtils.test.cpp
```diff
@@ -116,60 +129,83 @@ TEST(String, stringFrom)
...
+
+	T negOne{-1};
+	T zero{0};
+	T one{1};
```
DanRStevens: These should probably be `protected`. Or, maybe better, just inline them as `T{-1}`, `T{0}`, and `T{1}`. They make more sense as the numerical constants in the context where they are used. It's easier to see how the numerical constants translate to string results.

... and yes, you may end up with an empty class.

test/StringUtils.test.cpp
```diff
@@ -116,60 +129,83 @@ TEST(String, stringFrom)
...
+};
+
+using StringFromSignedIntegerTestTypes = ::testing::Types<short, int, long, long long>;
+TYPED_TEST_SUITE(StringFromSignedIntegerTest, StringFromSignedIntegerTestTypes);
```
DanRStevens: It might be possible to use parenthesis, if you want to avoid the `using` alias:
```cpp
TYPED_TEST_SUITE(StringFromSignedIntegerTest, (::testing::Types<short, int, long, long long>));
```

test/StringUtils.test.cpp
```diff
@@ -116,60 +129,83 @@ TEST(String, stringFrom)
...
+
+TYPED_TEST(StringFromSignedIntegerTest, StringFromSignedIntegerTestEdgeCases)
+{
+	using namespace std::literals::string_literals;
+
+	EXPECT_EQ("-1"s, NAS2D::stringFrom<TypeParam>(this->negOne));
+	EXPECT_EQ("0"s, NAS2D::stringFrom<TypeParam>(this->zero));
+	EXPECT_EQ("1"s, NAS2D::stringFrom<TypeParam>(this->one));
```
DanRStevens: These might be easier to read with inline numeric constants, rather than named constants:
`TypeParam{-1}`, `TypeParam{0}`, `TypeParam{1}`

test/StringUtils.test.cpp
```diff
@@ -116,60 +129,83 @@ TEST(String, stringFrom)
...
+  public:
+	StringFromUnsignedIntegerTest<T>()
+	{}
+
+	T zero{0};
+	T one{1};
+};
+
+using StringFromUnsignedIntegerTestTypes = ::testing::Types<unsigned short, unsigned int, unsigned long, unsigned long long>;
+TYPED_TEST_SUITE(StringFromUnsignedIntegerTest, StringFromUnsignedIntegerTestTypes);
```
DanRStevens: Same as previous class.

----

Actually, if both classes are empty,  maybe there's a way to reuse the same class.

What really matters is the test function needs to be properly paired with the test type list. The test harness (class) doesn't really do anything on it's own. Though I suppose it serves as the name that ties the test function(s) to the list of test types.

test/StringUtils.test.cpp
cugone: Documentation says it's required to make the macro work because of the commas.

test/StringUtils.test.cpp
cugone: Just tried reusing an empty class. `TYPED_TEST_SUITE` requires the test suite name (the first parameter of the macro) to be unique or it throws an "redefinition; different basic types" error.

test/StringUtils.test.cpp
DanRStevens: I realize that. I don't think the documentation is entirely accurate.

Without the parenthesis, it will certain give an error due to the commas.

DanRStevens:
The current failures look familiar. I think it was a bug in Google Test. Can't remember where I saw it before though.

cugone:
Linux is failing to compile because of `-Wgnu-zero-variadic-macro-arguments` temporarily disabling or turning the warning off will get rid of the failure.

test/StringUtils.test.cpp
cugone: It  errors for the following change:

```
//using StringFromSignedIntegerTestTypes = ::testing::Types<short, int, long, long long>;
TYPED_TEST_SUITE(StringFromSignedIntegerTest, (::testing::Types<short, int, long, long long>));
```

Intellisense:
`Function definition 'TYPED_TEST_SUITE' not found. expected a type specifier` (remember, if it can be parsed as a function, it will. Macros are just text replacement, after all.)

Compiler freaks out:
```
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(140,1): error C2059: syntax error: '>'
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(140,1): error C2062: type 'unknown-type' unexpected
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(140,1): error C2039: 'type': is not a member of '`global namespace''
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(142,1): error C2065: 'gtest_type_params_StringFromSignedIntegerTest_': undeclared identifier
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(142,1): error C2923: 'testing::internal::TypeParameterizedTest': 'gtest_type_params_StringFromSignedIntegerTest_' is not a valid template type argument for parameter 'Types'
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(142,1): error C2955: 'testing::internal::TypeParameterizedTest': use of class template requires template argument list
1>E:\Git\vcpkg\installed\x86-windows\include\gtest\internal\gtest-internal.h(689): message : see declaration of 'testing::internal::TypeParameterizedTest'
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(142,1): error C2672: 'testing::internal::GenerateNames': no matching overloaded function found
1>E:\Git\nas2d-core\test\StringUtils.test.cpp(142,1): error C2974: 'testing::internal::GenerateNames': invalid template argument for 'Types', type expected
1>E:\Git\vcpkg\installed\x86-windows\include\gtest\internal\gtest-internal.h(675): message : see declaration of 'testing::internal::GenerateNames'
1>E:\Git\vcpkg\installed\x86-windows\include\gtest\internal\gtest-internal.h(698,1): fatal error C1903: unable to recover from previous error(s); stopping compilation
1>INTERNAL COMPILER ERROR in 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\bin\HostX86\x86\CL.exe'
1>    Please choose the Technical Support command on the Visual C++
1>    Help menu, or open the Technical Support help file for more information
1>Done building project "test.vcxproj" -- FAILED.
```

DanRStevens:
The Google Test project has an open issue for this bug:
https://github.com/google/googletest/issues/2271

Looks like the Docker image we are using for the Clang build has an up-to-date version of Google Test.

cugone:
The advice given in the open issue and the several linked comments and prior issues is: "disable the warning when it provides no value"; "we don't support every warning"; "why are you trying to compile with `-Wextra`?"...Gee, it's almost as if they don't want to fix it.

DanRStevens:
Yeah, that comment was a little disappointing. Especially given how common `-Wall` and `-Wextra` are.

----

I found some stuff relevant to the `extern template` attempt: 
[What's the right way to specialize a template when using “extern template”?](https://stackoverflow.com/questions/5356269/whats-the-right-way-to-specialize-a-template-when-using-extern-template)

It seems specializations must be known up front, before the template is used. That kind of makes sense, since if the compiler doesn't know about the specialization, it would try to use the generic version with appropriate substituted types. I'm a little surprised the linking doesn't work a bit differently for `extern` specializations, though I suppose the compiler is always free to inline functions if it wants to, and knowing about specializations will affect inlining.

It seems for free standing functions we don't need the `extern` keyword. Apparently it's mostly useful when you have class member functions that you want to make `extern`.

src/StringUtils.cpp
```diff
@@ -14,196 +14,380 @@
...
+
+	template<>
+	char rangeCheckHelper(const std::string& value)
+	{
+		if (value.size() != 1)
+		{
+			throw std::invalid_argument("rangeCheckHelper string length must be exactly one (1).");
+		}
+		return value[0];
+	}
```
DanRStevens: Rather than specialize `rangeChangeHelper`, it may make more sense to move this code into the calling function:
```cpp
template<> char stringTo(const std::string& value)
```

test/StringUtils.test.cpp
```diff
@@ -1,8 +1,370 @@
...
+	EXPECT_EQ("b"s, NAS2D::stringFrom<char>('b'));
+	EXPECT_EQ("129"s, NAS2D::stringFrom<unsigned char>(129));
+	EXPECT_EQ("-1"s, NAS2D::stringFrom<signed char>(-1));
+
+}
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+#endif
```
DanRStevens: We can probably reduce the scope of this block. I think it might only need to wrap the use of `TYPED_TEST_SUITE`.

We can potentially combine the code for multiple types into one region. Looks like they can all be declared up front too. Maybe we should set a generic type alias at the head of the file for `SignedTypes` and `UnsignedTypes`, and then use the macro up near the top. The test methods can be below. It looks like that ordering is already used, so it should work.

src/StringUtils.cpp
```diff
@@ -14,196 +14,373 @@
...
-	std::string curString{};
-	while (std::getline(ss, curString, delim))
+	template<>
+	bool stringTo(const std::string& value)
 	{
-		result.push_back(curString);
+		auto v = toLowercase(value);
+		if(v == "true") return true;
+		if(v == "false") return false;
+		throw std::invalid_argument("Value is not 'true' or 'false' value: " + value);
```
DanRStevens: I would prefer not using a highly abbreviated name such as `v`.

----

For that matter, I would prefer not including the `toLowercase` as part of this method. It takes choice away from the caller on how strict they want to be. Instead, let the caller decide if they should call `toLowercase` themselves.

This solves the above problem, since the `value` parameter can then be used throughout. There is no need for a secondary modified variable.

cugone: Normalizing strings is standard practice when comparing them; otherwise you'd have to write:

```cpp
if(value == "true" || value == "TRUE" || value == "True" || value == "tRue" || value == "trUe" || value == "truE") return true;
//...etc
```

In this particular case, taking the argument by value to let the compiler generate a copy is probably better.

DanRStevens: In this case it would be `stringTo<bool>(toLowercase(value))`. That's easy enough to deal with. It also doesn't take away my choice if I want a case sensitive compare, in which case I can write `stringTo<bool>(value)`.

By trying to be too clever and doing too much, we actually limit expressability and composability.

----

If we get rid of the `toLowercase`, the string use becomes entirely read-only. That removes the potential for move semantic optimizations.

Issue #545, opened 2020-03-24
Relax GLSL support requirement in RendererOpenGL


DanRStevens:
Perhaps we should remove this restriction for RendererOpenGL:
```cpp
	if (glsl_v.empty())
	{
		throw renderer_no_glsl();
	}
```

When running Windows in a Virtual Machine, I encountered a case where there was no GLSL support.

Commenting out this line allowed the game to load and display startup graphics. (It failed later due to a missing font file, so I didn't check if game play could display graphics).

cugone:
Where is that code?

cugone:
After a bit of googling, several sites and stack exchange network question/answers state that VMs don't get access to the physical GPU. Instead they get a virtual software GPU. (Unless you have HyperV and GPU passthrough, but that's highly unlikely). Trying to support VMs is a rabbit hole that should be avoided.

DanRStevens:
My point was, we may not have an actual hard requirement here. If we don't require it, we shouldn't abort when it doesn't exist.

When I commented out the check, I was getting graphics output.

----

Tagging #544 as a related issue. This check follows immediately after the code to retrieve strings from OpenGL.

DanRStevens:
Closed by #646.

Issue #544, opened 2020-03-24
Possible crash bug in RendererOpenGL initialization


DanRStevens:
The function [`glGetString`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetString.xhtml) can potentially return 0, which can cause an access violation if it is dereferenced. This can happen during `RendererOpenGL` initialization in the `std::string` constructor:
```cpp
	// Spit out system graphics information.
	std::cout << "\t- OpenGL System Info -" << std::endl;

	driverName((char*)glGetString(GL_RENDERER));

	std::cout << "\tVendor: " << glGetString(GL_VENDOR) << std::endl;
	std::cout << "\tRenderer: " << driverName() << std::endl;
	std::cout << "\tDriver Version: " << glGetString(GL_VERSION) << std::endl;

	std::string glsl_v = (char*)glGetString(GL_SHADING_LANGUAGE_VERSION);
```

When running Windows in a virtual machine, the last line causes a crash.

----

Since we call `glGetString` multiple times, perhaps we should wrap it in a method that can safely return an empty `std::string` when a `nullptr` is returned.

cugone:
[The documentation](http://docs.gl/gl4/glGetString) says it only returns zero if there was an error and that an error only occurs if you give it the wrong `name` value. Since we are already giving it all the possible *name* values, the error can never occur. (`glGetStringi` and `GL_EXTENTIONS` notwithstanding as that's a different function.)

cugone:
After a bit of googling, several sites and stack exchange network question/answers state that VMs don't get access to the physical GPU. Instead they get a virtual software GPU. (Unless you have HyperV and GPU passthrough, but that's highly unlikely). Trying to support VMs is a rabbit hole that should be avoided.

DanRStevens:
Nevertheless, I did get an error, and the program crashed.

We should at minimum provide a proper error message and shutdown gracefully when an error occurs.

PR #546, opened 2020-03-25
Adds std::string-safe wrappers to glGetString

cugone:
Related #544 and #545

### DIFF
--- a/include/NAS2D/Renderer/RendererOpenGL.h
+++ b/include/NAS2D/Renderer/RendererOpenGL.h
@@ -92,9 +92,13 @@ private:
 
 	void initGL();
 	void initVideo(unsigned int resX, unsigned int resY, bool fullscreen, bool vsync);
+	std::string getGlString(unsigned int stringId) const;
+	std::string getGlRendererString() const;
+	std::string getGlVendorString() const;
+	std::string getGlVersionString() const;
+	std::string getGlShadingLanguageVersionString() const;
 
 	void onResize(int w, int h);
-
 
 	Point_2df desktopResolution;
 
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -684,13 +684,13 @@ void RendererOpenGL::initGL()
 	// Spit out system graphics information.
 	std::cout << "\t- OpenGL System Info -" << std::endl;
 
-	driverName(reinterpret_cast<const char*>(glGetString(GL_RENDERER)));
+	driverName(getGlRendererString());
 
-	std::cout << "\tVendor: " << glGetString(GL_VENDOR) << std::endl;
+	std::cout << "\tVendor: " << getGlVendorString() << std::endl;
 	std::cout << "\tRenderer: " << driverName() << std::endl;
-	std::cout << "\tDriver Version: " << glGetString(GL_VERSION) << std::endl;
+	std::cout << "\tDriver Version: " << getGlVersionString() << std::endl;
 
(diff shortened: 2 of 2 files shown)


DanRStevens:
Ahh, I see you beat me to it. Guess I should have checked messages before doing the work on #547.

include/NAS2D/Renderer/RendererOpenGL.h
```diff
@@ -92,10 +92,14 @@ class RendererOpenGL: public Renderer
 
 	void initGL();
 	void initVideo(unsigned int resX, unsigned int resY, bool fullscreen, bool vsync);
+	std::string getGlString(unsigned int stringId) const;
+	std::string getGlRendererString() const;
+	std::string getGlVendorString() const;
+	std::string getGlVersionString() const;
+	std::string getGlShadingLanguageVersionString() const;
```
DanRStevens: I suppose that's one way to get around adding an OpenGL include to the header file.

----

The [`glGetString`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetString.xhtml) function takes a `GLenum` parameter. Here an `unsigned int` is being passed. That could potentially lead to a skew in types between the caller and the API.

cugone: `GLenum` is a typedef for an `unsigned int`, additionally, each string ID is a `#define` no bigger than `#FFFF`.

And yes, I did not want to pollute the header with an include for OpenGL. :)

DanRStevens: The first part of my comment was aimed more at the specialized functions for each enum value. By not taking a parameter, they don't have a type dependence on the enum.

I only realized a bit later that `getGlString` was a general method used by all of them. That sort of spoils the attempt in a way, since it does take an enum value.

I would strongly recommend against changing types just to prevent namespace pollution in header files. Particularly since types and type aliases sometimes change overtime, like during the transition from 32-bit to 64-bit. You might be able to get away with it here, but I recommend that you don't.

----

Another thought is those methods don't need access to class member variables, so they could potentially be made static. Of course, if you have a static method, you can potentially convert it to a regular function in an unnamed namespace in the source file, and completely remove all traces of it from the header file.

A benefit to that approach, is you can use types in the signature that are only included by the implementation file, so that's another approach to avoid namespace pollution in the header file.

That was the approach I took in #547.

cugone: Yeah, I like your approach a little better. It's just icky that you have to use the value provided by OpenGL.

My approach was since there are so few valid enum values in the first place, using a common function to call in to allowed for one point of failure if either an invalid value was passed or more values were added, in that case, just add another function.

Could even go so far as to have a private `ConvertOpenGLStringToEnum` function that did the conversion via `switch` instead of just a direct call.

DanRStevens: You mean the reverse operation? I'm afraid I don't quite follow that last part.

cugone: It's an alternative way to remove the header dependency by only using "our" types.

DanRStevens: Ahh, I assume you mean replace their enum with our own enum, and then translate between them with a `switch`.

cugone: Yes. :)

cugone:
Closing. #547 is a little better. Next time we should assign tasks before starting work so we don't end up doing the same work. :)

PR #547, opened 2020-03-25
Wrap calls to glGetString

### LINKED ISSUE #544: Possible crash bug in RendererOpenGL initialization

DanRStevens:
Closes #544

Wraps call to `glGetString`, allowing for translation of `nullptr` results to an empty `std::string`. Previously the code would crash in the `std::string` constructor if `nullptr` result was returned.

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -697,7 +697,7 @@ void RendererOpenGL::initGL()
 	std::cout << "\tRenderer: " << driverName() << std::endl;
 	std::cout << "\tDriver Version: " << glString(GL_VERSION) << std::endl;
 
-	std::string glsl_v = glString(GL_SHADING_LANGUAGE_VERSION);
+	auto glsl_v = glString(GL_SHADING_LANGUAGE_VERSION);
 	if (glsl_v.empty())
 	{
 		throw renderer_no_glsl();
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -697,13 +697,13 @@ void RendererOpenGL::initGL()
 	std::cout << "\tRenderer: " << driverName() << std::endl;
 	std::cout << "\tDriver Version: " << glString(GL_VERSION) << std::endl;
 
-	auto glsl_v = glString(GL_SHADING_LANGUAGE_VERSION);
-	if (glsl_v.empty())
+	auto glShadingLanguageVersion = glString(GL_SHADING_LANGUAGE_VERSION);
+	if (glShadingLanguageVersion.empty())
 	{
 		throw renderer_no_glsl();
 	}
 
-	std::cout << "\tGLSL Version: " << glsl_v << std::endl;
+	std::cout << "\tGLSL Version: " << glShadingLanguageVersion << std::endl;
 
 	glEnable(GL_TEXTURE_2D);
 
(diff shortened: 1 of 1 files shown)


src/Renderer/RendererOpenGL.cpp
```diff
@@ -58,6 +58,13 @@ namespace {
 
 	GLfloat vertexArray[12] = {}; /**< Vertex array for quad drawing functions (all blitter functions). */
 	GLfloat textureCoordArray[12] = {}; /**< Texture coordinate array for quad drawing functions (all blitter functions). */
+
+
+	std::string glString(GLenum name)
+	{
+		const auto apiResult = glGetString(name);
+		return apiResult ? reinterpret_cast<const char*>(apiResult) : "";
```
cugone: Consider using braced initialization and moving the cast wrap the `glGetString` call:
```
const auto* apiResult = reinterpret_cast<const char*>(glGetString(name));
return { apiResult ? apiResult : "" };
```

DanRStevens: That looks fine. I'm curious though, is there a particular reason?

cugone: RVO gets invoked instead of a construct-then-copy/move. Makes it so there's no move and no copy in the first place.

DanRStevens: I'm not quite sure I follow. It should be the same either way? The `const char*` is used to construct a `std::string`. As that `std::string` is being directly returned RVO should apply.

cugone: Copy elision versus RVO. six of one, half-a-dozen of the other. :P

DanRStevens: Add obligatory reference to [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision) (which also talks about RVO and NRVO).

DanRStevens: So where do we stand on this now? The suggestions seem largely about style. I can't think of any technical reason for the change.

----

I think I would prefer leaving the cast where it is. It makes more sense to me to catch the raw API type/value as is. In some cases (not applicable here) if you were to interact further with an API, you would want values still in the original data type. Converting data (even if there is no actual runtime cost) seems like it should be done only when absolutely necessary.

----

The braces I could go either way on. They don't seem to do anything, so I'm a bit inclined to leave them off, though you could potentially make a case for readability or style here. Certainly they would be used when constructing `struct` return values.

cugone: All I ask is the optioned be considered. Having done that...

PR #551, opened 2020-03-25
Simplify .gitignore

### LINKED ISSUE #549: The .gitignore file lists versioned files
I noticed the `.gitignore` file lists at least one versioned file: `InstallVcpkgDeps.bat`
The ignore is through the line: `/*.bat`

We should maybe go through that file and clean out stuff that doesn't need to be in there.

From the Linux side, the following ignores are used:
```
.build/
lib/
```

DanRStevens:
Closes #549

I believe this is the minimal set of ignores. If we need more, we can add more, though we should probably try to be minimal about it so the file contents remain easy to audit.


DanRStevens:
I've tested this with VS, and there are no extra files generated that are not caught by this.

Can you provide examples of what files are generated but not ignored?

cugone:
I blew out the `.vs/` folders and all the output folders and did a full rebuild of every target. Git tried to add the following to source control:

```
proj/vs2019/.vs/NAS2D/vs16/.suo
proj/vs2019/.vs/NAS2D/vs16/Browse.VC.db
proj/vs2019/.vs/NAS2D/vs16/Browse.VC.opendb
proj/vs2019/.vs/NAS2D/vs16/Solution.VC.db
proj/vs2019/.vs/NAS2D/vs16/Solution.VC.opendb
proj/vs2019/.vs/NAS2D/vs16/Solution.VC.db-shm
proj/vs2019/.vs/NAS2D/vs16/Solution.VC.dc-wal
```

There is a `./vs` folder for both the solution AND the project files, location relative to both, so its location will change. A blanket ignore `*/.vs` is required, especially if the files are moved around during development.

The template that was used is directly from GitHub so I would trust that its original unmodified values are there for a reason.

DanRStevens:
That template has stuff in it that's super old. A lot of it doesn't seem to be relevant anymore. Like VS6 specific files, project types we don't use, and tools we don't use and have no reason to use.

Having a smaller file is easier to maintain.

----

Using the syntax `.vs/` will ignore any folder named `.vs` in any subfolder, including the repository root. If you want it to be tied to the repository root, then prefix with `/`, as in `/.build/` or `/lib/`.

I'm not completely certain, but I suspect the syntax `*/.vs` would ignore `.vs` (file or folder) that was nested one folder level deep.

----

I'm not too sure what you mean by both the solution and project files. I've only ever seen a single `.vs/` folder, though I imagine the contents could contain various items.

DanRStevens:
Ahh, I see now. I was building from the OPHD solution, which didn't generate a `.vs/` folder for NAS2D. It had just a single `.vs/` folder for OPHD.

If instead I open the NAS2D solution, then building will create a `.vs/` folder for NAS2D.

Issue #554, opened 2020-03-26
FpsCounter Bugs


DanRStevens:
In `FpsCounter` there appears to be an index error, where the index ranges from 1 .. ArraySize, rather than from 0 .. (ArraySize - 1).
```cpp
	FPS_COUNTS[++INDEX] = 1000 / TICK_DELTA;

	if (INDEX >= FPS_COUNTS_SIZE) { INDEX = 0; }
```

An easy fix would be to increment after use, and to immediately wrap it after increment:
```cpp
	FPS_COUNTS[INDEX] = 1000 / TICK_DELTA;
	++INDEX;
	if (INDEX >= FPS_COUNTS_SIZE) { INDEX = 0; }
```

----

The initial tick value is initialized to 0, which means the first calculated delta may be surprisingly large, depending on SDL startup time relative to construction of the `FpsCounter` object:
```cpp
	LAST_TICK = CURRENT_TICK;
	CURRENT_TICK = SDL_GetTicks();

	TICK_DELTA = CURRENT_TICK - LAST_TICK;
```

The current tick value should probably be initialized to `SDL_GetTicks()` during construction.

Of course the effects of the initial value are quite temporary in nature, and perhaps not a big concern. Especially since the error is weighted in a direction where it may not matter much, and would be similar to a lack of data, or the initialization value.

cugone:
Howard Hinnant (the guy who *wrote* the std::chrono library!) did a great write-up on how to use `std::chrono` with the `SDL_GetTicks`. It might be useful. 

https://stackoverflow.com/a/59446610/421178

DanRStevens:
Thank you for that. I was looking at `FpsCounter` earlier today, and wondering if we even need it. It's not used within the NAS2D library itself, and the only use in OPHD is not actually used for anything. Was thinking perhaps we could scrap it, or use a standard library replacement. I thought it might be a bit early to make the call though. I figured for now, at least fix any warnings or glaring bugs, if it is easy.

PR #555, opened 2020-03-26
Refactor FpsCounter

DanRStevens:
Reference: #528 (`-Wmissing-variable-declarations`)

Noticed a few bugs (#554) while working on this. As fixing them involves a behaviour change, I wanted to keep them separate from refactoring changes.

Refactors the `FpsCounter` class to remove global variables, and replace them with member variables and function local variables. Rename them from ALL_CAPS to camelCase.

### DIFF
--- a/include/NAS2D/FpsCounter.h
+++ b/include/NAS2D/FpsCounter.h
@@ -26,6 +26,13 @@ class FpsCounter
 {
 public:
 	unsigned int fps();
+
+private:
+	static constexpr unsigned int FpsCountsSize = 25;
+	unsigned int fpsCounts[FpsCountsSize] = { 0 };
+
+	unsigned int currentTick = 0;
+	unsigned int fpsCountIndex = 0;
 };
 
 } // namespace
--- a/include/NAS2D/FpsCounter.h
+++ b/include/NAS2D/FpsCounter.h
@@ -25,8 +25,6 @@ namespace NAS2D {
 class FpsCounter
 {
 public:
-	FpsCounter();
-
 	unsigned int fps();
 };
 
--- a/include/NAS2D/FpsCounter.h
+++ b/include/NAS2D/FpsCounter.h
@@ -17,10 +17,6 @@ namespace NAS2D {
  *
  * FPS values are only approximates. As the FPS count gets higher, the returned value
  * becomes a more average count.
- *
- * \note	Because of the nature of FPS counters and their limited context, the
- *			FpsCounter has been designed with shared state. This means that multiple
- *			instances of FpsCounter will share values.
  */
 class FpsCounter
 {
--- a/src/FpsCounter.cpp
+++ b/src/FpsCounter.cpp
@@ -28,7 +28,7 @@ unsigned int FpsCounter::fps()
 	const auto lastTick = currentTick;
 	currentTick = SDL_GetTicks();
 
-	auto tickDelta = std::min(currentTick - lastTick, 1u);
+	const auto tickDelta = std::min(currentTick - lastTick, 1u);
 
 	fpsCounts[++fpsCountIndex] = 1000 / tickDelta;
 
--- a/src/FpsCounter.cpp
+++ b/src/FpsCounter.cpp
@@ -14,14 +14,17 @@
 
 using namespace NAS2D;
 
-const unsigned int FPS_COUNTS_SIZE = 25;
-unsigned int FPS_COUNTS[FPS_COUNTS_SIZE] = { 0 };
 
-unsigned int	CURRENT_TICK		= 0;
-unsigned int	LAST_TICK			= 0;
-unsigned int	TICK_DELTA			= 0;
-unsigned int	INDEX				= 0;
-unsigned int	ACCUMULATOR			= 0;
+namespace {
+	const unsigned int FPS_COUNTS_SIZE = 25;
+	unsigned int FPS_COUNTS[FPS_COUNTS_SIZE] = { 0 };
+
+	unsigned int	CURRENT_TICK		= 0;
+	unsigned int	LAST_TICK			= 0;
+	unsigned int	TICK_DELTA			= 0;
+	unsigned int	INDEX				= 0;
+	unsigned int	ACCUMULATOR			= 0;
+}
 
 
 /**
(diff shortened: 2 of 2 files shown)


src/FpsCounter.cpp
```diff
@@ -14,44 +14,27 @@
...
 
-	if (TICK_DELTA == 0) { TICK_DELTA = 1; }
+	auto tickDelta = currentTick - lastTick;
+	if (tickDelta == 0) { tickDelta = 1; }
 
-	FPS_COUNTS[++INDEX] = 1000 / TICK_DELTA;
+	fpsCounts[++fpsCountIndex] = 1000 / tickDelta;
 
-	if (INDEX >= FPS_COUNTS_SIZE) { INDEX = 0; }
+	if (fpsCountIndex >= FpsCountsSize) { fpsCountIndex = 0; }
```
cugone: This is a slower version of `fpsCountIndex %= FpsCountsSize` due to the branch.

Also, avoid starting a variable name with a capital letter.

src/FpsCounter.cpp
```diff
@@ -14,44 +14,27 @@
...
-	for (auto i : FPS_COUNTS)
+	unsigned int accumulator = 0;
+	for (auto i : fpsCounts)
 	{
-		ACCUMULATOR += i;
+		accumulator += i;
 	}
 
-	return ACCUMULATOR / FPS_COUNTS_SIZE;
+	return accumulator / FpsCountsSize;
```
cugone: Refactor this to: `std::accumulate(std::begin(fpsCounts), std::end(fpsCounts), 0) / FpsCountsSize;`

`std::reduce` is faster, but I'm not sure of it's adoption yet due to the `std::execution::par` requirements.

src/FpsCounter.cpp
DanRStevens: The `%` operator effectively uses division. Is is unlikely to be faster than a branch. An exception would be when the modulus is a power of 2, in which case it can be translated to a fast bit mask. In this case though, we are not using a power of 2.

The `%` operator would be less code, and I was tempted to use it.

DanRStevens: The `FpsCountsSize` is a constant, hence the UpperCamelCase.

src/FpsCounter.cpp
DanRStevens: I was considering using a standard library algorithm here, but wanted to defer on that. The syntax I saw for working examples was less clear the the original code here.

DanRStevens:
The `std::chrono` stuff looks interesting. I may need a bit of time to read up on it though.

src/FpsCounter.cpp
cugone: Offline testing shows a branch vs modulo to be between 3x and 500 microseconds faster depending on the cutoff and if it's a power of two or not. I'm okay with that.

---

As for the name being capitalized "because it's a constant" (I assume for "at a glance" as well), that's not a good reason. The type system and Intellisense does that for you.

DanRStevens: It's a pretty common convention to use UpperCamelCase for global constants. You can already find examples of it in the code base.

This is for defined constants. Not `const&` references to mutable variables.

src/FpsCounter.cpp
cugone: This day and age, `std::accumulate` is a well known algorithm. It clearly means "Add these up", `std::reduce` on the other hand: "reduce these to a single value using this function" is quote-unquote more clear, because it's confusing to use std::accumulate to *multiply* values.

src/FpsCounter.cpp
cugone: But it's not a global constant. It's a private static member variable.

DanRStevens: A `static` class variable is effectively the same as a global `const`. They're used in equivalent contexts. They're both stored in the same area of the data section. They're initialized the same way.

Actually, I might also recommend UpperCamelCase even if it was function local.

cugone: The wiki doesn't have a style guide. :/ Until otherwise, I'll just live with it.

include/NAS2D/FpsCounter.h
```diff
@@ -17,17 +17,18 @@ namespace NAS2D {
...
  */
 class FpsCounter
 {
 public:
-	FpsCounter();
-
 	unsigned int fps();
+
+private:
+	static constexpr unsigned int FpsCountsSize = 25;
```
cugone: Consider making this `60` to compare the average across the minimally ideal 60 FPS every gamer strives for.

Adding min/max FPS as well is a useful metric.

src/FpsCounter.cpp
DanRStevens: I've sent Leeor a question about it. It should probably be his decision. Most of the existing global constants used `UPPER_SNAKE_CASE`, which should be reserved for macros, and which we are moving away from. There doesn't seem to be well established practice for what to move to.

include/NAS2D/FpsCounter.h
DanRStevens: This is maybe starting to get out of scope with functional changes. This PR was meant to be  a refactor.

Though yes, I was considering changing that constant. It seemed a little arbitrary.

More long term, I'm questioning the calculation used for the FPS. It feels a bit inverted to me. Leeor wanted something stable that wouldn't jump around rapidly. Perhaps there are other ways to do the averaging to get a stable result. Maybe we can store the time delta, rather than an inverted FPS estimation based on the time for each individual frame.

cugone: Milliseconds per frame is a good metric that most game devs use. Gamers on the other hand, only understand FPS.

I'm okay with either, just pick one and stick with it. :)

DanRStevens: I like knowing milliseconds per frame. Though yes, we should display FPS to end users.

Not sure yet on the internal representation to use. We can decide on that for future work. Probably look into the `std::chrono` stuff first.

DanRStevens:
Ok, quick summary of changes that I plan to defer to the future:
- Use of `std::chrono`
- Changing the buffer size
- Changing the internal tracking metric (frame delay, versus FPS estimated per frame)
- Providing min/max FPS (or min/max frame delay)
- Capitalization of constants (Let's ask Leeor about desired style)

PR #560, opened 2020-03-26
Fix double promotion

DanRStevens:
Reference: #528 (`-Wdouble-promotion`)

Fix warnings about promotion of `float` to `double`.

In one case, a `double` was intended.
Reference: [`std::stod`](https://en.cppreference.com/w/cpp/string/basic_string/stof)

### DIFF
--- a/src/Trig.cpp
+++ b/src/Trig.cpp
@@ -76,8 +76,8 @@ bool NAS2D::lineIntersectsCircle(const Point_2d& p, const Point_2d& q, const Poi
 
 	float t = -(((p.x() - c.x()) * dx) + ((p.y() - c.y()) * dy)) / ((dx * dx) + (dy * dy));
 
-	if (t < 0.0) { t = 0.0; }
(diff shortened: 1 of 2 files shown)


src/Trig.cpp
```diff
@@ -76,8 +76,8 @@ bool NAS2D::lineIntersectsCircle(const Point_2d& p, const Point_2d& q, const Poi
 
 	float t = -(((p.x() - c.x()) * dx) + ((p.y() - c.y()) * dy)) / ((dx * dx) + (dy * dy));
 
-	if (t < 0.0) { t = 0.0; }
-	else if (t > 1.0) { t = 1.0; }
+	if (t < 0.0f) { t = 0.0; }
+	else if (t > 1.0f) { t = 1.0; }
```
cugone: This is just `t = std::clamp(t, 0.0f, 1.0f);`

DanRStevens: Indeed it is.

PR #561, opened 2020-03-26
Explicit override

DanRStevens:
Reference: #528 (`-Winconsistent-missing-destructor-override`)

Add `override` keyword for derived class destructors.

Remove explicit default destructors in derived classes.

### DIFF
--- a/include/NAS2D/Mixer/MixerNull.h
+++ b/include/NAS2D/Mixer/MixerNull.h
@@ -8,12 +8,6 @@ namespace NAS2D
 	class MixerNull : public Mixer
 	{
 	public:
-		MixerNull() = default;
-		MixerNull(const MixerNull& other) = default;
-		MixerNull(MixerNull&& other) = default;
-		MixerNull& operator=(const MixerNull& other) = default;
-		MixerNull& operator=(MixerNull&& other) = default;
-
 		virtual void playSound(Sound& sound) override;
 
 		virtual void stopSound() override;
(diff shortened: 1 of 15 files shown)


include/NAS2D/Mixer/MixerNull.h
```diff
@@ -13,7 +13,6 @@ namespace NAS2D
 		MixerNull(MixerNull&& other) = default;
 		MixerNull& operator=(const MixerNull& other) = default;
 		MixerNull& operator=(MixerNull&& other) = default;
-		virtual ~MixerNull() = default;
```
cugone: This is inconsistent with other changes where `= default` was *added*. Did you mean to mark it override or leave it as-is and just got caught up in the sweeping changes, instead?

DanRStevens: That was intentional. It was done as it's own commit because it was different from the rest.

Since `MixerNull` is a derived class, there is no reason to write an explicit destructor if it's not adding anything to it. There is already a `default` `virtual` destructor from the base class. Better to just omit it entirely.

cugone: If you only remove the destructor you are violating rule [CS.21: If you define or delete any default operation, define or delete them all](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-default-operation-define-or-delete-them-all) by leaving the rest.

See also: [Special Member Functions](https://en.wikipedia.org/wiki/Special_member_functions)

Since this class is designed to do nothing, recommend Rule of Zero.

DanRStevens: I agree. They should all be removed.

PR #562, opened 2020-03-26
Fix indexing bug in FpsCounter

### LINKED ISSUE #554: FpsCounter Bugs

DanRStevens:
Closes #554
Depends on #555

Fixes indexing bug in `FpsCounter`.

### DIFF
--- a/src/FpsCounter.cpp
+++ b/src/FpsCounter.cpp
@@ -30,8 +30,9 @@ unsigned int FpsCounter::fps()
 
 	const auto tickDelta = std::min(currentTick - lastTick, 1u);
 
-	fpsCounts[++fpsCountIndex] = 1000 / tickDelta;
+	fpsCounts[fpsCountIndex] = 1000 / tickDelta;
 
+	++fpsCountIndex;
 	if (fpsCountIndex >= FpsCountsSize) { fpsCountIndex = 0; }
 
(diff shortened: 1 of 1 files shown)


DanRStevens:
Rebased after merge of base branch. This makes the change set introduced here more obvious and tightly focused.

PR #570, opened 2020-03-28
Rename colors to be consistent

cugone:
Naming (0, 255, 0) "Bright Green" is inconsistent with the other primary and secondary colors of Red, Blue, Yellow, Cyan, etc. in the sRGB color space.


DanRStevens:
That had actually been bothering me too.

PR #571, opened 2020-03-28
Adds raw numeric constructor and helper functions

cugone:
Adds a numeric constructor for Colors.

Intended for hex-value construction when dealing with somewhat complicated values: `Color morning_blue{0x8DA399FF};`.

As a side effect generating a random color is much easier:

```cpp
Color c = Color::RandomWithAlpha();
//...
Color Color::RandomWithAlpha()
{
    return Color{MathUtils::RandomIntLessThan(0x100000000u)};
}
```
and random full-alpha colors:
```cpp
Color c = Color::Random();
//...
Color::Random()
{
    return Color{(MathUtils::RandomIntLessThan(0x1000000) << 8) | 0xFF};
}
```

### DIFF
--- a/src/Renderer/Color.cpp
+++ b/src/Renderer/Color.cpp
@@ -48,6 +48,40 @@ const Color Color::NoAlpha(0, 0, 0, 0);
 Color::Color(uint8_t r, uint8_t g, uint8_t b, uint8_t a) : mR(r), mG(g), mB(b), mA(a)
 {}
 
+Color::Color(uint32_t value)
+{
+	setRgbaFromRawValue(value);
+}
+
+void Color::setRgbaFromRawValue(uint32_t value)
+{
+	mR = static_cast<uint8_t>((value & 0xFF000000u) >> 24);
+	mG = static_cast<uint8_t>((value & 0x00FF0000u) >> 16);
+	mB = static_cast<uint8_t>((value & 0x0000FF00u) >> 8);
+	mA = static_cast<uint8_t>((value & 0x000000FFu) >> 0);
+}
+
+void Color::setRgbFromRawValue(uint32_t value)
+{
+	mR = static_cast<uint8_t>((value & 0xFF000000u) >> 24);
+	mG = static_cast<uint8_t>((value & 0x00FF0000u) >> 16);
+	mB = static_cast<uint8_t>((value & 0x0000FF00u) >> 8);
+}
+
+uint32_t Color::getRgbaAsRawValue() const
+{
+	return static_cast<uint32_t>(((static_cast<uint32_t>(mR) << 24) & 0xFF000000u)
+								| ((static_cast<uint32_t>(mG) << 16) & 0x00FF0000u)
+								| ((static_cast<uint32_t>(mB) << 8) & 0x0000FF00u)
+								| ((static_cast<uint32_t>(mA) << 0) & 0x000000FFu));
+}
+uint32_t Color::getRgbAsRawValue() const
+{
+	return static_cast<uint32_t>(((static_cast<uint32_t>(mR) << 24) & 0xFF0000u)
+								| ((static_cast<uint32_t>(mG) << 16) & 0x00FF00u)
+								| ((static_cast<uint32_t>(mB) << 8) & 0x0000FFu));
+}
+
 
 /**
  * Sets a Color with a given RGBA value set.
--- a/include/NAS2D/Renderer/Color.h
+++ b/include/NAS2D/Renderer/Color.h
@@ -40,9 +40,15 @@ public:
 
 	Color() = default;
 	Color(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255);
+	explicit Color(uint32_t value);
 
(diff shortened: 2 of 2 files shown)


src/Renderer/Color.cpp
```diff
@@ -48,6 +48,40 @@ const Color Color::NoAlpha(0, 0, 0, 0);
...
+{
+	setRgbaFromRawValue(value);
+}
+
+void Color::setRgbaFromRawValue(uint32_t value)
+{
+	mR = static_cast<uint8_t>((value & 0xFF000000u) >> 24);
+	mG = static_cast<uint8_t>((value & 0x00FF0000u) >> 16);
+	mB = static_cast<uint8_t>((value & 0x0000FF00u) >> 8);
+	mA = static_cast<uint8_t>((value & 0x000000FFu) >> 0);
```
DanRStevens: This has implied endian assumptions.

cugone:
> Couple of concerns here. One is there is an implied endian assumption throughout most of the code here. We my want to pay special attention to how each field is defined and used. We may need to add some documentation detailing this concern.

The class could be self-documenting be renaming it `Rgba`. That way there could be say, `Bgra` and `Argb` classes that imply their byte orders.

> Another concern is we probably want to transition `Color` to have public fields, much like `Point` and `Rectangle`. Being able to use structs as aggregate types is useful. Having constructors gets in the way of that.
> 
> The existing constructors would disappear if the member fields became public. The new constructor you added wouldn't. Perhaps this might work better as a factory function.

Public fields don't affect the ability to have a special constructor. Just default-construct the SMFs:

```
//...
public:
Color()=default;
Color(const Color&)=default;
Color(Color&&)=default;
~Color()=default;
Color& operator=(const Color&)=default;
Color& operator=(Color&&)=default;
explicit Color(uint32_t value);

uint8_t r{0};
uint8_t g{0};
uint8_t b{0};
uint8_t a{0};

```

There have been cases where I've given POD `struct`s constructors out of convenience back in the day and they worked just fine.

DanRStevens:
The order of the fields is already set by the struct. The order of an `int` or `unsigned int` is not set. That's machine dependent. If the struct order is RGBA then an `unsigned int` overlaid on top of it would use the ordering:
Little endian: `0xAABBGGRR`
Big endian: `0xRRGGBBAA`

That might lead to people using different orderings on different platforms when accessing the fields through an `unsigned int`. If we avoid such access, we side step the problem.

----

Structs can of course have constructors, but it makes them non-aggregate.
[What are Aggregates and PODs and how/why are they special?](https://stackoverflow.com/questions/4178175/what-are-aggregates-and-pods-and-how-why-are-they-special)

> An aggregate is an array or a class (clause 9) with no user-declared constructors (12.1), no private or protected non-static data members (clause 11), no base classes (clause 10), and no virtual functions (10.3).

I would prefer these types being aggregate types.

In terms of implications:
> All POD classes are aggregates, or, to put it the other way around, if a class is not an aggregate then it is sure not a POD

Using a factory function is one way to construct such objects without defining a user-declared constructor. And you might as well use a factory method, since all fields being public means you don't need a member method to initialize potentially private fields.

src/Renderer/Color.cpp
```diff
@@ -48,6 +48,40 @@ const Color Color::NoAlpha(0, 0, 0, 0);
...
+}
+
+void Color::setRgbFromRawValue(uint32_t value)
+{
+	mR = static_cast<uint8_t>((value & 0xFF000000u) >> 24);
+	mG = static_cast<uint8_t>((value & 0x00FF0000u) >> 16);
+	mB = static_cast<uint8_t>((value & 0x0000FF00u) >> 8);
+}
```
DanRStevens, line 69: I was just thinking, this leaves the alpha channel unset. I think for this purpose it would make more sense if the alpha channel was set to `255`.

src/Renderer/Color.cpp
DanRStevens, line 61: This actually seems like a reasonable assumption concerning endianess. It assumes red is in the higher order bits, and blue in the lower order bits. This is consistent with how hex color codes are typically written, such as in HTML. As they are translated, they are written to named fields. This seems like a reasonably safe thing to do.

src/Renderer/Color.cpp
cugone: It's intentional. The function name even says so. The alpha value defaults to 255 anyway via the class initializer. This makes raw values that are only 24-bit RGB values from older standards easier to convert to 32-bit by: `setRgbFromRawValue(raw24bit);` (you can then set the alpha value to something other than 255); or, given an external 32-bit value that you want to ignore the alpha value entirely: `setRgbFromRawValue(raw32bit)`.

If you want to set all channels at once, use the `setRgbaFromRawValue` version.

DanRStevens: This isn't a constructor though. The alpha channel could be set to anything at the point when this function is called.

Actually, I don't think we need a set function. An initialization function should be enough. I think we should avoid modifiers.

PR #575, opened 2020-03-29
Adds 8 shades of gray

### LINKED ISSUE #573: Add Color::*Gray
Add shades of common greys:

```cpp
static Color::Gray{128,128,128};
static Color::LightGray{192,192,192};
static Color::DarkGray{64,64,64};
```

Consider also the colors listed on the [wikipedia page](https://en.wikipedia.org/wiki/Grey)

cugone:
Closes #573 

Silver and LightGrey are the same, but the former is an "artist color" and the latter is a "programmer color". Same values, different meanings. Similar to "Normal" and "White"...Although I would argue that distinction is not needed either because it's a well known requirement for shaders. Like how the default "do nothing" for emissive textures is Black; the "do nothing" for a normals/UV texture is `{0, 0, 128}`; and the "do nothing" color for Bump/Heightmaps is "Programmer Gray". You just have to learn and know what the different colors mean in each context.


cugone:
I got the warm and cool versions from the wikipedia page. 6% of 256 is 15, so I added 15 to the appropriate RGB values to get "6% yellow" and "6% blue"

cugone:
> I would prefer programmer names for colors, since these constants appear in source code. There doesn't seem to be much reason to use artistic names. Though I noticed the linked Wikipedia article distinguishes between "Silver" and "LightGrey", giving them different color values.
> 
> Btw, are we settling on American spelling? I thought that was "Gray".

Silver existed prior, I just added LightGrey. "Grey" instead of "Gray" is keeping consistent with the previously existing color "Grey". I can change them all at once now, if you want.

DanRStevens:
Probably a good idea to use consistent spelling. That would mean gray color, not grey colour.

PR #578, opened 2020-03-31
Silence warning SDL_bool

DanRStevens:
Reference: https://github.com/OutpostUniverse/OPHD/issues/308

### DIFF
--- a/src/Renderer/RendererOpenGL.cpp
+++ b/src/Renderer/RendererOpenGL.cpp
@@ -619,6 +619,9 @@ void RendererOpenGL::resizeable(bool resizable)
 		return;
 	}
 
+	#if defined(_MSC_VER)
+	#pragma warning(suppress: 26812) // C26812 Warns to use enum class (C++), but SDL is a C library
+	#endif
 	SDL_SetWindowResizable(underlyingWindow, resizable ? SDL_TRUE : SDL_FALSE);
 }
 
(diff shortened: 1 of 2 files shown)


src/Renderer/RendererOpenGL.cpp
```diff
@@ -619,6 +619,9 @@ void RendererOpenGL::resizeable(bool resizable)
 		return;
 	}
 
+	#if defined(_MSC_VER)
+	#pragma warning(suppress: 26812) // C26812 Warns to use enum class (C++), but SDL is a C library
+	#endif
 	SDL_SetWindowResizable(underlyingWindow, resizable ? SDL_TRUE : SDL_FALSE);
```
cugone: Wrap this in a pop and push so the suppression does not affect more than it's supposed to:

```cpp
#if defined(_MSC_VER)
#pragma warning(push)
#pragma warning(suppress : 26812)
SDL_SetWindowResizable(underlyingWindow, resizable ? SDL_TRUE : SDL_FALSE);
#pragma warning(pop)
#endif
```

DanRStevens: I used `suppress`. It only applies to one following line. No need for `push`, `disable`, `pop` if you're only silencing for 1 line.

cugone: Oh wow. I never saw that on the documentation page. Cool.

PR #579, opened 2020-03-31
Treat warnings as errors

cugone:
Normalized Warning Level for the test projects to Warning Level 4 as well.

----

Edit:
Related:
- Issue #528

### DIFF
--- a/test-graphics/TestGraphics.cpp
+++ b/test-graphics/TestGraphics.cpp
@@ -40,13 +40,13 @@ NAS2D::State* TestGraphics::update()
 
 	{
 		const auto jitter = [&]() -> decltype(auto) {
-			std::uniform_int_distribution<int> dist(0, 64);
+			std::uniform_int_distribution<uint8_t> dist(0, 64);
 			return dist(generator);
 		};
-		for (int i = 0; i < 2000; ++i)
+		for (auto i = 0u; i < 2000u; ++i)
 		{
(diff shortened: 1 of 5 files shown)


test-graphics/TestGraphics.cpp
```diff
@@ -40,13 +40,13 @@ NAS2D::State* TestGraphics::update()
 
 	{
 		const auto jitter = [&]() -> decltype(auto) {
-			std::uniform_int_distribution<int> dist(0, 64);
+			std::uniform_int_distribution<uint8_t> dist(0, 64);
 			return dist(generator);
 		};
-		for (int i = 0; i < 2000; ++i)
+		for (auto i = 0u; i < 2000u; ++i)
```
DanRStevens: Uhh, was this really necessary? The loop variable isn't even used anywhere.

Though to be fair, I do tend to prefer unsigned when possible.

DanRStevens (commented):
Should probably have separated the code fixes from the project setting adjustments. The code fixes could have been applied first.

cugone:
Clicking too fast, I didn't realize the code needed to be changed until after I pushed the change.

test-graphics/TestGraphics.cpp
cugone: The test is kinda bad in the first place. I don't even know what it's supposed to be testing.

DanRStevens (approved):

PR #583, opened 2020-04-01
Fix FpsCounter bug

DanRStevens:
Should use `max` to clamp the low end and set a minimum value.

Noticed a bug I introduced which causes crashes in the test project. The clamping meant to guard against dividing by 0 was actually causing divide by 0.


DanRStevens:
Uhh, that concern may be a bit out of scope for this bug fix.

Additionally I'm not sure we actually want to do that. If you've got the program paused at a breakpoint in debug mode, well of course that's going to affect the frame rate. I would rather not lie about the framerate.

PR #585, opened 2020-04-03
Rename `skip` to `setFrame` in `Sprite` class

### LINKED ISSUE #523: Sprite::skip versus Sprite::setFrame

DanRStevens:
Closes #523

The name update makes the name consistent with the implementation. This should allow callers to have a better expectation of what actually happens when it is called.

Will need a corresponding adjustment in OPHD, where the method is used.

### DIFF
--- a/include/NAS2D/Resources/Sprite.h
+++ b/include/NAS2D/Resources/Sprite.h
@@ -49,7 +49,7 @@ public:
 	void pause();
 	void resume();
 
-	void skip(int frameCount);
+	void setFrame(int frameIndex);
 
 	void update(Point<float> position);
 	void update(float x, float y);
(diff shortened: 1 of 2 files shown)


DanRStevens:
Oops, forgot to save and commit the parameter rename in the header file. Fixed now.

PR #589, opened 2020-04-06
Redirect intermediate files to root folder

### LINKED ISSUE #375: Redirect all intermediate visual studio output to "Temporary" or "Intermediate" folder
It's a pain to do a complete fresh clean and rebuild of the project. Running "Rebuild Solution" doesn't catch everything so it's easier and safer to just delete the intermediate files. The problem is, they are scattered among several different folders and it would be much easier to just delete one root folder and you get a fresh compile immediately.

Similarto linux's `.build` folder, this can be solved by redirecting the intermediate file output to a single folder then further subdirectories split it up by `projectname_configurationname_platformname`. Simply deleting the toplevel (`/proj/vs2019/Temporary` or `proj/vs2019/Intermediate` in Unreal Engine speak) folder will cause a full rebuild of the project/configuration the next time the source is built.

cugone:
Closes #375

Redirects the intermediate files to the root folder `$(SolutionDir)Temporary/$(ProjectName)_$(PlatformShortName)_$(Configuration)`.

Windows-only change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /lib/
 
 # Windows - Visual Studio
+Temporary/
 Release/
 Debug/
 x64/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 /lib/
 
 # Windows - Visual Studio
+[Tt]emp*/
```
DanRStevens, line 6: I would suggest making this match the exact folder name for the build folder.

cugone: Fixed

PR #591, opened 2020-04-08
Color compare

DanRStevens:
Add `Color` comparison operators.

If we make the fields public, we probably get the comparison for free. Until then, these can be used to do full color comparisons. At any rate, the added unit tests will ensure `Color` comparisons are still possible afterwards.


DanRStevens:
I'd like to avoid inequalities for now. There doesn't seem to be a natural ordering between all color values.

And thanks for the reminder that comparison operators are never default provided. For some reason I was thinking there was a default provided member-wise comparison for aggregate types, though that doesn't seem to be the case in C++.

PR #593, opened 2020-04-08
Remove Color::operator()

DanRStevens:
This method is not currently used.

For an alternative in new code, construct a new `Color` object. This can be done by overwriting an object in place:

```
color(r, g, b, a); // Old syntax
color = {r, g, b, a}; // Overwrite with newly constructed object
```

I would claim the intent of the later is perhaps more clear anyway.

### DIFF
--- a/NAS2D/Renderer/Color.cpp
+++ b/NAS2D/Renderer/Color.cpp
@@ -70,23 +70,6 @@ bool Color::operator!=(Color other) const
 
 
 /**
- * Sets a Color with a given RGBA value set.
- *
- * \param r	Red compontent. Valid values are 0 - 255.
- * \param g	Green compontent. Valid values are 0 - 255.
- * \param b	Blue compontent. Valid values are 0 - 255.
- * \param a	Alpha compontent. Valid values are 0 - 255.
- */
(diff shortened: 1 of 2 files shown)


cugone:
Be cautious with the latter syntax as it may invoke unintended behavior. Prefer instead `Color color{r, g, b, a};`

See the long-winded notes and examples in https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

DanRStevens:
To be clear, this is about resetting an already initialized value to a new value. The `operator()` syntax does not work for initializing a new object. It only works on an existing object. Similarly, the `= {}` syntax only works on an existing object.

For a new object, you may initialize with:
```cpp
Color color{r, g, b, a};
```

Once that object is declared, it may then be rest to a new value with:
```cpp
color = {r, g, b, a};
```

It would be an error to use initialization syntax on an already existing object:
```cpp
Color color{r, g, b, a}; // Initial definition

Color color{r, g, b, a}; // Error, redefinition of 'color'
color{r, g, b, a}; // Error, expect ';' after expression
```

PR #600, opened 2020-04-14
Reorder Renderer methods for consistency

DanRStevens:
Generally the packed parameter methods come before the unpacked parameter methods.

### DIFF
--- a/NAS2D/Renderer/Renderer.cpp
+++ b/NAS2D/Renderer/Renderer.cpp
@@ -35,6 +35,12 @@ Renderer::~Renderer()
 }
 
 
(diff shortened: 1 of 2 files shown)


DanRStevens:
Huh, I think there was a bit of user interface lag here. The tests showed as passed, so I merged. Once merged, AppVeyor detected this, and cancelled some pending builds, marking them as failed. It seems the status of the AppVeyor tests was not accurately reflected in the initial display. I noticed sometimes there is lag picking up which tests are running when a PR is first opened. Perhaps it never updated to show the pending AppVeyor runs. I'll have to be more careful about this in the future. No way to re-run the tests now, as AppVeyor just says the PR is already closed.

PR #627, opened 2020-04-18
Add Vector reflect methods

DanRStevens:
Add `Vector::reflectX()` and `Vector::reflectY()` methods.

The `reflectX()` methods reflects the X-coordinate (flips the vector about the Y-axis).
The `reflectY()` methods reflects the Y-coordinate (flips the vector about the X-axis).

Hopefully the meaning of the names isn't too confusing, and won't cause people to swap them by accident.


DanRStevens:
Looks like Codacy is undergoing maintenance and is unable to run. Maybe check on it in the morning to see if it can be re-run.

PR #631, opened 2020-04-20
Mark Font reference as const in drawText calls

DanRStevens:
Mark `Font` reference as const in `drawText` calls. This makes `drawText` usable with `Font` objects that were marked with `const`.


DanRStevens:
Doh, merged too early again. The PR build was still running, and got cancelled once I merged. Looks like the build was actually succeeding though.

PR #639, opened 2020-04-26
Remove splitSkipEmpty

DanRStevens:
This method isn't used. If we really need it, we can bring it back from history.

Additionally, the method seems too highly specialized for the library. If a workaround is needed, a user could potentially use split, and then filter the results to remove empty entries. That may be less efficient, though hard to say how much of an impact it would have without measuring it.


ldicker83:
Agreed. There are plenty of other string manipulation libraries available that this doesn't need to be reinvented.

PR #685, opened 2020-06-27
Remove unneeded use of `startPoint()`

DanRStevens:
Remove unneeded use of `startPoint()` on `Rectangle` instance.


DanRStevens:
Gah, UI update was lagging and I merged before the PR build was done, which cancelled it and caused it to show up as failed. Should be good though.

Issue #302, opened 2020-01-02
Remove reference getters from Point and Rectangle structs


ldicker83:
Both Point and Rectangle structs offer getters that return a non-const refernece to the internal member data. This effectively breaks encapsulation and may as well expose the internal members.

That stated, since the conversion from class to struct in #300 and #268, it may make sense to simply remove get/set functions entirely, rename the internal members from mX/mY (and etc.) to simply x/y (and etc.) and make them public.

The operator overloads should be examined as well as some don't make sense (rectangle multiplication and division? And how is subtraction supposed to work exactly? Hell, how is addition supposed to work?). But that's a separate issue.

DanRStevens:
Small side point, this will necessarily break backwards compatibility with downstream uses. Before I had suggested keeping the accessor methods and deprecating them as a compatibility hack. However, you can't have both a member variable and a member function shared the same name `x` (or `y`). Hence there would be no way to maintain backwards compatibility while offering the member variable `x`.

As other names would be unexpected and add to the confusion, I would agree with Leeor's suggestion of just dropping the accessors and renaming the member variables to `x` and `y`. Access is then through `x` rather than the accessor `x()`. It breaks backwards compatibility, but in a very minor and easy to fix way. I think it's worth it.

PR #687, opened 2020-07-09
Add Rectangle size and startPoint setters

DanRStevens:
Add setter methods for:
- `Rectangle::size`
- `Rectangle::startPoint`

----

This also raises the question on how `Rectangle` should be structured internally. Currently it consists of effectively 4 fields for:
- `x`
- `y`
- `width`
- `height`

Potentially this could be updated to packed internal structure:
- `Point startPoint`
- `Vector size`

That would remove the need for get/set methods if those fields were to become public.

Reference: #302

### DIFF
--- a/NAS2D/Renderer/Rectangle.h
+++ b/NAS2D/Renderer/Rectangle.h
@@ -82,6 +82,16 @@ struct Rectangle
 		return (mW == 0) || (mH == 0);
 	}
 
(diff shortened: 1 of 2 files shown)


DanRStevens:
Whoops, merged too early, before the PR build finished running, which caused it to be cancelled. Should be good though, since with no intermediate changes to master, the PR build should match the branch build.

PR #692, opened 2020-07-10
Add getAngle

### LINKED ISSUE #690: Add Vector to angle method
The `Trig.h` header offers a way to convert unpacked point values to an angle:
```cpp
float angleFromPoints(float x, float y, float x2, float y2);
```

We should offer a way to convert a `Vector` to an angle. In particular, we should be most interested in `Vector<float>` being converted to an angle.

Since a `Vector` already represents a difference between two points, we could have just a single `Vector` parameter, rather than two `Point` parameters. Usage would look something like:
```cpp
const auto angle = getAngle(point2 - point1);
```

DanRStevens:
Closes #690

Add new Trig method `getAngle`, which converts a `NAS2D::Vector<float>` to a `float` angle in degrees.

----

We may want to consider using radians by default, or tagging methods with either Rad or Deg, possibly providing both alternatives.

We may want to consider what coordinate system should be used, such as a Cartesian coordinate system or a screen coordinate system. In particular, which direction corresponds to angle 0 (up? right?), and which direction does the angle increase (clockwise? counterclockwise?).

We may want to consider renaming the methods to improve consistency and clarity.

### DIFF
--- a/NAS2D/Trig.cpp
+++ b/NAS2D/Trig.cpp
@@ -36,7 +36,7 @@ float degToRad(float degree)
  */
 float radToDeg(float rad)
 {
-	return rad * -RAD2DEG;
+	return rad * RAD2DEG;
 }
 
 
@@ -45,7 +45,7 @@ float radToDeg(float rad)
  */
 float angleFromPoints(float x, float y, float x2, float y2)
 {
-	return 90.0f - radToDeg(std::atan2(y2 - y, x2 - x));
+	return 90.0f + radToDeg(std::atan2(y2 - y, x2 - x));
 }
 
 
@@ -57,7 +57,7 @@ float angleFromPoints(float x, float y, float x2, float y2)
  */
 float getAngle(Vector<float> direction)
 {
-	return 90.0f - radToDeg(std::atan2(direction.y, direction.x));
+	return 90.0f + radToDeg(std::atan2(direction.y, direction.x));
 }
 
 
--- a/NAS2D/Trig.cpp
+++ b/NAS2D/Trig.cpp
@@ -50,6 +50,18 @@ float angleFromPoints(float x, float y, float x2, float y2)
 
 
 /**
+ * Gets the angle of a direction vector
+ *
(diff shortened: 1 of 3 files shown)


DanRStevens:
@Brett208, I was going to request review from you, though it seems you're not part of the Lairworks organization.

Brett208:
What if you did the following:

```Cpp
namespace Screen {
    getAngleRadians();
    getAngleDegrees();
}

namespace Cartesian {
    getAngleRadians();
    getAngleDegrees();
}

```

I don't really know what the right answer is though. Your initial answer of clockwise starting from 12 oclock is probably my default answer but I'm a poor mathematician...

DanRStevens:
Hmm, interesting solution. I hadn't thought to use namespaces. I guess I generally try to avoid nesting namespaces.

Of course the translation between zero points, direction of increase, and radians or degrees should be a linear transform. Maybe a base radian method using standard math Cartesian coordinates could be wrapped in various linear translation methods. The zero point would be the addition term, while the direction and unit would be the multiplier.

DanRStevens:
After some thought, I figured we should provide only `radian` measure trigonometry functions, and require explicit wrapping with a function to convert between radians and degrees. The trig functions work natively with radians, so that's the most efficient way to use them. Plus, many trig functions are used in pairs, so the exact unit being used will often cancel out anyway.

Similarly, I thought we should go with the standard Cartesian coordinate system and counterclockwise direction of angle increase. That matches standard math conventions, so should be the least surprising. Converting to angles in a different coordinate basis can be done by wrapping methods, the same as the conversion between degrees and radians. Either way, it's a fairly simple linear transformation between the various angle measurements. Actually, the change in direction and the flipped y-axis both contribute a `-1` multiplier, which cancel each other out, so the difference between the two coordinate systems is simply adding or subtracting 90 degrees. By using the standard math convention, we can avoid adding that offset into every function, and like above, when using functions in pairs, the constant would cancel itself out anyway.

That leaves renaming for better clarity, which I'm going to avoid as out of scope for the moment.

PR #734, opened 2020-07-25
Refactor Sprite

DanRStevens:
General refactoring. Convert error messages to exceptions. Remove extraneous error checks. Const correct XML loading. Remove default `Sprite` constructor.

### DIFF
--- a/NAS2D/Resources/Sprite.cpp
+++ b/NAS2D/Resources/Sprite.cpp
@@ -100,10 +100,7 @@ void Sprite::resume()
  */
 void Sprite::setFrame(std::size_t frameIndex)
 {
-	if (mActions.find(mCurrentAction) != mActions.end())
-	{
-		mCurrentFrame = frameIndex % mActions[mCurrentAction].size();
-	}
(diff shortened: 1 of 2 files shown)


NAS2D/Resources/Sprite.cpp
```diff
@@ -139,12 +139,12 @@ void Sprite::update(Point<float> position)
...
 		{
 			mCurrentFrame = 0;
-			mFrameCallback(); // Notifiy any frame listeners that the animation sequence has completed.
+			mAnimationCompleteCallback(); // Notifiy any frame listeners that the animation sequence has completed.
 		}
 	}
 	else if (frame.frameDelay == FRAME_PAUSE)
 	{
-		mFrameCallback();
+		mAnimationCompleteCallback();
```
ldicker83, line 127: You know, I'm thinking in a case like this the callback shouldn't get called. It's a single frame, why would we call this every update?

DanRStevens: You know, I was kind of wondering about that event trigger too. I suppose it was intended to trigger when it first hit a frame with a `frameDelay` of `FRAME_PAUSE`?

Issue #763, opened 2020-08-03
Remove internal caching of Resources


DanRStevens:
Certain resource classes implement internal caching. This makes for a rather rigid design. We should instead remove this internal caching, and instead rely on external caching, such as is provided by the `ResourceCache` class (#760).

The impacted classes are:
- `Font`
- `Image`
- `Music`

----

Related prep-work: #401

DanRStevens:
While looking at the `Image` class, there seems to be some differences between the `SDL_Surface` field, and the `textureId` and `frameBufferObjectId` fields, which impacts the appropriateness of where they are stored, and how/if they are cached.

I'm kind of thinking `Image` should be split to handle just the `SDL_Surface` portion of the code, while the `textureId` and perhaps the `frameBufferObjectId` could become a `Texture` class. Only `RendererOpenGL` really cares about the later fields. They're just stored with `Image` for convenience.

Of course, such a split would open up a few design issues. In particular, what object type should `Renderer` accept as arguments? The texture fields are specific to OpenGL, so a `Texture` class based on them would be specific to `RendererOpenGL`. It would not apply equally well to the other `Renderer` types. How might we maintain a neutral interface where any image can be passed to any kind of `Renderer`? One possibility might be to cache private `Texture` objects inside the `Renderer` objects, which are private and specific to the type of `Renderer`. Perhaps an internal map could map generic `Renderer` neutral `Image` objects to cached `Renderer` specific `Texture` objects.

Another issue would be sync between an `Image` and a `Texture`. An `Image` could be modified after a `Texture` was generated from it. If the generation of a `Texture` was explicit, that would make reasonable sense, and the user might be required to regenerate a new `Texture` if they want to see the update. If however the `Texture` was automatically converted and cached, like suggested above, how would the cache know to regenerate an updated `Texture`? This makes the caching idea seem much less clean.

Of course, the current code likely already has this issue, so maybe the sync issue should be considered separately at a later time.

----

[...]

PR #784, opened 2020-08-11
Resource copy operations

DanRStevens:
Reference: #763

Removed some resource copy operations impacted by the removal of internal caching. As default implemented they were unsafe, and so were marked as `deleted`.

Marked some special methods as protected and gave friend access where needed. Use proper types and removed uses of `static_cast`. Did some final cleanup to make the classes more consistent with each other.

### DIFF
--- a/NAS2D/Resources/Music.h
+++ b/NAS2D/Resources/Music.h
@@ -39,7 +39,7 @@ public:
 
 protected:
 	friend class MixerSDL;
-	void* music() const;
+	Mix_Music* music() const;
 
 private:
 	std::string mResourceName; /**< File path */
--- a/NAS2D/Resources/Music.h
+++ b/NAS2D/Resources/Music.h
@@ -37,8 +37,8 @@ public:
 
 	const std::string& name() const { return mResourceName; }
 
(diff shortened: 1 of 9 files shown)


NAS2D/Resources/Music.h
```diff
@@ -39,7 +39,7 @@ class Music
 
 protected:
 	friend class MixerSDL;
-	void* music() const;
+	Mix_Music* music() const;
```
ldicker83, line 42: This pollutes the interface with low-level implementation details which I'd hoped to avoid. Not going to nitpick it but that was the justification here for use of void pointer.

DanRStevens: I know. I'm not sure of a clean way to allow needed access, while keeping `Music` separate from `MixerSDL`.

That's why it's a `protected` method with `friend` only access. The only place that can use it, `MixerSDL`, will already be importing the SDL-mixer headers. The type is just forward declared for this file, to prevent header pollution affecting other users of the `Music` class.

Granted, it does create the small bit of pollution that there is an incomplete global type named `Mix_Music`, that will be defined somewhere.

Issue #739, opened 2020-07-26
Generalize Configuration class


DanRStevens:
The `Configuration` class contains quite a bit of code specific to `Mixer` and `Renderer`. It can be generalized to support loading of arbitrary configuration data.

Additionally, the `Configuration` class can serve as a bridge between the XML serialized format, and the in memory configuration data, which is largely key/value storage with support for various data types, now implemented as `Dictionary`. It is one of the few places in NAS2D that is heavily dependent on XML code, the other being `Sprite`, and can be used as a basis for the `Sprite`, along with downstream uses in OPHD.

One of the end goals of this work, is to decouple the library and downstream project code from the XML code. That will make it easier to upgrade or replace the TinyXML1 dependency. In particular, it can be made an external dependency installed through `vcpkg`, it can be upgraded to TinyXML2, or it can even possibly be swapped out for another serialization format such as JSON.

Reference: #211, #239

----

A lot of work has already been put into this effort, though was not previously tracked using a specific issue. By opening this issue, that work from various PRs can be collected into a single place, along with new work to complete that goal.

DanRStevens:
Mostly thinking out loud here.

Looking at the `Configuration` class, there are sections of methods relating to video and audio that should be removed. Each method comes in get/set pairs.

Video methods to remove:
```cpp
	// Video Options
	int graphicsWidth() const;
	int graphicsHeight() const;
	int graphicsColorDepth() const;
	bool fullscreen() const;
	bool vsync() const;

	void graphicsWidth(int width);
	void graphicsHeight(int height);
	void graphicsColorDepth(int bpp);
	void fullscreen(bool fullscreen);
	void vsync(bool vsync);
```

Audio methods to remove:
```cpp
	// Audio Options
	int audioMixRate() const;
	int audioStereoChannels() const;
	int audioSfxVolume() const;
	int audioMusicVolume() const;
	int audioBufferSize() const;
	std::string mixer() const;

	void audioMixRate(int mixrate);
	void audioStereoChannels(int channels);
	void audioSfxVolume(int volume);
	void audioMusicVolume(int volume);
	void audioBufferSize(int size);
	void mixer(const std::string& mixer);
```

Similarly the ability to set defaults is currently too specific to video and audio settings, and so may be removed:
```cpp
	void setDefaultValues();
```

Related to the above are the private methods to parse those specific options, which will need to be removed:
```cpp
	void parseGraphics(const Dictionary& dictionary);
[...]

PR #792, opened 2020-08-14
Add method findMissingAndUnexpected

DanRStevens:
Reference: #739

Add method `findMissingAndUnexpected` to help with error reporting. Given a set of `keys`/`names`/`values`, and a set of `required` and `optional` values, find the sets of `missing` and `unexpected` values. This can be used to combine a bunch of individual value name checks into a single comprehensive error report.

The `findMissingAndUnexpected` method is effectively implemented through set minus.

----

Is this method a good idea? It contains 3 parameters of the same type, which could potentially be confused with each other. Unless named values are passed in, this design might not be very self documenting. The unit test code kind of shows the lack of self documentation. Similarly the method returns two collections of the same type.

Is this method needed? Using the custom `operator-` for `std::vector` allows this code to be written inline quite easily. Doing so may be more self documenting due to variable names. That may be less standard though, and so leave room for bugs, such as getting the order of sets swapped. That is the case with the current code on `master`, which contains such a bug.

This method is meant to replace `ReportProblemNames`, which is private to `Configuration`. It is only used 2 times, and one of those uses is scheduled to be removed.

This method may be more useful if combined with an exception throwing wrapper, that throws when either result set is non-empty, and formats a list of the missing and unexpected sets.

Where's the best place to put this new method? It's generic enough that it shouldn't be tied to `Configuration`. Maybe not so generic that it belongs in `ContainerUtils.h`? What about an exception throwing wrapper method?

I kind of feel like this method is a good idea, but I'm not quite certain how to get the most value out of it.

### DIFF
--- a/NAS2D/ContainerUtils.h
+++ b/NAS2D/ContainerUtils.h
@@ -62,4 +62,30 @@ namespace NAS2D {
 		}
 		return result;
 	}
+
+
+	template <typename T>
+	struct MissingAndUnexpected {
+		std::vector<T> missing;
+		std::vector<T> unexpected;
+
+		bool operator==(const MissingAndUnexpected& other) const {
+			return missing == other.missing && unexpected == other.unexpected;
+		}
+		bool operator!=(const MissingAndUnexpected& other) const {
+			return !(*this == other);
+		}
+	};
+
+	template <typename T>
+	MissingAndUnexpected<T> findMissingAndUnexpected(
+		const std::vector<T>& names,
+		const std::vector<T>& required,
+		const std::vector<T>& optional = {}
+	) {
+		using namespace ContainerOperators;
+
+		const auto expected = required + optional;
+		return {required - names, names - expected};
+	}
 }
--- a/test/ContainerUtils.test.cpp
+++ b/test/ContainerUtils.test.cpp
@@ -3,6 +3,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <string>
 #include <array>
 #include <vector>
 #include <list>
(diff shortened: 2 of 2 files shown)


DanRStevens:
@Brett208, can I get your opinion on this change? I feel like I need someone to bounce this idea off of.

DanRStevens:
It seems there isn't such a pressing need to make this method available to the caller. The original code was used to check for required section names. The caller can instead set default values for all required `Configuration` settings, which would remove the need to check for missing required sections. Of course, that's also with the assumption that extra required sections go unreported.

We might want to still optionally provide this feature, for cases where it might be useful. One potential use is to report missing required settings within a section. There might be no obvious default values for certain settings.

Brett208:
I think you are trying to accompolish too much with one function (more than one primary task)
* Using the word and in the function name 
* Returning a custom structure containing 2 containers 

Maybe just:

```C++
std::vector findMissingValues(const std::vector& actualValues, const std::vector& possibleValues);
std::vector findUnexpectedValues(const std::vector& actualValues, const std::vector& possibleValues);
```

This would eliminate ambiguity in how to read the return value and reduce complexity of provided argument inputs. There is still a chance the end user could flip the two arguments. I think this wouldn't be too difficult of a logic error to catch in debugging though and I don't really see a way around it.

I think this function or function set has utility in serialization checking. On the other hand, in the case of a video game saved file, if someone is manipulating the saved game, maybe it doesn't matter if the parsing completes and the game breaks due to their manipulation without warning ahead of time? I would consider leaving the function in place, although discovery of these sort of utility functions can be difficult if they are sort of standalone and not used much in the codebase.

Hopefully that helps some. I forgot to look over yesterday as I had planned.

-Brett

DanRStevens:
I think you're right. The thought of splitting it into two functions had briefly crossed my mind. Initially I thought it would lower efficiency if I split them, but that doesn't actually seem to be the case. To detect `missing` values we need only know the `required` values. It's the `unexpected` values that needs to know both the `required` values and `optional` values, or otherwise be passed a combined set of `expected` values.

Splitting also offers more flexbility, in that some code might consider any extra values to be optional ones. The only error case might be missing required values.

PR #794, opened 2020-08-15
Use freesrc parameter to fix SDL_RWops leaks

### LINKED ISSUE #786: SDL_RWops stream leaks
I believe there may be a leak of [`SDL_RWops`](https://wiki.libsdl.org/SDL_RWops) objects affecting:
- `Font`
- `Image`
- `Music`
- `Sound`

----

Many of the SDL resource handing methods take a parameter `int freesrc`, which is used as a boolean flag. If it is set to 1, the `SDL_RWops` object is automatically freed after the method is done using it, otherwise it must be manually freed.

Links to relevant documentation:
[`TTF_OpenFontRW`](https://www.libsdl.org/projects/SDL_ttf/docs/SDL_ttf_15.html)
[`IMG_Load_RW`](https://www.libsdl.org/projects/SDL_image/docs/SDL_image_12.html)
`Mix_LoadMUS_RW` (No documentation found)
[`Mix_LoadWAV_RW`](https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer_20.html)

Quote from documentation:
> `freesrc`
>   - A non-zero value means it will automatically close and free the src for you after it finishes using the src, even if a noncritical error occured.

Setting the parameter to `1` (`true`) allows stream creation methods to be chained, without worrying about resource leaks.

Example usage:
```cpp
font=TTF_OpenFontRW(SDL_RWFromFile("font.ttf"), 1, 16);
```

----

The above pattern is used by the above mentioned objects, but the code is setting `freesrc` to `0`. This is likely leaking the `SDL_RWops` objects responsible for streaming the input.

Font.cpp:
```cpp
TTF_Font *font = TTF_OpenFontRW(SDL_RWFromConstMem(fontBuffer.raw_bytes(), static_cast<int>(fontBuffer.size())), 0, static_cast<int>(ptSize));
```

Image.cpp:
```cpp
mSurface = IMG_Load_RW(SDL_RWFromConstMem(imageFile.raw_bytes(), static_cast<int>(imageFile.size())), 0);
```

Music.cpp:
```cpp
mMusic = Mix_LoadMUS_RW(SDL_RWFromConstMem(mBuffer.raw_bytes(), static_cast<int>(mBuffer.size())), 0);
```

Sound.cpp:
```cpp
mMixChunk = Mix_LoadWAV_RW(SDL_RWFromConstMem(soundFile.raw_bytes(), static_cast<int>(soundFile.size())), 0);
```

----

I believe we should change the above calls to pass a `freesrc` value of `1`. I believe that would eliminate potential leaks of the associated `SDL_RWops` stream objects.

Bonus if we can find a way to prove there is a memory leak, along with verifying a fix. Perhaps `valgrind` could come in handy for that.

DanRStevens:
Closes #786

Using `valgrind`, I was able to determine this change significantly reduced the number of definitely lost blocks of memory.

Test 1:
Loading game and closing the main window as soon as it displays, before the intro is started.
- Without patch: 14 definitely lost blocks
- With path: 8 definitely lost blocks

Test 2:
Loading game to the main menu, and then exiting.
- Without patch: 114 definitely lost blocks
- With path: 8 definitely lost blocks

Looks like a number of resources are loaded by the time the main menu is displayed, which results in leaked `SDL_RWops`. (I would like to quickly verify this before merging).

### DIFF
--- a/NAS2D/Resources/Font.cpp
+++ b/NAS2D/Resources/Font.cpp
@@ -194,7 +194,7 @@ namespace {
 			throw std::runtime_error("Font file is empty: " + path);
 		}
 
-		TTF_Font *font = TTF_OpenFontRW(SDL_RWFromConstMem(fontBuffer.raw_bytes(), static_cast<int>(fontBuffer.size())), 0, static_cast<int>(ptSize));
+		TTF_Font *font = TTF_OpenFontRW(SDL_RWFromConstMem(fontBuffer.raw_bytes(), static_cast<int>(fontBuffer.size())), 1, static_cast<int>(ptSize));
 		if (!font)
 		{
 			throw std::runtime_error("Font load function failed: " + std::string{TTF_GetError()});
@@ -231,7 +231,7 @@ namespace {
 			throw std::runtime_error("Font file is empty: " + path);
 		}
 
-		SDL_Surface* fontSurface = IMG_Load_RW(SDL_RWFromConstMem(fontBuffer.raw_bytes(), static_cast<int>(fontBuffer.size())), 0);
+		SDL_Surface* fontSurface = IMG_Load_RW(SDL_RWFromConstMem(fontBuffer.raw_bytes(), static_cast<int>(fontBuffer.size())), 1);
 		if (!fontSurface)
 		{
 			throw std::runtime_error("Font loadBitmap function failed: " + std::string{SDL_GetError()});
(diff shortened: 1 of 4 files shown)


DanRStevens:
I verified the numbers on this. They match up well with the results seen in the post above.

For the intro screen, 6 resources need to be loaded (1 `Music`, 5 `Image`). To load the main menu, about an additional 100 resources are loaded, mostly skin images for the window, the buttons, and the text field for the `FileIo` dialog used to load saved games.

Issue #373, opened 2020-01-19
Make use of `vcpkg` friendly for new developers


DanRStevens:
We should find a way to make use of `vcpkg` more friendly for new developers. That might involve better error messages, or increased automation.

Some important tasks:
- Installing `vcpkg`
- Updating `vcpkg`
- Running the `InstallVcpkgDeps.bat` file to install dependencies
- Updating dependencies

We should try to automate it as much as possible. Where automation is not likely to work, we should provide easy to read and understand error messages. We might even add web links for places they need to go to download `vcpkg`.

cugone:
I found an automated script that clones a git repo if it doesn't exist, otherwise pulls it:

https://gist.github.com/nicferrier/2277987

There are some interesting additions to it in the later comments.

DanRStevens:
I like the last comment with the example use of the `-C` option to run a `git` command without having to change the current working directory.

cugone:
Except `/dev/null` isn't a thing on windows. :P

DanRStevens:
Not anymore than the shell syntax. :wink:

----

I did some experimenting with Visual Studio project file features, which might be relevant here.

In #365 I made brief mention of the [`Exec` task](https://docs.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2019).

Example used to capture the `vcpkg` install path:
```
  <Target Name="Find vcpkg" BeforeTargets="PrepareForBuild">
    <Exec Command="where vcpkg" ConsoleToMsBuild="true" >
      <Output TaskParameter="ConsoleOutput" PropertyName="VcpkgExePath" />
    </Exec>
  </Target>
```

Different sources seem to swap out `PropertyName` with `ItemName`, depending on use.

We could potentially use the `Exec` task to automatically install dependencies. We can likely even include a `Condition` which checks if dependency folders exist. I haven't experimented with the condition for that, though running the batch file might look something like:
```
  <Target Name="Install vcpkg dependencies" BeforeTargets="PrepareForBuild">
    <Exec Command="call InstallVcpkgDeps.bat" />
  </Target>
```

You can set variables (conditionally) in the `Globals` section. I expect you can use your own custom sections too:
```
  <PropertyGroup Label="Globals">
    ...
    <VcpkgTriplet Condition="'$(Platform)'=='Win32'">x86-windows</VcpkgTriplet>
    <VcpkgTriplet Condition="'$(Platform)'=='x64'">x64-windows</VcpkgTriplet>
[...]

PR #803, opened 2020-08-30
Update vcpkg options

DanRStevens:
It seems there have been some AppVeyor updates, which have brought in an updated version of `vcpkg`, which has updated install options for the `SDL2-mixer` package. Since a package check is done before building, the sdl2-mixer option discrepancy was causing failures during package installation, which prevented builds from starting.


DanRStevens:
Adding reference to #373, since the `--recurse` option kind of relates to making `vcpkg` easier to use.

Issue #797, opened 2020-08-17
Refactor Sprite XML loading code


DanRStevens:
Related: #739, #211

The `Sprite` XML loading code should be examined and refactored. One goal here is to limit the use of XML code, such that it can be largely decoupled from the NAS2D library. This may make it easier to remove the TinyXML dependency from direct inclusion, allowing us to switch to `vcpkg` dependency installation, while also making it easier to transition to TinyXML2.

One feature needed by the `Sprite` code that is absent from the `Configuration` code is the ability to process arrays of data (the frame list).

Similar to the `Configuration` code, we may need to report errors due to missing required keys, or for extra unexpected keys which are neither required nor optional. This is likely to be true whether the we XML, or JSON, or some other text format to store the data, so such error detecting and reporting code may need to be general and accessible to all such text processing methods.

DanRStevens:
It would be helpful to have template code that could parameterized with an arbitrary `struct`, and then serialize that struct. Or at least convert between a `struct` and a `Dictionary`.

There is no reflection in C++ that would allow automatic extraction of `struct` field names. It's also entirely possible the field names in the serialized form will be different than the field names in the `struct`. As such, we may need to specify field names for the serialized form, which may differ from the `struct` field names.

Usage might look something like:
```cpp
struct SomeData {
  int field1;
  std::string field2;
};

const auto someData = SomeData{1, "data"};
const auto dictionary = structToDictionary<"fieldName1", "fieldName2">(dataStruct);

EXPECT_EQ(Dictionary{{{"fieldName1", 1}, {"fieldName2", "data"}}}, dictionary);
```

----

To support the above, I was doing a bit of reading on templates. The structured binding feature can be used to some advantage here, though there is still a problem of determining how many fields are in any given `struct`. A long while back, I read a bit about the magic_get library, which seems to have come up again. Here are some relevant links:
- [Stack Overflow: get the number of fields in a class](https://stackoverflow.com/questions/45820379/get-the-number-of-fields-in-a-class)
- [Counting the number of fields in an aggregate in C++20](https://towardsdev.com/counting-the-number-of-fields-in-an-aggregate-in-c-20-c81aecfd725c)
- [GitHub: magic_get](https://github.com/apolukhin/magic_get)
- [Boost.PFR](https://www.boost.org/doc/libs/1_75_0/doc/html/boost_pfr.html)

DanRStevens:
[...]

PR #815, opened 2020-10-22
Remove deprecated code to parse Configuration options format

DanRStevens:
Rely solely on the consistent attribute format.

The custom options format was getting in the way of moving generalized code into ParserHelper. As it isn't likely to be used anymore, it should be removed.

**This is a breaking change which removes support for older configuration file formats.**

Reference: #797

### DIFF
--- a/NAS2D/Configuration.cpp
+++ b/NAS2D/Configuration.cpp
@@ -34,37 +34,6 @@ namespace {
 	}
 
 
-	// This method should be scheduled for deprecation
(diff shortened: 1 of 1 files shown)


DanRStevens:
@ldicker83: If you plan to update the config files for older projects that depends on NAS2D, it should be done before merging this. (Though it's probably not overly onerous to update old config files manually if needed).

ldicker83 (approved):

PR #823, opened 2021-01-20
Add implicit string conversion to StringFrom

Brett208:
Incorporates improvement to function discussed https://github.com/OutpostUniverse/OP2Utility/pull/374

### DIFF
--- a/NAS2D/StringUtils.h
+++ b/NAS2D/StringUtils.h
@@ -16,6 +16,7 @@
 #include <utility>
 #include <vector>
 #include <stdexcept>
+#include <type_traits>
 
 
 namespace NAS2D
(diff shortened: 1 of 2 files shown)


NAS2D/StringUtils.h
```diff
@@ -25,6 +26,8 @@ namespace NAS2D
 	{
 		if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, const char*>) {
 			return value;
+		} else if constexpr (std::is_convertible<T, std::string>::value) {
+			return value;
```
DanRStevens, line 30: As a bit of an aside, I think the `std::is_convertible` case would also cover the `const char*` case above. It might even cover the `std::string` case too, now that I think about it. We could potentially simplify. And the existing unit tests should let us know right away if it would work.

```cpp
if constexpr (std::is_convertible<T, std::string>::value) {
	return value;
} ...
```

DanRStevens (approved):
Thank you for back porting this change.

PR #824, opened 2021-01-20
Simplify stringFrom by collapsing cases

DanRStevens:
The `std::is_convertible` check should handle all of `std::string`, `const char*`, and objects with an `operator std::string() const` method.


DanRStevens:
@Brett208: You might want to check this one.

Brett208:
Cool, I hadn't thought about this being possible.

Issue #826, opened 2021-02-08
Update compilers for C++20 support


DanRStevens:
Would be nice to start using c++20 features. This of course would require the CI builds to be using compilers with C++20 support.

A quick test with enabling C++20 results in successful builds for:
- `build-linux-gcc`  (Ubuntu 20.04)
- `build-linux-clang`  (Ubuntu 20.04)

Enabling C++20 results in build failures for:
- `build-linux-mingw`  (Ubuntu 20.04)
- `build-linux`  (Ubuntu 18.04)
- `build-macos`

> error: unrecognized command line option '-std=c++20'

I did not test updating MSVC to enable C++20 support.

----

We can drop the build for Ubuntu 18.04. That image is getting a bit old now. It's probably not reasonable to expect C++20 features on Ubuntu 18.04.

We may need to look into updating the image for Mingw. It is based on GCC, so if GCC has been updated to support C++20, there will probably be corresponding support in an updated Mingw.

On MacOS, the C++ compiler shipped with Xcode is based on Clang, though tends to lag significantly behind the latest Clang release. On CircleCI we are currently using the MacOS image with Xcode `12.3.0`, which apparently is based on Clang `12.0.0`. It seems the `c++20` option should have been added in Clang 10, so not quite sure what the problem is here.

DanRStevens:
It's seems the `c++2a` standard (pre-c++20 official release) is available for most of the compilers. Not sure we want to target a standard pre-release though. Probably better to just wait until there is widespread compiler support for the new `c++20` standard before upgrading. That might  mean revisiting this in about a year.

cugone:
VS 16.10+ (VS2019) fully supports C++20 when the  `/std:c++latest` language switch is supplied. Future versions (VS2022) will support a C++20 switch `/std:c++20` explicitly. See: https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-160

ldicker83:
Was looking into VS2022, looks pretty promising. Looking forward to upgrading!

The big drag here is probably going to be macOS X support. As always Apple will probably drag their feet unnecessarily long.

cugone:
> Was looking into VS2022, looks pretty promising. Looking forward to upgrading!
> 
> The big drag here is probably going to be macOS X support. As always Apple will probably drag their feet unnecessarily long.

[...]

PR #828, opened 2021-02-08
Update C++ standard target

DanRStevens:
Reference: #826

This PR gives a snapshot of support for C++20 using the currently configured CI compilers. Currently `c++20` fails for MacOS, Mingw-w64, and GCC on Ubuntu 18.04. If we use the earlier `c++2a` target, from before the C++20 standard became official, than some of the older compilers can manage to build successfully. We'll need to drop support for Ubuntu 18.04 though.

Perhaps someone would like to add the corresponding update for Windows.


DanRStevens:
Closing. Will wait until the full `c++20` standard becomes more widespread before upgrading.

PR #829, opened 2021-02-08
Fix undefined behaviour in `split`

### LINKED ISSUE #825: NAS2D::split() throws unhandled exception when passed an empty string
I suspect that most use cases here would not pass empty strings through `split()` but there is a case in OutpostHD that could pass an empty string when loading a `RobotCommandCenter` that's been saved before any robots can be assigned to it.

I'm not sure if this should be considered an error and throw our own `std::runtime_error` or if it should be silently ignored and just pass back an empty `vector<string>`.

In the mean time I'll use a work around in the OutpostHD code that doesn't attempt to load an empty robot list string.

As a note, adding:

```cpp
if (str.empty()) { return std::vector<std::string>(); }
```

to the top of the function eliminates the throw. It's the silent swallowing of potentially incorrect use that has me unsure of the best way to approach this.

DanRStevens:
We should not be calling `std::string::back` on an empty string.

Closes #825

### DIFF
--- a/NAS2D/StringUtils.cpp
+++ b/NAS2D/StringUtils.cpp
@@ -56,7 +56,7 @@ std::vector<std::string> split(const std::string& str, char delim /*= ','*/)
 	{
 		result.push_back(curString);
 	}
-	if (ss.eof() && str.back() == delim)
+	if (ss.eof() && !str.empty() && str.back() == delim)
 	{
 		result.push_back(std::string{});
 	}


DanRStevens:
@ldicker83: See if this fixes the debug assertion you were getting.

May be able to avoid: https://github.com/OutpostUniverse/OPHD/pull/762

ldicker83:
Definitely fixes the debug assertion. Simpler change than I suggested too. :D Gonna go ahead and merge this change set.

PR #837, opened 2021-03-13
Convenience function to ease retrieving text from an XmlElement

Brett208:
Converted function GetText from TinyXml 2.4.3 to work in NAS2D

Looking to push planet attributes into an xml file in OPHD. This function will improve readability of effort in OPHD.

### DIFF
--- a/NAS2D/Xml/XmlElement.cpp
+++ b/NAS2D/Xml/XmlElement.cpp
@@ -246,7 +246,7 @@ XmlNode* XmlElement::clone() const
 	return clone;
 }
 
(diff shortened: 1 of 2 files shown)


NAS2D/Xml/XmlElement.cpp
```diff
@@ -246,6 +246,20 @@ XmlNode* XmlElement::clone() const
 	return clone;
 }
 
+std::string XmlElement::GetText() const
```
DanRStevens, line 249: I believe the naming convention in this project is lowerCamelCase for function names.

PR #838, opened 2021-03-18
Xml serializer

Brett208:
Port code from OPHD for reading xml files

Allows reusing code when loading Sprite metadata


Brett208:
If this changeset looks good, will want to pull code from OPHD and redirect calls to open xml documents in OPHD

DanRStevens (commented):
So this is probably fine, though I do have some concerns.

Do we want this in NAS2D? There's been a push recently to get XML code out of NAS2D. If we put this in, then it becomes another feature that downstream projects will depend on, and that we may want to remove in the future.

As a lesser concern, is there anything more we can do to improve the usability of `stringToEnum`? Some partial ideas were presented in the original thread in OPHD, though it wasn't a fully worked out idea. It might be easier to experiment with the feature while it's still in the same project that's using it.

And as a bit of a side note, this may obsolete the local branch I had where I was experimenting with updates to this code. Instead of returning an `XmlDocument`, I had instead tried to take a lambda to call with the root `XmlElement`. That way processing can be done while `XmlDocument` is still in scope, without having to return it. The result of the function would the result of the lambda.

Brett208:
I am trying to centralize some of the xml calls and move them out of OPHD as a first step to pulling. I had noticed the way the sprite data is loaded and initially error checked was exactly the same as how saved games were loaded and error checked in OPHD. It is a small effort and I don't mind pausing or abandoning the effort if others are not keen on it. 

I guess the stringToEnum code just came along since it was in the same module. It could be left in OPHD for sure if it made this PR more palatable.

DanRStevens:
I had also noticed the duplication of code for opening and validating XML files. That is something I want to address.

I'm uncertain where I want to go with this. (And probably too tired to make a decision right now). Maybe Leeor can weigh in on the ideas.

DanRStevens:
Hmm, I was expecting a merge conflict after merging #839, though that doesn't seem to be the case.

ldicker83:
A lot of the code is lifted directly out of NAS2D's sprite loading/validation code. Much of it is 'quick and dirty' 'make it work' kind of deals which is why it's error prone, messy and inefficient. Lots of copy paste.

I am fine with any attempts at making this a less catastrophic mess. I still think that XML should be pulled from the interface of the NAS2D library -- if it's used internally to handle stuff that's an implementation detail that doesn't need to be exposed.

OutpostHD's use of XML can just be shifted over to using any XML library, or, really, any serialization mechanism. XML was chosen primarily to make it easy during early development to make sure I was writing values properly and to easily adjust values outside of the game. As the save game format has effectively matured the use of XML can be retired and replaced with a different method of serialization, even binary though I think a human-readable form is preferable. A bit out of the scope of this task but I really do want to move away from 'rolling my own' kind of thinking to using mature well-tested libraries that are easy to install using package managers (like homebrew, apt get, vcpkg, etc.)

DanRStevens:
Err, I don't know if we wanted to merge both #839 and #838, as they both have similar aims. @Brett208, can you comment on this? I've kind of lost context on this one, and I might not have time to delve into the details for a little while.

PR #883, opened 2021-06-24
Adds Dictionary class files to appropriate MSVC project filters

cugone:
The dictionary class files had its files applied to the base project filter tree instead of Source and Header. This happens when you try and edit the .vcxproj file directly to add files to a project instead of using the IDE.


DanRStevens:
Thank you

That happens all too frequently when I add source files from Linux.

cugone:
Have you tried Visual Studio Code? It's getting a lot of praise among Linux developers and it integrates with Visual Studio projects quite well.

DanRStevens:
I've dabbled with it. It seems similar to Atom, though so far I've preferred using Atom. Maybe it's time to revisit.

Issue #861, opened 2021-05-01
Improve documentation


DanRStevens:
We should try to improve the API documentation. Currently we have a lot of "low value" comments, which I've sometimes taken to just removing as they only add clutter, without telling people anything new. Examples would be comments that simply restate what's already implied by a class or variable name.

In some cases, we should improve class or variable names so there is no need to document what they are for. Ideally a variable's use should be self evident from it's name, and so not require documentation.

For functions and classes, we should try to give examples of usage whenever possible. To some extent, the unit test code serves as example usage, though that's not an obvious place for people to look. Actually, I wonder if there might be a way for Doxygen to pick up certain blocks of unit test code and display it as part of the documentation. That might be a good way of having up-to-date documentation that is guaranteed to compile and function as expected. Since the unit tests are regularly run as part of CI, the code there should always be up-to-date.

DanRStevens:
Looks like all the low value comments have largely been removed.

Looks like any variables that needed comments have been renamed to not need a comment.

We could possible use more examples of usage in some places, though perhaps we should rely on the unit tests for that. I'm not sure anyone is actually using the auto generated documentation, so maybe not something to put too much effort into at present.

With that said, I think we can close this issue.

PR #884, opened 2021-06-25
Adds VS-generated .editorconfig file

cugone:
Ref: #861

This adds a `.editorconfig` file to the root to allow for generated Doxygen comments using `/**` or `Ctrl+/` above a function to generate doxygen comments.

Visual Studio Intellisense can interpret this to generate correct tooltips without any extraneous character symbols.

See: https://devblogs.microsoft.com/cppblog/doxygen-and-xml-doc-comment-support/

The settings generated were based on local settings, to change them if any are inappropriate see: https://docs.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019
for valid values.

This will also make a future PR addressing the documentation much easier.


.editorconfig
```diff
@@ -0,0 +1,69 @@
...
+cpp_new_line_scope_braces_on_separate_lines = false
+cpp_new_line_close_brace_same_line_empty_type = false
+cpp_new_line_close_brace_same_line_empty_function = false
+cpp_new_line_before_catch = false
+cpp_new_line_before_else = false
+cpp_new_line_before_while_in_do_while = false
+cpp_space_before_function_open_parenthesis = remove
+cpp_space_within_parameter_list_parentheses = false
+cpp_space_between_empty_parameter_list_parentheses = false
+cpp_space_after_keywords_in_control_flow_statements = false
```
DanRStevens, line 39: I suspect we want this to be true. Assuming this represents a style such as the following, where a space follows `if`:
```cpp
if (boolExpression)
```

.editorconfig
```diff
@@ -0,0 +1,69 @@
...
+cpp_space_within_control_flow_statement_parentheses = false
+cpp_space_before_lambda_open_parenthesis = false
+cpp_space_within_cast_parentheses = false
+cpp_space_after_cast_close_parenthesis = false
+cpp_space_within_expression_parentheses = false
+cpp_space_before_block_open_brace = true
+cpp_space_between_empty_braces = false
+cpp_space_before_initializer_list_open_brace = false
+cpp_space_within_initializer_list_braces = false
```
DanRStevens, line 48: I like this. I think there's been a bit of conflict over this, largely due to default Visual Studio settings, rather than programmer intent.

.editorconfig
```diff
@@ -0,0 +1,69 @@
...
+cpp_space_before_open_square_bracket = false
+cpp_space_within_square_brackets = false
+cpp_space_before_empty_square_brackets = false
+cpp_space_between_empty_square_brackets = false
+cpp_space_group_square_brackets = true
+cpp_space_within_lambda_brackets = false
+cpp_space_between_empty_lambda_brackets = false
+cpp_space_before_comma = false
+cpp_space_after_comma = true
+cpp_space_remove_around_member_operators = true
```
DanRStevens, line 59: I was wondering about this one. I think in some cases we put spaces around operators when defining them, sometimes not.

.editorconfig
```diff
@@ -0,0 +1,69 @@
...
+cpp_space_before_inheritance_colon = true
+cpp_space_before_constructor_colon = true
+cpp_space_remove_before_semicolon = true
+cpp_space_after_semicolon = true
```
DanRStevens, line 63: This one seems odd. Do them mean when embedded in a line, such as `for` loops?
```cpp
for (int i =0; i < 10; ++i)
```

DanRStevens (commented):
This is a very good idea to have scripted style checks and code formatting.

I notice there is support for both EditorConfig and clangformat. I'm wondering if clangformat support might be more widespread. I remember there was an attempt at using clangformat in the past, though the config file didn't match existing project conventions very well. It was removed until update, which sadly never happened. I'd thought about revisiting that from time to time.

Would it be possible to create a throw-away branch, with these rules applied for auto formatting. It might help settle on some rules if we could see what abiding by them would do to the code base. I suspect Leeor would want to have something to take a look at.

cugone:
I suspect the only required settings are

```
# Visual Studio generated .editorconfig file with C++ settings.
root = true

[*.{c++,cc,cpp,cppm,cxx,h,h++,hh,hpp,hxx,inl,ipp,ixx,tlh,tli}]

# Visual C++ Code Style settings

cpp_generate_documentation_comments = doxygen_slash_star

```
and the formatting settings after these lines can probably be removed in favor of a `.clang-format` file.

`.clang-format` support has increased in VS and all the troublesome settings that did not exist before, now do.

.editorconfig
cugone: I think they mean in all cases, such as `for` loops, and:
`if-init-expressions`:
```cpp
if(int i = CalculateValue(); i > 5) {
//...
}
```
Empty `for` loops (alternative infinite while-loop generation that does not throw a warning for "Expression is always true"):

```cpp
for(;;) {
//...
}
```
Multi-expression "short" const-preserving-lambda initializers:
```cpp
const auto facing_end = [=]()->Vector2 { auto end = Vector2::X_AXIS; end.SetLengthAndHeadingDegrees(orientation, cosmetic_radius); return center + end; }();
const auto velocity_end = [=]()->Vector2 { auto end = entity->GetVelocity().GetNormalize(); end.SetLengthAndHeadingDegrees(end.CalcHeadingDegrees(), cosmetic_radius); return center + end; }();
const auto acceleration_end = [=]()->Vector2 { auto end = entity->GetAcceleration().GetNormalize(); end.SetLengthAndHeadingDegrees(end.CalcHeadingDegrees(), cosmetic_radius); return center + end; }();
```

All of which benefit from generous whitespace. (The infinite-for-loop is debatable...)

cugone:
Did a little more reading in the EditorConfig article and this stood out (emphasis mine):

>The Visual Studio C++ formatter has a rich set of configurable settings that can be applied globally. To set C++ formatting settings for a specific workspace, **use clangformat or EditorConfig**. Both Visual Studio and Visual Studio Code have built-in EditorConfig support for each of the global Visual Studio C++ formatting settings, **with the EditorConfig settings taking precedence**. This means you can add EditorConfig files to your workspace to configure C++ formatting on a more granular level and enforce consistent code style for everyone contributing to the project.

Which I interpret as:
- .clang-format or .editorconfig formatting. Not both. (The Doxygen style declarations not-withstanding)
- If `.editorconfig` is desired, everyone on the project should use Visual Studio or Visual Studio Code in order to take advantage of the formatting enforcement.
- Otherwise, use `.clang-format` for cross-platform development environments.

ldicker83:
This would be really helpful for IDE tooltips. May integrate this with OPHD as well.

DanRStevens:
Based on the later removal of most formatting rules, I assume you're working on a `clangformat` rule set?

As for the documentation, I take what they wrote to mean you can use both, though if rules are ever in conflict, the `.editorconfig` rules take precedence over the `clangformat` rules.

We definitely don't want conflicting rule sets though. That would be confusing, both for tools that apply auto formatting and might fight with each other, and for programmers who are left wondering which set of rules to follow.

Anyway, I'm not opposed to starting out with the `.editorconfig` rule set. I suppose you might call it a mild preference for whatever format is less likely to result in lock-in with a proprietary or non-open source solution.

DanRStevens:
So I was doing a bit of reading:
https://editorconfig.org/

I may have had some misunderstandings about what or who are behind EditorConfig. Seems to be it's own thing, and is much more widespread than I had expected. It's just that I had never heard of it, and given the context of either Visual Studio or Visual Studio Code, I assumed it was a new Microsoft thing.

I briefly tried it with Visual Studio code, though I've since found there is an Atom plugin for it too. I'm giving it a try now with the set of rules from the first commit. So far I've managed for format an individual file. No idea how to run it on the entire project though.

cugone:
>
> Based on the later removal of most formatting rules, I assume you're working on a clangformat rule set?

No, I just didn't want a conflict.

Funnily enough, there's already a clangformat file in the repo, I think it's just disabled from a Visual Studio perspective. Might want to look over it to see if there's anything you want to change. One thing I did notice was that the `Standard` was set to `Cpp11` instead of `Cpp17`; that may have to do with either: a) at the time it was written, support in visual studio didn't exist; or, MacOS's lack of support. I don't remember.

> 
> I may have had some misunderstandings about what or who are behind EditorConfig. Seems to be it's own thing, and is much more widespread than I had expected. It's just that I had never heard of it, and given the context of either Visual Studio or Visual Studio Code, I assumed it was a new Microsoft thing.

I thought it was a Microsoft thing, too. Nope. They just added native support; albeit, personally, the Doxygen documentation generation is about the only useful thing about it. I used to use [Atomineer Utils](https://www.atomineerutils.com/), but since Doxygen is natively supported now I might switch.

DanRStevens:
This seemed particularly relevant (1st answer, quote):
https://stackoverflow.com/questions/39343031/editorconfig-how-to-autofix-all-files-in-a-project

> To me, EditorConfig is all about – would you believe it – configuring your editor and nothing more. I don't think it should be responsible for linting or fixing, as there are language-specific tools to accomplish those goals (e.g., ESLint for JavaScript). The only way you can reliably lint or fix indentation is to be aware of the language in question and EditorConfig is language agnostic.

That would seem to suggest `clang-format` may be a better tool for source code formatting, as I believe it has language specific rules built in. Though in terms of editor configuration, such as spaces versus tabs, and tab size, that seems very appropriate for an EditorConfig file.

Anyway, I installed the `eclint` tool (`npm install -g eclint`), and gave it a try to see if I could apply EditorConfig format changes to the whole project. It didn't seem to do anything. Perhaps not surprising, considering the file was originally meant to configure editors, and not to apply code formatting. Actually, looking back at the `.editorconfig` file, I see the format rules are commented as "# Visual C++ Formatting settings". I suppose that makes sense in light of the above quote.

Now that you've reminded me the `.clang-format` file is still there, I'm kind of tempted to play a bit.

DanRStevens:
Reference: #893

I opened a new issue so we can track changes related to code formatting in a single place.

PR #899, opened 2021-06-28
First pass on the documentation

cugone:
Ref: #861 

I apologize in advance. Due to a typo in the copyright header, or the file missing it entirely, this PR touches *every* file and is intended as a first-pass only. Future documentation PRs will be more focused.

I've added a few TODOs to consider that would create self-documenting code in lieu of comments.

"Low-quality" comments were removed. These include:
- Comments that add no benefit such as pontifications.
- Dead code
- Comments that restate information that can be inferred by simply reading the code.

### DIFF
--- a/NAS2D/Configuration.cpp
+++ b/NAS2D/Configuration.cpp
@@ -5,9 +5,8 @@
 // = NAS2D is distributed under the terms of the zlib license. You are free to copy,
 // = modify and distribute the software under the terms of the zlib license.
 // =
-// = Acknowledgement of your use of NAS2D is appriciated but is not required.
+// = Acknowledgment of your use of NAS2D is appreciated but is not required.
 // ==================================================================================
-
 #include "Configuration.h"
 #include "ContainerUtils.h"
 #include "ParserHelper.h"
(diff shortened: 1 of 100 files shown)


DanRStevens:
Oh man, I think the PR I merged just a moment ago is going to have quite a few merge conflicts with this.

DanRStevens:
I was looking back at this, wondering what to do about the merge conflicts. Doesn't look like it'll be easy to fix.

I would really like to merge the header typo fixes. Is there a way we could extract that part and PR it?

@ldicker83, what's your preference for adding copyright headers to each file? I know I haven't added them to new files that I created. I kind of don't care, and in a way, kind of don't want to clutter up the files. Though I've noticed many projects prefix each file, rather than have a generic licence file for the whole project. I've never been too clear on the reason for this. I suspect it's for when people take just a single file from a project, though that seems pretty much the same as removing the copyright header notice from a file, so I don't know if it really changes anything from a legal perspective.

As for comment removals, I agree most comments can probably be removed. Many of them just restate the code. Though I felt like this PR maybe removed a few too many comments. There were a few explanatory comments that were removed here, which may have contained reasons that were difficult to infer from the code. In some cases I haven't been too sure where to draw the line, so have tended to leave comments which maybe should be removed.

cugone:
I'm splitting this into 2 PRs.

Ref: #921 and #922

Issue #701, opened 2020-07-20
Extract Window class from Renderer


DanRStevens:
The `Renderer` methods largely fall into two categories:
- Drawing methods
- Windows management methods

We should split the Window management methods into their own `Window` class.

The `Window` class should control things like the `icon` and the `title`, as well as the control area (`Rectangle`). It should also provide the event handlers for the `onResize` event, the `minimumSize` property, the `resizeable` property, and the `fullscreen` property.

Potentially this class could allow access to the underlying `HWND` handle on Windows, which could be used as an escape for calling native Windows API methods. In particular, it could be used for modal popup boxes (which we should generally discourage using).

cugone:
This might be more involved than previously realized. A basic Rendering Hardware Interface (RHI) setup for refactoring out the instance, device, context, and output might be required. Especially in differences between DX/OGL and ownership of particular instances. Modern Metal/Vulkan/DX12 minimizes the differences due to similar support for various features but that's way beyond the scope of the engine.

DanRStevens:
Yes, I expect this one to be quite involved. Though I believe it's important work. I'm not entirely sure of what the final design would look like yet. There are some obvious awkward points in terms of underlying window handles and rendering contexts.

I've been tempted to keep the same methods initially, but to somehow split them into a `Window` class and a `Renderer` class, and perhaps give `friend` access between them to minimize external API burdens. I'm sure a better design could be evolved from there.

DanRStevens:
I guess technically the points of this issue have been met. There's probably more room for improvement, perhaps decoupling the new `Window` class from `Renderer` so there is no longer inheritance between them, but perhaps that can be done under a separate issue when the time comes.

PR #903, opened 2021-07-03
Window interface class

cugone:
Ref: #701

This is the first step in refactoring out a self-contained Window class.

I've added comments where I think there would be confusion. Otherwise, I think it's self-explanatory.

I have reservations about using `std::pair<int,int>` instead of an explicit `IntVector`-like type, but, with the advent of structured bindings and initializer lists:

```cpp
window.Dimensions({800,600});
auto& [x,y] = window.GetClientDimensions();
```

The code ends up looking exactly the same from a user standpoint, so I guess I can live with it.

### DIFF
--- a/NAS2D/Renderer/Window.h
+++ b/NAS2D/Renderer/Window.h
@@ -3,103 +3,108 @@
 #include <string>
 #include <utility>
 
-enum class WindowDisplayMode
+namespace NAS2D
 {
-	Windowed
-	,Borderless
-	,Borderless_Fullscreen
-};
 
-struct WindowDesc
-{
-	//The client dimensions
-	//Window creation process should automatically add space for borders, menus and title bars
-	std::pair<int, int> dimensions{1600,900};
-	std::pair<int, int> position{0,0};
-	WindowDisplayMode mode{WindowDisplayMode::Windowed};
-};
+	enum class WindowDisplayMode
+	{
+		Windowed,
+		Borderless,
+		Borderless_Fullscreen,
+	};
 
-class Window
-{
-public:
-	virtual ~Window() noexcept {}
+	struct WindowDesc
+	{
+		//The client dimensions
+		//Window creation process should automatically add space for borders, menus and title bars
+		std::pair<int, int> dimensions{1600,900};
+		std::pair<int, int> position{0,0};
+		WindowDisplayMode mode{WindowDisplayMode::Windowed};
+	};
 
-	virtual void open() noexcept = 0;
-	virtual void close() noexcept = 0;
+	class Window
+	{
+	public:
+		virtual ~Window() noexcept {}
 
-	virtual void show() noexcept = 0;
-	virtual void hide() noexcept = 0;
-	virtual void unhide() noexcept = 0;
+		virtual void open() noexcept = 0;
+		virtual void close() noexcept = 0;
 
-	[[nodiscard]] virtual bool isOpen() const noexcept = 0;
-	[[nodiscard]] virtual bool isClosed() const noexcept = 0;
-	[[nodiscard]] virtual bool isFullscreen() const noexcept = 0;
-	[[nodiscard]] virtual bool isWindowed() const noexcept = 0;
+		virtual void show() noexcept = 0;
+		virtual void hide() noexcept = 0;
+		virtual void unhide() noexcept = 0;
 
-	/**
-	 * @brief Get the underlying OS-specific window handle.
-	 * @return void* Handle to the window. static_cast to appropriate OS-specific type. For Windows this would be HWND.
-	*/
-	[[nodiscard]] virtual void* nativeHandle() const noexcept = 0;
+		[[nodiscard]] virtual bool isOpen() const noexcept = 0;
+		[[nodiscard]] virtual bool isClosed() const noexcept = 0;
+		[[nodiscard]] virtual bool isFullscreen() const noexcept = 0;
+		[[nodiscard]] virtual bool isWindowed() const noexcept = 0;
 
-	/**
-	 * @brief Set the underlying OS-specific window handle.
-	 * @param handle: The handle to the window. For Windows this would be a variable of type HWND.
-	 * @return None
-	*/
-	virtual void nativeHandle(void* handle) noexcept = 0;
+		/**
+		 * @brief Get the underlying OS-specific window handle.
+		 * @return void* Handle to the window. static_cast to appropriate OS-specific type. For Windows this would be HWND.
+		*/
+		[[nodiscard]] virtual void* nativeHandle() const noexcept = 0;
 
-	/**
-	 * @brief Get the underlying OS-specific window-owned device context.
-	 * @return void* handle to the Device Context. static_cast to appropriate OS-specific type. For Windows this would be HDC.
-	*/
-	[[nodiscard]] virtual void* nativeDeviceContext() const noexcept = 0;
+		/**
+		 * @brief Set the underlying OS-specific window handle.
+		 * @param handle: The handle to the window. For Windows this would be a variable of type HWND.
+		 * @return None
+		*/
+		virtual void nativeHandle(void* handle) noexcept = 0;
 
-	/**
-	 * @brief Set the underlying OS-specific window-owned device context.
-	 * @param hDC: The handle to the device context this window will now own.
-	 * @comment For Windows this would be a variable of type HDC.
-	 * @return None
-	*/
-	virtual void nativeDeviceContext(void* hDC) noexcept = 0;
+		/**
+		 * @brief Get the underlying OS-specific window-owned device context.
+		 * @return void* handle to the Device Context. static_cast to appropriate OS-specific type. For Windows this would be HDC.
+		*/
+		[[nodiscard]] virtual void* nativeDeviceContext() const noexcept = 0;
 
-	/**
-	 * @brief Dimensions of the entire window including any border, menus, or title bars.
-	 * @return std::pair<int, int> x/y components of the window dimensions
-	*/
-	[[nodiscard]] virtual std::pair<int, int> dimensions() const noexcept = 0;
+		/**
+		 * @brief Set the underlying OS-specific window-owned device context.
+		 * @param hDC: The handle to the device context this window will now own.
+		 * @comment For Windows this would be a variable of type HDC.
+		 * @return None
+		*/
+		virtual void nativeDeviceContext(void* hDC) noexcept = 0;
 
-	/**
-	 * @brief Dimensions of the renderable area of the window not including any borders, menus, or title bars.
-	 * @return std::pair<int, int> x/y components of the client area dimensions
-	*/
-	[[nodiscard]] virtual std::pair<int, int> clientDimensions() const noexcept = 0;
-	virtual void dimensions(std::pair<int, int> newClientDimensions) noexcept = 0;
+		/**
+		 * @brief Dimensions of the entire window including any border, menus, or title bars.
+		 * @return std::pair<int, int> x/y components of the window dimensions
+		*/
+		[[nodiscard]] virtual std::pair<int, int> dimensions() const noexcept = 0;
 
-	virtual void title(std::string newTitle) noexcept = 0;
-	[[nodiscard]] virtual std::string title() noexcept = 0;
+		/**
+		 * @brief Dimensions of the renderable area of the window not including any borders, menus, or title bars.
+		 * @return std::pair<int, int> x/y components of the client area dimensions
+		*/
+		[[nodiscard]] virtual std::pair<int, int> clientDimensions() const noexcept = 0;
+		virtual void dimensions(std::pair<int, int> newClientDimensions) noexcept = 0;
 
-	virtual void displayMode(WindowDisplayMode newDisplayMode) noexcept = 0;
-	[[nodiscard]] virtual WindowDisplayMode displayMode() noexcept = 0;
+		virtual void title(std::string newTitle) noexcept = 0;
+		[[nodiscard]] virtual std::string title() noexcept = 0;
 
-	[[nodiscard]] virtual std::pair<int, int> position() const noexcept = 0;
-	virtual void position(std::pair<int, int> newPosition) noexcept = 0;
+		virtual void displayMode(WindowDisplayMode newDisplayMode) noexcept = 0;
+		[[nodiscard]] virtual WindowDisplayMode displayMode() noexcept = 0;
 
-	virtual void setDimensionsAndPosition(std::pair<int, int> newClientDimensions, std::pair<int, int> newPosition) noexcept = 0;
+		[[nodiscard]] virtual std::pair<int, int> position() const noexcept = 0;
+		virtual void position(std::pair<int, int> newPosition) noexcept = 0;
 
-	/**
-	 * @brief Set the native OS-specific window icon resource.
-	 * @param iconResource: Handle to the icon resource.
-	 * @return 
-	*/
-	virtual void icon(void* iconResource) noexcept = 0;
+		virtual void setDimensionsAndPosition(std::pair<int, int> newClientDimensions, std::pair<int, int> newPosition) noexcept = 0;
 
-	/**
-	 * @brief Get the native OS-specific window icon resource.
-	 * @return void* handle to the native OS-specific window icon resource.
-	*/
-	[[nodiscard]] virtual void* icon() const noexcept = 0;
-protected:
-private:
-	
-};
+		/**
+		 * @brief Set the native OS-specific window icon resource.
+		 * @param iconResource: Handle to the icon resource.
+		 * @return
+		*/
+		virtual void icon(void* iconResource) noexcept = 0;
+
+		/**
+		 * @brief Get the native OS-specific window icon resource.
+		 * @return void* handle to the native OS-specific window icon resource.
+		*/
+		[[nodiscard]] virtual void* icon() const noexcept = 0;
+	protected:
+	private:
+
+	};
+
+}
--- a/NAS2D/Renderer/Window.h
+++ b/NAS2D/Renderer/Window.h
@@ -85,6 +85,18 @@ public:
 
 	virtual void SetDimensionsAndPosition(std::pair<int, int> newClientDimensions, std::pair<int, int> newPosition) noexcept = 0;
 
+	/**
+	 * @brief Set the native OS-specific window icon resource.
+	 * @param iconResource: Handle to the icon resource.
+	 * @return 
+	*/
+	virtual void Icon(void* iconResource) noexcept = 0;
+
+	/**
+	 * @brief Get the native OS-specific window icon resource.
+	 * @return void* handle to the native OS-specific window icon resource.
+	*/
+	[[nodiscard]] virtual void* Icon() const noexcept = 0;
 protected:
 private:
 	
--- a/NAS2D/Renderer/Window.h
+++ b/NAS2D/Renderer/Window.h
@@ -14,7 +14,7 @@ struct WindowDesc
 {
 	//The client dimensions
 	//Window creation process should automatically add space for borders, menus and title bars
-	std::pair<int, int> dimensions{1600,600};
+	std::pair<int, int> dimensions{1600,900};
 	std::pair<int, int> position{0,0};
 	WindowDisplayMode mode{WindowDisplayMode::Windowed};
 };
--- a/NAS2D/Renderer/Window.h
+++ b/NAS2D/Renderer/Window.h
@@ -3,13 +3,15 @@
 #include <string>
 #include <utility>
 
-enum class WindowDisplayMode {
+enum class WindowDisplayMode
+{
 	Windowed
 	,Borderless
 	,Borderless_Fullscreen
 };
 
-struct WindowDesc {
+struct WindowDesc
+{
 	//The client dimensions
 	//Window creation process should automatically add space for borders, menus and title bars
 	std::pair<int, int> dimensions{1600,600};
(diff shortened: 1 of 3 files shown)


ldicker83 (changes requested):
Only comment is stylistic choice to maintain consistency with the rest of the project. Method names in classes start with a lower case.

I'm not going to whine too much about brace placement, just keep in mind that for classes and structs opening braces are generally on a new line.

Otherwise this generally looks good. With the small changes recommended above this should be ready to merge.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
+#pragma once
+
+#include <string>
+#include <utility>
+
+enum class WindowDisplayMode {
+	Windowed
+	,Borderless
+	,Borderless_Fullscreen
+};
```
DanRStevens: We use line trailing commas. Optional terminating commas (after the last list entry) are often included, particularly when list items are spilled onto multiple lines. We tend to avoid `_` in our more recent naming conventions (though there may still be examples of older uses that haven't yet renamed).

```cpp
enum class WindowDisplayMode
{
	Windowed,
	Borderless,
	BorderlessFullscreen,
};
```

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+
+struct WindowDesc {
+	//The client dimensions
+	//Window creation process should automatically add space for borders, menus and title bars
+	std::pair<int, int> dimensions{1600,600};
+	std::pair<int, int> position{0,0};
+	WindowDisplayMode mode{WindowDisplayMode::Windowed};
+};
```
DanRStevens: It would be more consistent to use `Point` and `Vector` here. The namespace prefix can be omitted if the header contents is wrapped in `namespace NAS2D`. Also, spaces after commas in data lists.
```cpp
struct WindowDesc
{
	Vector<int> dimensions{1600, 900};
	Point<int> position{0, 0};
	WindowDisplayMode mode{WindowDisplayMode::Windowed};
};
```

Might also make sense to move the comment to a function where that's more relevant. Data structures tend to be pretty dumb about details like that. It might be a class constructor or size method that would handle any sort of size adjustment, or documentation for a user about how they should specify parameters.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+	 * @return void* Handle to the window. static_cast to appropriate OS-specific type. For Windows this would be HWND.
+	*/
+	[[nodiscard]] virtual void* NativeHandle() const noexcept = 0;
+
+	/**
+	 * @brief Set the underlying OS-specific window handle.
+	 * @param handle: The handle to the window. For Windows this would be a variable of type HWND.
+	 * @return None
+	*/
+	virtual void NativeHandle(void* handle) noexcept = 0;
```
DanRStevens: I feel like this should perhaps be split off into a subclass, such as `SdlWindow`. That detail gets a bit specific about how windows are implemented. I understand the importance of having accessor methods like that. Not really sure where the corresponding set method would be used though. I feel like how native window handles are set is normally a private internal detail.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+	*/
+	[[nodiscard]] virtual void* NativeDeviceContext() const noexcept = 0;
+
+	/**
+	 * @brief Set the underlying OS-specific window-owned device context.
+	 * @param hDC: The handle to the device context this window will now own.
+	 * @comment For Windows this would be a variable of type HDC.
+	 * @return None
+	*/
+	virtual void NativeDeviceContext(void* hDC) noexcept = 0;
```
DanRStevens: I have similar concerns about this as the native window handle accessor.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+	 * @brief Dimensions of the entire window including any border, menus, or title bars.
+	 * @return std::pair<int, int> x/y components of the window dimensions
+	*/
+	[[nodiscard]] virtual std::pair<int, int> GetDimensions() const noexcept = 0;
+	
+	/**
+	 * @brief Dimensions of the renderable area of the window not including any borders, menus, or title bars.
+	 * @return std::pair<int, int> x/y components of the client area dimensions
+	*/
+	[[nodiscard]] virtual std::pair<int, int> GetClientDimensions() const noexcept = 0;
```
DanRStevens: Should use a `Vector<int>` for the return type.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+	virtual void Title(std::string title) noexcept = 0;
+	[[nodiscard]] virtual std::string Title() noexcept = 0;
+
+	virtual void DisplayMode(WindowDisplayMode mode) noexcept = 0;
+	[[nodiscard]] virtual WindowDisplayMode DisplayMode() noexcept = 0;
+
+	[[nodiscard]] virtual std::pair<int, int> Position() const noexcept = 0;
+	virtual void Position(std::pair<int, int> newPosition) noexcept = 0;
+
+	virtual void SetDimensionsAndPosition(std::pair<int, int> newClientDimensions, std::pair<int, int> newPosition) noexcept = 0;
```
DanRStevens: Maybe `setArea(Rectangle<int> newArea)`? Currently the point/vector parameters are in reverse order from how the rest of the project would typically pass area coordinates.

NAS2D/Renderer/Window.h
```diff
@@ -0,0 +1,91 @@
...
+
+protected:
+private:
+
```
DanRStevens, line 104: I would probably just omit access specifiers if they aren't being used.

NAS2D/Renderer/Window.h
```diff
@@ -85,6 +85,18 @@ class Window {
...
+	 * @param iconResource: Handle to the icon resource.
+	 * @return 
+	*/
+	virtual void Icon(void* iconResource) noexcept = 0;
+
+	/**
+	 * @brief Get the native OS-specific window icon resource.
+	 * @return void* handle to the native OS-specific window icon resource.
+	*/
+	[[nodiscard]] virtual void* Icon() const noexcept = 0;
```
DanRStevens: We might want to use `Image` references, rather than `void*`. At least that would make sense for set. Implementing get might be more awkward.

DanRStevens (commented):
You probably want to wrap the contents of that header in a `namespace NAS2D`, like other header files.

Had a few thoughts about some design aspects, though I'm afraid I don't have a suggestion for improvement with all of them. Maybe something to address in future work.

NAS2D/Renderer/Window.h
cugone: These don't actually use images. For Windows it can either be a reference to a Resource file Resource ID, a handle inside the actual executable, an OEM Icon, or they can be a LPCSTR string path to the file if loading from disk or a string of the form `#xxx` indicating a resource ID.

See: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadimagea

From what I understand on Linux, it's a reference inside of, or to, a `.desktop` file.

The `void*` is for avoiding referencing the OS-specific types in the base class and the fact that, at least for Windows, the argument can be one of a couple different data formats. The intended derived types `WindowWindows` and `WindowLinux` and `WindowMacOS` should handle the appropriate types inside the function:

This also exemplifies the use of `nativeHandle()` through the interface.

I would figure Linux to be similar.

Example of loading off disk:
```cpp
WindowWindows::icon(void* iconResource)
{
    HWND hWnd = static_cast<HWND>(nativeHandle());
    bool isResourceIdOrPath = !!IS_INTRESOURCE(iconResource);
    if(isResourceIdOrPath)
    { //Resource ID or OEM Resource ID
        //iconResource would be a custom resource ID from a .rc Resource file or one of the following defined in WinUser.h:
        //OIC_SAMPLE, OIC_HAND, OIC_QUES, OIC_BANG, OIC_NOTE, OIC_WINLOGO, OIC_WARNING, OIC_ERROR, OIC_HAND, OIC_INFORMATION, OIC_SHIELD
        auto resourceAsInt = static_cast<int>(iconResource);
        //0,0 and fuLoad != LR_DEFAULTSIZE  says use resource size
        HICON hIcon = ::LoadImageA(nullptr, MAKEINTRESOURCE(resourceAsInt), IMAGE_ICON, 0, 0, 0);
        ::SendMessage(hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hIcon);
    }
    else
    { //Path on disk to file or resource ID string
        const char* imagePath = reinterpret_cast<const char*>(iconResource);
        if(imagePath[0] == '#')
        {
            std::string imagePathAsResourceId{imagePath};
            auto resourceIdAsInt = static_cast<int>(std::stoi(imagePathAsResourceId.substr(1)));
            //0,0 and fuLoad != LR_DEFAULTSIZE  implies using the actual resource size. For multi-icon .ico files, the first one is picked if no size is supplied.
            HICON hIcon = ::LoadImageA(nullptr, MAKEINTRESOURCE(resourceAsInt), IMAGE_ICON, 0, 0, LR_LOADFROMFILE);
            ::SendMessage(hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hIcon);
        }
        else
        {
            //0,0 and fuLoad != LR_DEFAULTSIZE  implies using the actual resource size. For multi-icon .ico files, the first one is picked if no size is supplied.
            HICON hIcon = ::LoadImageA(nullptr, imagePath, IMAGE_ICON, 0, 0, LR_LOADFROMFILE);
            ::SendMessage(hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hIcon);
        }
    }
}
```

NAS2D/Renderer/Window.h
cugone: That's the intention, but in actual use-case, the user should call this from a `Window` interface pointer through the `Renderer` interface and cast appropriately to avoid OS-specific types bleeding through the interface. For example, DirectX needs a reference to the native window handle in some cases and you'd need to cast to an HWND pointer.

For reference of use see reply to your comment on SetIcon

NAS2D/Renderer/Window.h
cugone: As above for NativeHandle, this is needed for interoperability between the Window and DX/OpenGL renderers.

NAS2D/Renderer/Window.h
cugone: Note to self: The documentation for WM_SETICON is awful.

DanRStevens: That is one of the challenges here. Each platform has a rather different way of doing this, and there isn't really a way to have a single function signature that can handle all platforms, and still be type safe without need of any casts. That should maybe give pause about putting such a method in a base class interface.

I'm afraid I don't have a good suggestion for this.

----

I should probably mention that I would really like to stay away from platform specific code. It's very hard for a single developer to properly test and update code if it runs differently on their local development platform from other platforms. I would have quite some difficulty in maintaining Windows specific code. I would rather see subtypes specific to generic windowing libraries, such as SDL, that abstract away platform differences. There may be other windowing libraries that could be used instead of SDL. I suppose someone could write a Windows API specific one. Though I would like to avoid putting that in the NAS2D library. Honestly, I only expect us to support SDL. Anything else would be a client concern. I don't want to take support any further than having a generic enough API that developers could derive their own classes to handle other libraries.

----

It might be possible to kind of abstract away differences with a generic `Image` type. I don't think it's a perfect abstraction though. The current `RendererOpenGL::window_icon` method has SDL specific code in it, which is similar to SDL specific code in the `Image` class. I don't have any great ideas for how a supposed generic `Image` class could be used by different windowing libraries though. I imagine it would be tied to a specific library, such as `SdlImage`. At that point, you could use `friend` declarations between related classes that depend on the same underlying library. Of course that means splitting a lot of the resource classes into a generic base type and a library specific version. I'm not really looking forward to doing that.

NAS2D/Renderer/Window.h
DanRStevens: I would much prefer an interface that avoids requiring casts to function correctly. I also want to avoid leaking OS or library specific type information. I know that will be quite a challenge here. I don't really have an answer at this point. This is perhaps one of the reasons why I've been avoiding working on that part of the code. I suspect there is a way to do it though.

NAS2D/Renderer/Window.h
cugone: Just tested this with my own code seen below, loading an `.ico` file at the point of window registration with the OS makes this trivial and it "just works" on Windows. There's no need for a `SetIcon` or `GetIcon`. Changing them at runtime is not a thing anyway.

```cpp
bool Window::Register() noexcept {
    _hInstance = GetModuleHandle(nullptr);
    memset(&_wc, 0, sizeof(_wc));
    auto window_class_name = "Simple Window Class";
    _wc.cbSize = sizeof(_wc);
    _wc.style = CS_OWNDC | CS_HREDRAW | CS_VREDRAW | CS_DBLCLKS;
    _wc.lpfnWndProc = EngineMessageHandlingProcedure;
    _wc.cbClsExtra = 0;
    _wc.cbWndExtra = 0;
    _wc.hInstance = _hInstance;
    //Enforce downstream projects must put their custom icon.ico file in a specific spot to be found by the engine.
    auto path = FileUtils::GetKnownFolderPath(FileUtils::KnownPathID::GameData) / "Resources/Icons/icon.ico";
    //.ico file contains multiple image sizes. Load image Big image that goes in the taskbar and Alt+Tab window-switching list.
    HICON iconTaskbar = (HICON)::LoadImage(nullptr, path.string().c_str(), IMAGE_ICON, 64, 64, LR_LOADFROMFILE);
    _wc.hIcon = iconTaskbar;
    _wc.hCursor = ::LoadCursor(NULL, IDC_ARROW);
    _wc.hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
    _wc.lpszMenuName = NULL;
    _wc.lpszClassName = window_class_name;
    //Load the small version that goes in the title bar.
    HICON iconTitlebar = (HICON)::LoadImage(nullptr, path.string().c_str(), IMAGE_ICON, 16, 16, LR_LOADFROMFILE);
    _wc.hIconSm = iconTitlebar;
    return 0 != RegisterClassEx(&_wc);
}
```

The above makes the icon appear in all four places: The title bar, the task bar, the alt+tab list, and the .exe image. There's no need for dealing with `.rc` resource files and no WM_SETICON message is ever sent (and I've never seen a program change them at runtime due to branding familiarity); making setup much easier.

Enforcing specific asset locations to be read by a game engine is not new. Unreal expects assets to be in specific locations in the Content folder to be loaded correctly and it's no different for the project's window icon. Unreal enforces this by specifying the image in the project settings at compile time that is then copied into the correct place and the path is "hard coded" by the code generator for use during window creation.

DanRStevens: You make a good point about not changing the icon at runtime. It would make reasonable sense to have the icon set once during construction, with no way to change it afterwards.

DanRStevens: I took a look at the original code from `RendererOpenGL`. It will handle loading icons from the `Filesystem` search path just fine. Seems like a reasonable way to continue setting window icons.

```cpp
void RendererOpenGL::window_icon(const std::string& path)
{
	auto iconData = Utility<Filesystem>::get().read(path);
	SDL_Surface* icon = IMG_Load_RW(SDL_RWFromConstMem(iconData.c_str(), static_cast<int>(iconData.size())), 1);
	if (!icon)
	{
		throw std::runtime_error("Failed to set window icon: " + path + " : " + SDL_GetError());
	}

	SDL_SetWindowIcon(underlyingWindow, icon);
	SDL_FreeSurface(icon);
}
```

I would make sense to add a call to such a method from the constructor, assuming an icon was specified during construction.

I suppose we could leave it as a separate method though, as that enables people to change the icon later, if they want, though would not be required to.

There is currently no way to retrieve the window icon. I'm fine with that. We don't need to support getting the window icon. Presumably they were the ones that set it, so they should already know what it is.

DanRStevens: I guess this boils down to setting the window icon by filename, rather than by some kind of `Image` reference. I'm fine with such a restriction. It matches existing behaviour, so no loss of functionality.

----

Looking back at the proposed code and my suggestion, I suppose we could generalize to an `Image` reference if we gave the `RendererOpenGL` friend access to the `SDL_Surface` member of an `Image` object. That would give the API more flexibility. Though that's more of a potential future upgrade to the API. I don't think we need to start there. And we might not want to use `friend` access. Probably better to make as minimal of a change as possible when starting out. I think I'd be more comfortable with keeping the icon specified through a filename to start with.

cugone: > 
> ...The above makes the icon appear in all four places: The title bar, the task bar, the alt+tab list, and the .exe image. There's no need for dealing with `.rc` resource files...

Minor correction to this statement. A resource file and associated resource.h file (generated by visual studio) is required in order to embed the icon into the .exe and to appear as the .exe image. The icon appears in all other places with just an appropriate path provided at window registration time.

DanRStevens:
This has been sitting for a while, and I fear the PR has lost momentum.

This task was actually quite difficult. It seems there were a number of awkward design considerations when tackling this issue. There's maybe a lack of consensus and clear plan on what the final design should look like. In cases like that, it can be hard to get agreement for larger change sets, as they are more likely to contain bits and pieces that may have new negative aspects compared to the old code (such as less type safety, or platform dependent code, or new features that may be awkward to use or implement, etc.). In those cases, it may be possible to minimize the set of changes to try and eliminate downsides that didn't exist in the original code, and defer dealing with some of the issues that may already be present. Basically, try to find a minimal change set that somehow improves the code, without introducing new concerns (and doesn't necessarily solve all the old problems).

In this case, we already had working code for windowing, icons, cursors, and various rendering methods. Could we tweak that code to sort those exact same methods, with the same basic API, so that the methods are split into two different classes, which might be strongly coupled to each other, possibly with `friend` access. As it stands, the original code was already tightly coupled, with everything in one class having access to everything else. Limiting that access to one or two friend functions would be an improvement.

Specifically, I would like to avoid changes to things like icon loading, public access to raw underlying window handles, and public access to device contexts. At least for now. Changes to those areas might be split off into future PRs. Basically, what's the minimal API disturbance needed to split the class, without adding any new features. Perhaps an extra method to get a `Renderer` for a `Window`'s client area?

cugone:
> 
> This has been sitting for a while, and I fear the PR has lost momentum.
> 
I was under the impression I was waiting for a pending review from @ldicker83 due to requested changes that had been made in the mean time.

> This task was actually quite difficult. It seems there were a number of awkward design considerations when tackling this issue. There's maybe a lack of consensus and clear plan on what the final design should look like. In cases like that, it can be hard to get agreement for larger change sets, as they are more likely to contain bits and pieces that may have new negative aspects compared to the old code (such as less type safety, or platform dependent code, or new features that may be awkward to use or implement, etc.). In those cases, it may be possible to minimize the set of changes to try and eliminate downsides that didn't exist in the original code, and defer dealing with some of the issues that may already be present. Basically, try to find a minimal change set that somehow improves the code, without introducing new concerns (and doesn't necessarily solve all the old problems).
> 
> In this case, we already had working code for windowing, icons, cursors, and various rendering methods. Could we tweak that code to sort those exact same methods, with the same basic API, so that the methods are split into two different classes, which might be strongly coupled to each other, possibly with `friend` access. As it stands, the original code was already tightly coupled, with everything in one class having access to everything else. Limiting that access to one or two friend functions would be an improvement.
> 
> Specifically, I would like to avoid changes to things like icon loading, public access to raw underlying window handles, and public access to device contexts. At least for now. Changes to those areas might be split off into future PRs. Basically, what's the minimal API disturbance needed to split the class, without adding any new features. Perhaps an extra method to get a `Renderer` for a `Window`'s client area?
> 
The whole point was to separate `Window` from `Renderer`. This does that with the caveat that sub-classes for the OS-specific versions must be made due to how they handle window creation differently.

In the mean time I've removed the icon functions in favor of resolving the conflict in one of two approaches:
- Make Window an ADT instead of a pure Interface and provide the Rule of 6 with a user-defined constructor that takes an icon path. IIRC, this is similar to SDL:
```cpp
Window() = default;
Window(const Window& other) = default;
Window(Window&& rother) = default;
Window& operator=(const Window& rhs) noexcept = default;
Window& operator=(Window&& rhs) noexcept = default;
virtual ~Window() noexcept = default;

explicit Window(std::filesystem::path iconPath) noexcept;
//...
private:
std::filesystem::path m_iconPath{};
```
- Preserve the Interface and require that for icons to work, the engine absolutely requires they be located in a specific path that is used at window registration time:
```cpp
//Internal OS-specific window registration that is called during construction.
bool Window::Register() noexcept {
    //...Snip
    auto path = FileUtils::GetKnownFolderPath(FileUtils::KnownPathID::GameData) / "Resources/Icons/icon.ico";
    HICON iconTaskbar = (HICON)::LoadImage(nullptr, path.string().c_str(), IMAGE_ICON, 64, 64, LR_LOADFROMFILE);
    _wc.hIcon = iconTaskbar;
    //...Snip
    HICON iconTitlebar = (HICON)::LoadImage(nullptr, path.string().c_str(), IMAGE_ICON, 16, 16, LR_LOADFROMFILE);
    _wc.hIconSm = iconTitlebar;
    return 0 != RegisterClassEx(&_wc);
}
```

DanRStevens:
I gave Leeor a bit of a poke about this.

----

I really want to avoid platform specific code. I know your PR doesn't really have platform specific code, most of that was in comments on the PR. Though the comments to the PR kind of imply some intention to introduce platform specific code. I'd rather rely on SDL to provide isolation from that. Further, the changes here expose certain untyped values that may encourage platform specific code. That's something I want to avoid in the library. I was also looking at this and thinking this is just a new interface class. It's not actually used anywhere, so we can't really fully assess how it would be used. I think in a way that kind of hides what will end up in the concrete subclasses, where I'd rather rely on SDL than platform specific code.

Currently we have a `RendererOpenGL` class with a hard dependency on SDL and OpenGL. I think a minimal improvement would be to have a `Window` class with a hard dependency on `SDL`, and a `RendererOpenGL` with a hard dependency on OpenGL. It might not quite make that cut quite so clean with an initial implementation, but that's the general idea I sort of had for the split.

ldicker83:
Wow... so there's a lot to go over here. I'm too tired to process any of it today, I'll have some time this weekend, I'll look over this then and comment.

cugone:
Any traction on the review? Will you need the rest of the weekend?

DanRStevens:
So re-reading the last few messages, my thought it this PR attempts to touch on too many problems all at once, which is the real reason why review got bogged down. More time to review won't fix that.

There are a number of significant new design aspects with the proposed interface, some of which may have negative characteristics. When new negative characteristics are introduced, they tend to bog down review, and make people hesitant to merge. I would recommend attempting to re-work some of this while trying to minimize the impact to high level design.

As it stands now, the new interface isn't a description of existing practice, but rather a new proposal. It contains significant new aspects in terms of native window handles and device contexts. There were also significant changes to how window icons were handled, though that part has since been cut. There are maybe a few other smaller design aspects too, such as terminology used in the interface (`Dimension` versus `Size`), or use of individual `Point` and `Vector` to effectively specify a `Rectangle` (and specifying those components in reverse from standard order).

To proceed with less resistance, it might help to start with a description of current behaviour, and then from there work on updating the design. I would recommend the following tasks, which could be separate PRs:
- Sort the `Renderer` methods into window and draw functions, keeping them in their original class, but as separate blocks. There would be no API changes here, only sorting the order of methods. It may help to keep header declaration order and implementation order in sync, so both are re-ordered into a new grouping.
- Split the `Renderer` class so window related functions go into a `Window` class, leaving draw functions in `Renderer`. Use `friend` declarations to keep the now independent classes talking to each other, and to minimize API changes. This should hopefully consist mostly of moving the window block of code from the above sorting into a new file, wrapped in a new class.
- Refactor the API, and perhaps start proposing more significant API changes. This might be where we'd add a constructor parameter for a window icon filename, and other such changes.

cugone:
I'd just like to point out that this is still in `RendererOpenGL`:

```cpp
// UGLY ASS HACK!
// This is required for mouse grabbing in the EventHandler class.
SDL_Window* underlyingWindow = nullptr;
```

Getting the underlying window is absolutely a required part of the Window interface, especially to solve this "Ugly Ass Hack" as it were.

DanRStevens:
I'm aware of this problem. That's part of why I suggested `friend` access to deal with details such as this.

I would rather deal with the `SDL_Window` type directly, rather that cast to and from a `void*`.

DanRStevens:
I think maybe I should close this. The associated issue is now closed. This PR has been inactive for a year. The code added in this PR wasn't used anywhere (even in this PR), so it was effectively untested. Though it did result in some good discussion. There are clearly some tough design decisions involved with this work.

There has been some related work in PR #1058, and PR #1059. These were very minimal changes, to kind of break up the `Renderer` interface and define what should be the responsibility of the new `Window` class. There was no real change in architecture, nor any attempt to solve some of the challenging design decisions. More work is needed, though probably something minimal, just enough to break the inheritance relationship between the new `Window` class and `Renderer`.

PR #905, opened 2021-07-04
Remove File class

DanRStevens:
Deprecate and then remove `File` class.

The deprecated commits can be (were) used to check dependent projects for uses of `File` or the deprecated `Filesystem` methods that used `File`. Once they were updated, it became safe to remove the `File` class and associated methods.

Reference: #904


cugone:
Be careful leaving `[[deprecate]]` attributes hanging around. MSVC treats any use of features marked deprecated in this way as an error in all cases instead of a warning like the standard suggests. It's good for enforcing a "do not use, this will be removed in the future" notation but there better be an alternative in place before marking the old one deprecated or the feature will become unusable in Visual Studio.

DanRStevens:
That's ultimately why I just outright removed stuff in the next two commits. You may notice the failed build after the first two commits deprecating stuff.

DanRStevens:
Resolved merge conflict with recently merged change.

PR #920, opened 2021-07-13
Remove static_cast in AnimationSet data loading

DanRStevens:
Reference: #797

Use the proper type in the `Dictionary::get<>` method, and avoid the need for a `static_cast` later.


DanRStevens:
This PR causes a bug with image loading due to the `-1` value used for `"delay"`.

PR #923, opened 2021-07-15
Documentation pass on files in Mixer folder

cugone:
Ref: #861

Smaller chunk of #899.

I plan on splitting the documentation fixes (re: deletions...) into smaller chunks by folder to avoid large-scale merge conflicts.

### DIFF
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -20,9 +20,10 @@ namespace NAS2D
 	class Music;
 
 	/**
-	 * @brief Mixer base class.
+	 * \brief Mixer base class.
 	 * 
-	 * Provides a standard Mixer interface. The base Mixer can be used but will act as a NULL interface.
+	 * Provides a standard Mixer interface.
+	 * The base Mixer can be used but will act as a NULL interface.
 	*/
 	class Mixer
 	{
@@ -61,14 +62,14 @@ namespace NAS2D
 		void resumeAllAudio();
 
 		/**
-		 * @brief Sets the sound volume.
-		 * @param level Volume level to set. Valid range is [0, 128].
+		 * \brief Sets the sound volume.
+		 * \param level Volume level to set. Valid range is [0, 128].
 		*/
 		virtual void soundVolume(int level) = 0;
 
 		/**
-		 * @brief Sets the music volume.
-		 * @param level Volume level to set. Valid range is [0, 128].
+		 * \brief Sets the music volume.
+		 * \param level Volume level to set. Valid range is [0, 128].
 		*/
 		virtual void musicVolume(int level) = 0;
 
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -21,7 +21,7 @@ namespace NAS2D
 
 	/**
 	 * \brief Mixer base class.
-	 * 
+	 *
 	 * Provides a standard Mixer interface.
 	 * The base Mixer can be used but will act as a NULL interface.
 	*/
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -20,10 +20,7 @@ namespace NAS2D
 	class Music;
 
 	/**
-	 * \brief Mixer base class.
-	 *
-	 * Provides a standard Mixer interface.
-	 * The base Mixer can be used but will act as a NULL interface.
+	Mixer base class.
 	*/
 	class Mixer
 	{
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -19,18 +19,16 @@ namespace NAS2D
 	class Sound;
 	class Music;
 
-
 	/**
-	 * Mixer base class.
-	 *
-	 * Provides a standard Mixer interface. The base Mixer can be used
-	 * but will act as a NULL interface.
-	 */
+	 * @brief Mixer base class.
+	 * 
+	 * Provides a standard Mixer interface. The base Mixer can be used but will act as a NULL interface.
+	*/
 	class Mixer
 	{
 	public:
-		static const int CONTINUOUS = -1; /**< Constant representing a continuous loop. */
-		static const int DEFAULT_FADE_TIME = 500; /**< Default fade time. */
+		static const int CONTINUOUS = -1;
+		static const int DEFAULT_FADE_TIME = 500;
 
 	public:
 		Mixer() = default;
@@ -41,121 +39,46 @@ namespace NAS2D
 		virtual ~Mixer() = default;
 
 	public:
-		/**
-		 * Plays a sound on the first available sound channel.
-		 *
-		 * \param	sound	A reference to a Sound Resource.
-		 */
 		virtual void playSound(const Sound& sound) = 0;
-
-		/**
-		 * Stops playing all sounds on all channels.
-		 */
 		virtual void stopSound() = 0;
-
-		/**
-		 * Pauses sound on all channels.
-		 */
 		virtual void pauseSound() = 0;
-
-		/**
-		 * Resumes sound on all channels.
-		 */
 		virtual void resumeSound() = 0;
-
-		/**
-		 * Starts playing a Music track.
-		 *
-		 * \param music	Reference to a Music Resource.
-		 * \param loops	Number of times to repeat the music.
-		 */
 		void playMusic(const Music& music, int loops = Mixer::CONTINUOUS);
-
-		/**
-		 * Stops all playing music.
-		 */
 		virtual void stopMusic() = 0;
-
-		/**
-		 * Pauses the currently playing Music.
-		 */
 		virtual void pauseMusic() = 0;
-
-		/**
-		 * Resumes the currently paused Music.
-		 *
-		 * \note	It is safe to call this function if the music is stopped or already playing.
-		 */
 		virtual void resumeMusic() = 0;
 
-		/**
-		 * Starts a Music track and fades it in to the current Music volume.
-		 *
-		 * \param	music	Reference to a Music Resource.
-		 * \param	loops	Number of times the Music should be repeated. -1 for continuous loop.
-		 * \param	time	Time, in miliseconds, for the fade to last. Default is 500.
-		 */
+		//TODO: Consider type-safety via std::chrono::milliseconds? or rename variable to fadeInMS
 		virtual void fadeInMusic(const Music& music, int loops = Mixer::CONTINUOUS, int time = Mixer::DEFAULT_FADE_TIME) = 0;
 
-		/**
-		 * Fades out the currently playing Music track.
-		 *
-		 * \param	time	Time, in miliseconds, for the fade to last. Default is 500.
-		 */
+		//TODO: Consider type-safety via std::chrono::milliseconds? or rename variable to fadeOutMS
 		virtual void fadeOutMusic(int time = Mixer::DEFAULT_FADE_TIME) = 0;
 
-		/**
-		 * Gets whether or not music is currently playing.
-		 */
 		virtual bool musicPlaying() const = 0;
 
-		/**
-		 * Stops all music and sound.
-		 */
 		void stopAllAudio();
-
-		/**
-		 * Pauses all music and sound.
-		 */
 		void pauseAllAudio();
-
-		/**
-		 * Resumes all paused music and sound.
-		 */
 		void resumeAllAudio();
 
 		/**
-		 * Sets the sound volume.
-		 *
-		 * \param	level	Volume level to set. Valid values are 0 - 128.
-		 */
+		 * @brief Sets the sound volume.
+		 * @param level Volume level to set. Valid range is [0, 128].
+		*/
 		virtual void soundVolume(int level) = 0;
 
 		/**
-		 * Sets the music volume.
-		 *
-		 * \param	level	Volume level to set. Valid values are 0 - 128.
-		 */
+		 * @brief Sets the music volume.
+		 * @param level Volume level to set. Valid range is [0, 128].
+		*/
 		virtual void musicVolume(int level) = 0;
 
-		/**
-		 * Gets the sound volume.
-		 *
-		 * \return	Volume level. Valid values are 0 - 128.
-		 */
 		virtual int soundVolume() const = 0;
-
-		/**
-		 * Gets the music volume.
-		 *
-		 * \return	Volume level. Valid values are 0 - 128.
-		 */
 		virtual int musicVolume() const = 0;
 
 		SignalSource<>& musicCompleteSignalSource();
 
 	protected:
-		Signal<> mMusicComplete; /**< Signal used when music finished playing. */
+		Signal<> mMusicComplete;
 	};
 
 } // namespace
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -71,10 +71,15 @@ namespace NAS2D
 		void pauseAllAudio();
 		void resumeAllAudio();
 
-		//Returned values are in the closed range [0, 128].
+		/**
+		 * \param level Volume level, valid values are in the range [0, 128]
+		*/
 		virtual void soundVolume(int level) = 0;
 
-		//Returned values are in the closed range [0, 128].
+
+		/**
+		 * \param level Volume level, valid values are in the range [0, 128]
+		*/
 		virtual void musicVolume(int level) = 0;
 
 		virtual int soundVolume() const = 0;
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -48,8 +48,8 @@ namespace NAS2D
 		/**
 		 * Starts playing a Music track.
 		 *
-		 * \param music	Reference to a Music Resource.
-		 * \param loops	Number of times to repeat the music.
+		 * \param music: Reference to a Music Resource.
+		 * \param loops: Number of times to repeat the music.
 		 */
 		void playMusic(const Music& music, int loops = Mixer::CONTINUOUS);
 		virtual void stopMusic() = 0;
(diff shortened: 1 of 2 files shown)


NAS2D/Mixer/Mixer.h
```diff
@@ -19,18 +19,16 @@ namespace NAS2D
...
 	class Music;
 
-
 	/**
-	 * Mixer base class.
-	 *
-	 * Provides a standard Mixer interface. The base Mixer can be used
-	 * but will act as a NULL interface.
-	 */
+	 * @brief Mixer base class.
```
DanRStevens, line 23: We're using [`JAVADOC_AUTOBRIEF = YES`](https://www.doxygen.nl/manual/config.html#cfg_javadoc_autobrief) in the `Doxyfile`, so it's possible to omit explicit `@brief`. I suppose this becomes a question of style now. Use of `\brief` was removed in PR #856.

----

We should probably settle on a convention between `@` and `\` for Doxygen tags. I think I prefer the `@`, though the code base has typically used `\`.

NAS2D/Mixer/Mixer.h
```diff
@@ -19,18 +19,16 @@ namespace NAS2D
...
+	 * 
+	 * Provides a standard Mixer interface. The base Mixer can be used but will act as a NULL interface.
```
DanRStevens, line 25: Is there a way we can set a standard line wrap width, perhaps with `.editorconfig`? I get the 80 column marker in my editor. I'd be fine with increasing that to 120 characters per line, which fits quite nicely on most modern monitors.

DanRStevens (commented):
I had a few technical comments that might be wide reaching in their impact.

NAS2D/Mixer/Mixer.h
cugone: The use of the `@` is a consequence of the `.editorconfig` file. Visual Studio auto-generates the documentation via `Ctrl+/` and inserts the `@`.

The happy accident this provides is that Intellisense correctly displays the comments...sort of. See below.

NAS2D/Mixer/Mixer.h
cugone: This was something I was concerned about. When looking at the comments via Visual Studio's Intellisense it flat out *ignores* newlines and I do not know how to add artificial ones as the `<br>` and `\n` tags are not supported (they show up as-is). This is frustrating.

NAS2D/Mixer/Mixer.h
```diff
@@ -41,121 +39,46 @@ namespace NAS2D
...
-		 */
 		virtual void resumeSound() = 0;
-
-		/**
-		 * Starts playing a Music track.
-		 *
-		 * \param music	Reference to a Music Resource.
-		 * \param loops	Number of times to repeat the music.
-		 */
 		void playMusic(const Music& music, int loops = Mixer::CONTINUOUS);
```
DanRStevens, line 46: Along with removing the documentation, would it also make sense to remove the `loops` parameter (in a separate PR)? I feel like as long as we keep that parameter, it should perhaps have some documentation, particularly since the default is to loop continuously.

NAS2D/Mixer/Mixer.h
```diff
@@ -41,121 +39,46 @@ namespace NAS2D
...
 	public:
-		/**
-		 * Plays a sound on the first available sound channel.
-		 *
-		 * \param	sound	A reference to a Sound Resource.
-		 */
 		virtual void playSound(const Sound& sound) = 0;
-
-		/**
-		 * Stops playing all sounds on all channels.
```
DanRStevens, line 52: I suppose the "all channels" part on some of those methods can be implied. That leaves the rest quite redundant.

NAS2D/Mixer/Mixer.h
```diff
@@ -41,121 +39,46 @@ namespace NAS2D
...
-		 * \return	Volume level. Valid values are 0 - 128.
-		 */
 		virtual int soundVolume() const = 0;
-
-		/**
-		 * Gets the music volume.
-		 *
-		 * \return	Volume level. Valid values are 0 - 128.
-		 */
 		virtual int musicVolume() const = 0;
```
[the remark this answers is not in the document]
cugone: Some implementations I see in places is a `float` that specifies `[0.0,1.0]` ([XAudio2](https://docs.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2voice-setvolume) does this, but also allows negative values to reverse the phase).

In this case, an `unsigned char` might be useful, just document a clamping of the value `[0,128]`. Not in comments, obviously.

NAS2D/Mixer/Mixer.h
DanRStevens: @ldicker83 : You should probably take a look at this comment.

----

I noticed newer Doxygen examples tend to use `@` rather than `\` for Doxygen commands. Should we perhaps consider switching style?

----

Should we include `@brief` explicitly, or rely on `JAVADOC_AUTOBRIEF = YES`? I would prefer to omit explicit commands if they don't need to be specified, and instead rely on the auto brief feature.

NAS2D/Mixer/Mixer.h
DanRStevens: I'm not sure I really understand the response here. Why would you be trying to force newlines? Wouldn't the point here be to auto flow the text?

NAS2D/Mixer/Mixer.h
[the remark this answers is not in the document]
DanRStevens: This one isn't resolved.

NAS2D/Mixer/Mixer.h
```diff
@@ -41,121 +39,46 @@ namespace NAS2D
...
 
-		/**
-		 * Starts a Music track and fades it in to the current Music volume.
-		 *
-		 * \param	music	Reference to a Music Resource.
-		 * \param	loops	Number of times the Music should be repeated. -1 for continuous loop.
-		 * \param	time	Time, in miliseconds, for the fade to last. Default is 500.
-		 */
+		//TODO: Consider type-safety via std::chrono::milliseconds? or rename variable to fadeInMS
 		virtual void fadeInMusic(const Music& music, int loops = Mixer::CONTINUOUS, int time = Mixer::DEFAULT_FADE_TIME) = 0;
```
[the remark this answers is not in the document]
DanRStevens: I'd like to get rid of `//TODO` comments. I'm fine with changing the parameter type to be `std::chrono::milliseconds`, or updating the variable name to be clear about milliseconds, or restoring the documentation to say milliseconds.

NAS2D/Mixer/Mixer.h
cugone: For horizontal flow, sure, but for vertical flow not so much. It's primarily a readability issue. Right now, you get giant walls of text regardless of the presence of vertical whitespace in the actual comment. There are clear delimited sections for the Brief, Parameters, and raw comment, but that's it. If the text of any of them go on too long you get the aforementioned walls of text in that section.

DanRStevens: I feel like there's some missing detail here.

Is it perhaps flowing the two lines together? Does it correctly handle auto brief text as such? Or does it need to be manually marked up with `@brief`?

The one line that was reflowed here doesn't seem particularly long, so I'm not sure what constitutes a wall of text.

----

The other part of this is that text was reflowed by the IDE, though not all IDEs will necessarily default to the same column marker (80 columns, 120 columns, etc.). Many IDEs do support `.editorconfig` though. Would it be possible to set a `.editorconfig` value such that most common IDEs would have the same column marker to guide how text is flowed? Doesn't matter if it's either manually or automatically flowed.

NAS2D/Mixer/Mixer.h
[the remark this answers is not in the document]
cugone: An API change should probably be a separate PR; Like you, I'd prefer switching to `std::chrono::milliseconds`. The comment can be removed at that time. (That's what it's for. :) )

NAS2D/Mixer/Mixer.h
cugone: The issue is Intellisense, the code-completion subsystem for Visual Studio and Visual Studio Code. The Doxygen interpreter does not handle comments on separate lines. It runs them all together until its internal, pre-defined, and unedittable, column length is reached then it does a basic wrap to the next line.

Example of ignored newlines: 
NoNewlines

Example of section breaks:
SectionBreaks


I noticed that unknown doxygen tags, like the "warning" tag in the `MixerSDL` comment header only appears in the raw comment section where it could conceivably be missed:
NoUnknowns

Example of Wall'o'Text:
WalloText


Long non-Doxygen comment blocks (as multiple single-line comments `//` or a block `/*...*/`) aren't much better, they only handle one newline per column. Multiple newlines are collapsed in to one. Additionally, regardless of the comment format, if it's over ten lines, it gets truncated.

This is why I advocate type-safety, and good function/variable names over trying to document features in comments.

If you need to explain how a function works or its acceptable use, use an actual long-form API document similar to docs.microsoft or khronos.org

NAS2D/Mixer/Mixer.h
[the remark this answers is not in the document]
cugone: Could you elaborate? Do you want to leave in the comment about the valid range? (Personally, the rest of the comment is duplicative and superfluous.)

```cpp
//Value returned is in the closed range [0,128].
virtual int soundVolume() const = 0;

//Value returned is in the closed range [0,128].
virtual int musicVolume() const = 0;
```

NAS2D/Mixer/Mixer.h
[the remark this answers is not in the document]
DanRStevens: In the case, we should probably keep the original documentation until such a change is made.

NAS2D/Mixer/Mixer.h
DanRStevens: Those pictures certainly help explain what's going on. For the last picture, does the original text flow match what is displayed in the raw comment section, or was it reflowed from the original source file? I see it's from a version of TinyXML2 that's not in the project, and checking online for the original source shows a file without Doxygen comments.

----

At any rate, I feel like this discussion is moving away from the original comment I was trying to make. I'm concerned about reflow noise in diffs from IDEs attempting to adjust comments according to different desired line widths. If one IDE is set to 80, and another is set to 120, and they both want to reflow comments to their own desired length, that could get annoying. It would be nice if we could have one settings files, such as `.editorconfig`, which most common IDEs will read and respect, which sets a common line width for all the IDEs currently being used on this project.

Though yes, if we don't have long comments, because the code is written clearly with good types and good names, that would of course be best. Though I suspect there will always be at least a few somewhat long comments where max line width will matter.

NAS2D/Mixer/Mixer.h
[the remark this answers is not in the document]
DanRStevens: Yes, I want the returned range to be documented. I would prefer the comment be in Doxygen style. I don't care so much about the `@brief` text.

NAS2D/Mixer/Mixer.h
cugone: This is one part of the API that didn't make sense to me. Shouldn't the default be to *not* loop?

NAS2D/Mixer/Mixer.h
cugone: > Those pictures certainly help explain what's going on. For the last picture, does the original text flow match what is displayed in the raw comment section, or was it reflowed from the original source file? I see it's from a version of TinyXML2 that's not in the project, and checking online for the original source shows a file without Doxygen comments.
> 
I didn't realize I had cut off the comment. Here's the text as-is:
```cpp

        /** Given an attribute name, Attribute() returns the value
            for the attribute of that name, or null if none
            exists. For example:

            @verbatim
            const char* value = ele->Attribute( "foo" );
            @endverbatim

            The 'value' parameter is normally null. However, if specified,
            the attribute will only be returned if the 'name' and 'value'
            match. This allow you to write code:

            @verbatim
            if ( ele->Attribute( "foo", "bar" ) ) callFooIsBar();
            @endverbatim

            rather than:
            @verbatim
            if ( ele->Attribute( "foo" ) ) {
                if ( strcmp( ele->Attribute( "foo" ), "bar" ) == 0 ) callFooIsBar();
            }
            @endverbatim
        */
```
As you can see, they have `@verbatim/@endverbatim` doxygen tags. The Doxygen Interpreter for Intellisense does not understand them so it ignores the tags and removes anything associated with them. Additionally the description section collapses any consecutive newlines into one and then truncates after 10 lines.

The Raw comment section is preserved up to 10 lines as well before being truncated.

> At any rate, I feel like this discussion is moving away from the original comment I was trying to make. I'm concerned about reflow noise in diffs from IDEs attempting to adjust comments according to different desired line widths. If one IDE is set to 80, and another is set to 120, and they both want to reflow comments to their own desired length, that could get annoying. It would be nice if we could have one settings files, such as `.editorconfig`, which most common IDEs will read and respect, which sets a common line width for all the IDEs currently being used on this project.
> 
> Though yes, if we don't have long comments, because the code is written clearly with good types and good names, that would of course be best. Though I suspect there will always be at least a few somewhat long comments where max line width will matter.

I agree. I know a specified maximum column length can be enforced with `.clang-format` or `.clang-tidy`, I do not know about `.editorconfig`. Regardless, the formatting options in `.editorconfig` were removed and the only thing retained was the commands to tell Visual Studio to use `/*` style comments over `///`.

NAS2D/Mixer/Mixer.h
DanRStevens: I'm thinking playlist control should be handled separate from this class. In which case, it would make sense to not loop.

I think the current design maybe assumes a single background music track as default. Maybe due to limited scope for the game assets early on.

cugone: It also may be a function of the underlying audio engine. [An XAudio2 Buffer type](https://docs.microsoft.com/en-us/windows/win32/api/xaudio2/ns-xaudio2-xaudio2_buffer) requires setting up initial values for looping sample ranges and loop counts as well, but as mentioned,  it defaults to not looping, and playing the whole buffer.

NAS2D/Mixer/Mixer.h
DanRStevens: Looks like it preserves single newlines, but collapses multiple consecutive newlines. Anyway, not something I'm overly concerned about. Sounds like a minor limitation of one particular tool, with no real workaround, and so nothing to really do about it.

----

I'm not sure if `.clang-format` settings would be picked up by Atom. I suspect `.editorconfig` settings would be though, at least for something like the margin ruler. Of course, that brings up the question of if `clang-format` would pick up a setting from `.editorconfig`, and if you can avoid specifying such a setting in two different places, which could potentially conflict with each other.

I'm perhaps somewhat inclined to ignore this concern for now.

NAS2D/Mixer/Mixer.h
DanRStevens: There hasn't been any comment from Leeor. Unless he wants to chime in and say otherwise, I would say remove `\brief` and rely on auto brief. That would at least be consistent with current usage. No reason to reject something for a part that wasn't changed. :wink:

Besides, I think I like the shorter format of auto brief. The formatting of the documentation pretty much dictates what part is the brief anyway.

NAS2D/Mixer/Mixer.h
DanRStevens: As per my original comment, it would probably be wise to keep the documentation until the `loops` parameter is removed.

NAS2D/Mixer/Mixer.h
```diff
@@ -19,18 +19,17 @@ namespace NAS2D
...
 
-
 	/**
-	 * Mixer base class.
-	 *
-	 * Provides a standard Mixer interface. The base Mixer can be used
-	 * but will act as a NULL interface.
-	 */
+	 * \brief Mixer base class.
+	 *
```
DanRStevens, line 24: Ahh, and we tend to periodically make style changes such as removing line trailing spaces, such as this one.

NAS2D/Mixer/Mixer.h
cugone: Just to clarify, Remove all `\brief` tags in favor of just a description?

DanRStevens: > [`JAVADOC_AUTOBRIEF`](https://www.doxygen.nl/manual/config.html#cfg_javadoc_autobrief)
> 
> If the JAVADOC_AUTOBRIEF tag is set to YES then doxygen will interpret the first line (until the first dot) of a Javadoc-style comment as the brief description. If set to NO, the Javadoc-style will behave just like regular Qt-style comments (thus requiring an explicit @brief command for a brief description.)
> 
> The default value is: NO.

We have:
`JAVADOC_AUTOBRIEF = YES`

So yes, remove the `\brief`. The description on it's own will already be split into a brief and the rest.

cugone: Ok. I am in the middle of doing that and as usual, Intellisense has to be a butt. Removing the brief tag and just leaving the javadoc-style comment like so (extreme example):

```cpp
/**
* Mixer base class.
* Mixer base class.
* Mixer base class.
* Mixer base class.
* Mixer base class.
*/
```
Causes the intermediate optional `*`'s to show up:

```cpp
* Mixer base class
* Mixer base class
* Mixer base class
* Mixer base class
* Mixer base class
```

Removing the intermediate `*`'s fixes the problem.

DanRStevens: Probably nothing, but I noticed the `*` are not all vertically aligned. Does that make any difference? It's possible it may be very sensitive to formatting.

cugone: Doesn't make a difference.

cugone: The interpreter (and specification) only looks for the pattern:
```cpp
/**
... text...
*/
```
It doesn't care what's in between.

cugone:
*Poke*

The status of this affects work starting on #943, #944 and #945. Additionally I don't want to start the next folder until the previous is merged due to conflicts and adding too much to reviewers' plates.

NAS2D/Mixer/Mixer.h
```diff
@@ -19,18 +19,14 @@ namespace NAS2D
...
 
-
 	/**
-	 * Mixer base class.
-	 *
-	 * Provides a standard Mixer interface. The base Mixer can be used
-	 * but will act as a NULL interface.
-	 */
+	Mixer base class.
+	*/
```
DanRStevens, line 24: I have some minor concerns about the style change here that removes the `*` from the left of comment blocks. This issue seems to involve Visual Studio tooling, so perhaps @ldicker83 would be in a better position to comment.

Reference to discussion thread:
https://github.com/lairworks/nas2d-core/pull/923#discussion_r670109187

I guess I'm not sure if we want to change style based on peculiar quirks of one particular tool. Or maybe we do. I'm a bit hesitant though if the original style was widespread.

----

Actually, now that it's a single line comment, using a single line comment block might sidestep the concern.
```cpp
/// Mixer base class.
```

Err, wait, that's kind of a useless comment, isn't it? Maybe it can just be removed entirely? That would also sidestep the problem.

Anyway, the discussion of style might still be relevant to future cases.

DanRStevens (commented):
I've only got one small concern left, and it looks like it can be sidestepped.

DanRStevens (approved):
Ok, I agree with all these changes.

And wow, looking back, it's amazing how much discussion was generated from mostly just deleting some comments that don't really say much.

DanRStevens:
Ehh, build failure on master due to the `:` added to the documentation strings:
https://app.circleci.com/pipelines/github/lairworks/nas2d-core/1760/workflows/86d177a3-59d4-432a-ae86-ce3b9b7ca530/jobs/7157

Clang does actually check for certain Doxygen errors when `-Wdocumentation` is passed.

PR #922, opened 2021-07-15
Add copyright headers

cugone:
Ref: #861

This PR adds the missing copyright headers.

PR #924, opened 2021-07-15
Convert copyright symbols to Unicode

DanRStevens:
Related: #922

I noticed my editor wanted to auto correct the copyright symbols to a Unicode encoding. It makes it frustrating to edit files that aren't already converted, as the change either adds noise to commits, or needs to be manually skipped when adding changes.

Would be nice to get confirmation from people using Windows that Unicode copyright symbols display correctly in Visual Studio.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
- Copyright © 2008 - 2017 LairWorks Entertainment
+ Copyright Â© 2008 - 2017 LairWorks Entertainment
 
   This software is provided 'as-is', without any express or implied
   warranty.  In no event will the authors be held liable for any damages
--- a/NAS2D/ContainerUtils.h
+++ b/NAS2D/ContainerUtils.h
@@ -1,6 +1,6 @@
 // ==================================================================================
 // = NAS2D
-// = Copyright © 2008 - 2020 New Age Software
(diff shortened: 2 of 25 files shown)


cugone:
Confirmed, they display correctly even if the project tells the compiler to Use Multi-byte Characterset.

DanRStevens:
Thank you.

cugone:
I did notice that even though it displays correctly in visual studio, GitKraken does not and they show up as the questionmark glyph. It's probably a setting that I need to change somewhere.

DanRStevens:
Do you mean the old encoding, or the new encoding?

https://superuser.com/questions/1405232/how-to-configure-gitkraken-to-iso-8859-1-encoding
> For the time being Gitkraken supports only UTF-8

cugone:
The new encoding shows up as the question mark in GitKraken file views. I found that question as well. The poster specifically said the problem appears as stashing/popping changes the encoding and another question on SO used hunk staging and they said that's that what caused it. I do not use hunk staging and instead use line-by-line, selected lines, or whole-file staging and simple tests did not reproduce the issue and the glyph correctly rendered in visual studio and did not change the glyph between saves, stages in and out, stashing/popping, and push/pulls; so, I wouldn't worry about it until it actually happens.

PR #925, opened 2021-07-15
Define parsing of negative values to unsigned

DanRStevens:
Define parsing of negative values to `unsigned` types for `stringTo`.

In particular, this allows `"-1"` to be parsed as the maximum possible unsigned integer value for the given `unsigned` type. It also allows wraparound of large positive values too. That may be expected for `unsigned` values, which have well define wraparound, as per the C++ standard.

Related, is the definition of [`std::stoul`](https://en.cppreference.com/w/cpp/string/basic_string/stoul), which is used for parsing the smaller `unsigned` values:

> If the minus sign was part of the input sequence, the numeric value calculated from the sequence of digits is negated as if by unary minus in the result type, which applies unsigned integer wraparound rules.

Fixes the bug found after merging PR #920.

### DIFF
--- a/NAS2D/StringUtils.h
+++ b/NAS2D/StringUtils.h
@@ -71,11 +71,7 @@ namespace NAS2D
 			return static_cast<T>(integerValue);
 		} else if constexpr (std::is_integral_v<T> && std::is_unsigned_v<T>) {
 			const auto integerValue = std::stoul(value);
-			if constexpr (std::numeric_limits<unsigned long>::max() > std::numeric_limits<T>::max()) {
-				if (integerValue > std::numeric_limits<T>::max()) {
-					throw std::out_of_range("Value out of range: " + value);
-				}
-			}
+			// Out of range values have well defined wraparound for unsigned values
 			return static_cast<T>(integerValue);
 		} else {
 			static_assert(true, "Unsupported type");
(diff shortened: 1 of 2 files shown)


DanRStevens:
Could probably use a second set of eyes on this one to check if this is even a good idea.

cugone:
> 
> Could probably use a second set of eyes on this one to check if this is even a good idea.

Hrm. :/ It depends. If you are intending for "negative one means infinite", and that's the only reason to support negative unsigned values, I wouldn't do it. Instead explicitly document the use, but parse it as a regular `int` and detect the use of `-1` explicitly but convert it to a "large" number that's likely to "take forever" or "never finish" before the end of the program.

DanRStevens:
Given the finding in #935, I think I'm going to close this and explore another option.

ldicker83:
So I see that this is closed but will comment anyway: I originally intended -1 to be a sentinel but later opted to just stick with 0 since that seemed simpler. Plus, why would someone want a 0-length delay?

Anyway, that's the direction I went with later sprite definitions. Will need to go over them to be sure they're set to '0' and not -1 but yeah, '0' should be the way to go forward with this.

DanRStevens:
Huh, I just checked. You did have one data asset that used the value `"0"`. Though that wouldn't have worked with the previous code.

```
grep -rn 'delay="0"'
structures/warehouse.sprite:11:		<frame sheetid="main" delay="0" x="0" y="0" width="128" height="128" anchorx="0" anchory="64" />
```

Anyway, with the updates I applied, that value would now work as expected. Of course, now that the default when not specified is `"0"`, that particular value has now been removed. I figure use the defaults when possible to simplify things.

PR #926, opened 2021-07-15
Refactor XML iteration for array values

DanRStevens:
Reference: #797

Only iterate over elements of desired type, simplifying checks within the loop.

### DIFF
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -184,10 +184,8 @@ namespace
 	{
 		std::map<std::string, std::vector<AnimationSet::Frame>> actions;
 
-		for (const auto* action = element->firstChildElement(); action; action = action->nextSiblingElement())
+		for (const auto* action = element->firstChildElement("action"); action; action = action->nextSiblingElement("action"))
 		{
-			if (action->value() == "action")
-			{
 				const auto dictionary = attributesToDictionary(*action);
 				const auto actionName = dictionary.get("name");
 
@@ -201,7 +199,6 @@ namespace
 				}
 
 				actions[actionName] = processFrames(imageSheetMap, actionName, action, imageCache);
-			}
 		}
 
 		return actions;
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -145,28 +145,28 @@ namespace
 
 		for (const auto* node = element->firstChildElement("imagesheet"); node; node = node->nextSiblingElement("imagesheet"))
 		{
-				const auto dictionary = attributesToDictionary(*node);
-				const auto id = dictionary.get("id");
-				const auto src = dictionary.get("src");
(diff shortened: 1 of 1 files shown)


cugone:
In regards to your last commit, It might be helpful to add a validator function that will, at the very least, print to a log file when an expected element or attribute is missing or an unexpected element is found. (I prefer required-but-missing elements fail loudly with a crash. Read up on the [Aliens: Colonial Marines fix](https://www.polygon.com/2018/7/15/17574248/aliens-colonial-marines-fixing-code-typo-ai-xenomorphs) as to why XML/script errors should fail loudly in debug builds.)

See my [ValidateXMLElement](https://github.com/cugone/Abrams2019/blob/f36fd6d7f23c773876532df09b695880610da82d/Engine/Code/Engine/Core/DataUtils.cpp#L15) function for reference. It's extremely fast at an average of 10us and is run every "level" of XML elements:

Given:
```xml
<sprite version="0.99">
	<imagesheet id="main" src="robodigger.png" />
	<action name="running">
		<frame sheetid="main" delay="-1" x="0" y="0" width="128" height="128" anchorx="0" anchory="64" />
	</action>
</sprite>
```

```cpp
//...Load tinyxml2::XMLDoc for .sprite file...

ValidateXMLElement(root_element, "sprite", "imagesheet", "version", "action", "");

//Load data from sprite tag as it is guaranteed to exist.

//Parse imagesheet child element...

ValidateXMLElement(sheet_element, "imagesheet", "", "id,src", "", "");

//Load data from imagesheet tag...

//Parse any actions...
for(/*...*/) {
    ValidateXMLElement(action_element, "action", "frame", "name", "", "");
    //Load data for action (create action with no frames)
    //For each frame of the action...
    for(/*...*/) {
        ValidateXMLElement(frame_element, "frame", "", "sheetid,delay,x,y,width,height,anchorx,anchory", "", "");
        //Load frames into parent action...
    }
}
```

Normally, I'd put this in a constructor: `SpriteSheet(XMLElement& element)` but this is just an example.

That said, in a case like this, any elements or attributes that are missing will crash the program. In all cases, if anything extra appears it acts like real XML/HTML: Ignore it in release builds, log it in debug builds

DanRStevens:
I would generally agree with validating assumptions, and warning or throwing errors for unexpected data. Though it is a bit of a balancing act. Too much validation can hurt extensibility and backwards compatibility.

I think what bothered me about this code was the inconsistency between functions. I would like to eventually write an `XmlElement` array parsing helper, which could help make the processing more consistent. It may make sense to add warning or validation code there. Though we may need to be careful, as it might not be desirable in all cases.

Anyway, ParserHelper.h has a couple of helper methods:
```cpp
void reportMissingOrUnexpected(const std::vector<std::string>& missing, const std::vector<std::string>& unexpected);
void reportMissingOrUnexpected(const std::vector<std::string>& names, const std::vector<std::string>& required, const std::vector<std::string>& optional);
```

I've been using those helper methods for `Dictionary` keys parsed from `XmlElement` attributes. So far I haven't applied it to `XmlElement` values.

PR #935, opened 2021-07-18
Revert "Remove static_cast in AnimationSet data loading"

DanRStevens:
Reference: #920

This reverts commit 23745702140209d65aa5cb0c14ad2e6330e2344c.

Revert change that lead to a bug where "-1" is used as a sentinel value for frame `delay` and wasn't being handled correctly.


DanRStevens:
I may make more sense to use a sentinel value of `0` rather than `-1` when there is no `frameDelay`. A `frameDelay` of `0` wouldn't already have a meaning.

Current definition:
```cpp
const auto FRAME_PAUSE = unsigned(-1);
```

Existing code:
```cpp
	if (!mPaused && (frame.frameDelay != FRAME_PAUSE))
	{
		while (frame.frameDelay > 0 && mTimer.accumulator() >= frame.frameDelay)
		{
			// ...
```

Actually, using `0` for the sentinel value might actually simplify the code.

PR #950, opened 2021-07-25
Removes comments for time argument for Mixer Fade functions

cugone:
Removes low-value comments for time argument for Fade functions due to explicit type for milliseconds.

### DIFF
--- a/NAS2D/Mixer/Mixer.h
+++ b/NAS2D/Mixer/Mixer.h
@@ -56,15 +56,9 @@ namespace NAS2D
 		 *
 		 * \param music Reference to a Music Resource.
 		 * \param loops Repeat count. -1 for continuous loop.
-		 * \param time Length of the fade in milliseconds. Default is 500.
 		 */
 		virtual void fadeInMusic(const Music& music, int loops = Mixer::CONTINUOUS, std::chrono::milliseconds time = Mixer::DEFAULT_FADE_TIME) = 0;
 
-		/**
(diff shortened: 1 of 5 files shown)


cugone:
@DanRStevens This shouldn't conflict with #951

DanRStevens:
Could you please either rebase this branch on master or merge master into this branch? I'm seeing a test failure, though I think it's the problem that was just fixed by merging #951.

NAS2D/Mixer/Mixer.h
```diff
@@ -56,15 +56,9 @@ namespace NAS2D
 		 *
 		 * \param music Reference to a Music Resource.
 		 * \param loops Repeat count. -1 for continuous loop.
-		 * \param time Length of the fade in milliseconds. Default is 500.
```
DanRStevens, line 59: I'm a little bit concerned that documenting some but not all parameters could trigger a warning/error from some of the documentation checks. Would be nice to get all CI builds to run and see if they complain.

cugone: Would removing the entire comment block be better?

DanRStevens: Possibly. I feel a little on the fence on this one.

I think maybe the parameter should be renamed to something like `fadeTime`. Then I'd probably be comfortable removing the documentation.

DanRStevens (approved):

PR #952, opened 2021-07-25
Vcpkg safe project integration

cugone:
Ref: #373 (deletable?)

Adds instructions to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ NAS2D  is licensed under the zlib license. See LICENSE.txt for details.
 The NAS2D project uses Vcpkg to integrate any dependencies without explicitly adding the source to the project directly. If you do not want to install vcpkg in a user-wide context, vcpkg allows for a local NuGet package to be created. The following steps are specific to NAS2D:
 
 1. Clone `vcpkg`
-1. Run the command `./bootstrap-vcpkg.bat`
-1. Run the command `vcpkg integrate project`
-1. Copy the generated NuGet install command.
-1. Open Visual Studio's Package Manager Console (Tools > NuGet Package Manager > Package Manager Console)
-1. Paste and Run the generated NuGet install command.
-1. Build the NAS2D solution to install all dependencies.
+2. Run the command `./bootstrap-vcpkg.bat`
+3. Run the command `vcpkg integrate project`
+4. Copy the generated NuGet install command.
+5. Open Visual Studio's Package Manager Console (Tools > NuGet Package Manager > Package Manager Console)
+6. Paste and Run the generated NuGet install command.
+7. Build the NAS2D solution to install all dependencies.
 
 From this point any missing headers should be found and future compilation will be successful.
 
@@ -49,10 +49,10 @@ Do not commit the NuGet-specific changes made to the project in `.vcxproj`, `.vc
 A workaround to allow for successful compilation or to work with the project file when adding or removing files or filters is as follows:
 
 1. Stash the NuGet-specific changes with an appropriate name.
-1. Complete any work, make any commits.
-1. Pop/Unstash the NuGet-changes.
-1. Compile.
-1. Re-stash the NuGet-changes.
+2. Complete any work, make any commits.
+3. Pop/Unstash the NuGet-changes.
+4. Compile.
+5. Re-stash the NuGet-changes.
 
 For best results, these steps should be completed immediately after installing the local NuGet package so as to not include any unrelated changes.
 
--- a/README.md
+++ b/README.md
@@ -32,13 +32,13 @@ NAS2D  is licensed under the zlib license. See LICENSE.txt for details.
 
 The NAS2D project uses Vcpkg to integrate any dependencies without explicitly adding the source to the project directly. If you do not want to install vcpkg in a user-wide context, vcpkg allows for a local NuGet package to be created. The following steps are specific to NAS2D:
 
-1. Clone `vcpkg`
-2. Run the command `./bootstrap-vcpkg.bat`
-3. Run the command `vcpkg integrate project`
-4. Copy the generated NuGet install command.
-5. Open Visual Studio's Package Manager Console (Tools > NuGet Package Manager > Package Manager Console)
-6. Paste and Run the generated NuGet install command.
-7. Build the NAS2D solution to install all dependencies.
+1.  Clone `vcpkg`
+2.  Run the command `./bootstrap-vcpkg.bat`
+3.  Run the command `vcpkg integrate project`
+4.  Copy the generated NuGet install command.
+5.  Open Visual Studio's Package Manager Console (Tools > NuGet Package Manager > Package Manager Console)
+6.  Paste and Run the generated NuGet install command.
+7.  Build the NAS2D solution to install all dependencies.
 
 From this point any missing headers should be found and future compilation will be successful.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,9 +42,9 @@ The NAS2D project uses Vcpkg to integrate any dependencies without explicitly ad
 
 From this point any missing headers should be found and future compilation will be successful.
 
-**WARNING**
-
-Do not commit the NuGet-specific changes made to the project in `.vcxproj`, `.vcxproj.filters`, `packages.config`, or the folder `./packages/...`. In addition to these changes being local to your system, they may contain potentially sensitive or personally identifiable information in the generated absolute path!
+**WARNING** |
+------------ |
+Do not commit the NuGet-specific changes made to the project in `.vcxproj`, `.vcxproj.filters`, `packages.config`, or the folder `./packages/...`. In addition to these changes being local to your system, they may contain potentially sensitive or personally identifiable information in the generated absolute path! |
```
DanRStevens, line 47: I'm not too sure why this is being done. I've tried checking Codacy, but it seems to redirect me to the Lairworks account, which doesn't show your branch.

I get the sense maybe a title would make more sense than a table.

README.md
```diff
@@ -25,3 +25,37 @@ Officially, *NAS2D* is supported on Windows (Vista+) and MacOS X (10.8+). Binari
 ## License
 
 NAS2D  is licensed under the zlib license. See LICENSE.txt for details.
+
+---
+
+## Vcpkg integration
+
+The NAS2D project uses Vcpkg to integrate any dependencies without explicitly adding the source to the project directly. If you do not want to install vcpkg in a user-wide context, vcpkg allows for a local NuGet package to be created. The following steps are specific to NAS2D:
```
DanRStevens, line 33: I feel like most people may want to just use a system wide install, and so instructions for that could be placed first. Perhaps also include instructions to remove the system wide install in case it causes problems for them.

README.md
```diff
@@ -25,3 +25,37 @@ Officially, *NAS2D* is supported on Windows (Vista+) and MacOS X (10.8+). Binari
...
+------------ |
+Do not commit the NuGet-specific changes made to the project in `.vcxproj`, `.vcxproj.filters`, `packages.config`, or the folder `./packages/...`. In addition to these changes being local to your system, they may contain potentially sensitive or personally identifiable information in the generated absolute path! |
+
+A workaround to allow for successful compilation or to work with the project file when adding or removing files or filters is as follows:
+
+1.  Stash the NuGet-specific changes with an appropriate name.
+2.  Complete any work, make any commits.
+3.  Pop/Unstash the NuGet-changes.
+4.  Compile.
+5.  Re-stash the NuGet-changes.
```
DanRStevens, line 55: I really don't like this. I've done this sort of thing in the past over the short term. It gets really tedious fast. Plus it's really easy to accidentally commit those changes by accident. Further, this interferes with being able to use the `git stash` command. You end up burying the changes that you don't want to check in, but want to use locally to build.

I think a better solution is to add extra `.gitignore` entries, and migrate those settings out of the main project file and into the ignored files. The ignored files can be conditionally included from the main project file. That would avoid breaking things for people who don't use the local settings file. The setup steps can then detail how to integrate `vcpkg` with the project and then move that config to the separate ignored file.

DanRStevens (changes requested):
I appreciate that progress has been made on this, and that you have a short term solution, though I'd like to see a few changes before I'd be comfortable merging this.

Also, I suspect with a bit of work, there would be much less need for some of those ReadMe steps and warnings.

README.md
DanRStevens: Adding link to relevant comment:
https://github.com/lairworks/nas2d-core/issues/373#issuecomment-886141847

> Ignoring the `packages` folder would be easy enough with `.gitignore`. Though handling the project file modifications would be more difficult. Perhaps you could conditionally load project settings from another file, and have that other file listed in `.gitignore`. I'm pretty sure the project file syntax allows for conditionally loading settings from another file. Pretty sure that support is already used with some of the system files for the Visual Studio project settings, and to handle plugins, or user specific settings.
> 
> Checking the project file, I see external settings being conditionally imported with:
> 
> ```xml
> <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
> ```
> Could do similar with a file loaded from the project folder. I assume paths would by default be relative to the project folder.
>

README.md
cugone: Codacy throws a hissy fit if you try to start a section via bolded text instead of a heading. I wanted something like a Warning Box; since they aren't supported, a table seemed an appropriate workaround.

README.md
cugone: After some real usage, unstashing the changes is not required and things still compile fine, even between branches.

DanRStevens: I feel like this needs to be understood better. How are the dependencies being accessed? Are files being left over from previous builds? If a full clean and recompile is done, would it have trouble finding dependent libraries? This may related to https://github.com/OutpostUniverse/OPHD/issues/1053.

README.md
DanRStevens: I'm maybe mildly uncomfortable with this.

I did a bit of reading, and found this interesting:
https://stackoverflow.com/questions/50544499/how-to-make-a-styled-markdown-admonition-box-in-a-github-gist
> Use emoji to call the user attention
> > :warning: **If you are using mobile browser**: Be very careful here!

There they used an emoji and block quotes to set the text apart with a box like area.

----

Of course, this is from the perspective of trying to get a layout you want stuffed into Markdown, when really it might be better to adapt what you want to fit what Markdown offers.

To me, this seems a bit like a subheading note.

----

At any rate, I'm still wondering if the process can somehow be improved so that warning becomes unnecessary, or more of a minor step.

README.md
DanRStevens: Curious if you ever got around to testing this?

cugone: Related to that, I was reading up on [the vcpkg integration section about open source projects](https://github.com/microsoft/vcpkg/blob/master/docs/users/integration.md#user-wide-for-msbuild-recommended-for-open-source-msbuild-projects) and it mentions that, some libraries, even using `vcpkg integrate install` do not play well and must be manually linked and it specifically calls out SDL2 as being one of the offenders.

DanRStevens: I think there's a misunderstanding there.

The SDL2 project compiles into two separate libraries: `SDL2`, `SDL2main`. The `SDL2` library is the main one, and will be linked automatically. The `SDL2main` library is only appropriate when using the SDL2 provided `main` function, and should not be linked when that is disabled. For that reason, `SDL2main` is not automatically linked. You must manually specify that in the project settings Additional Libraries section.

I've noticed Google Test does similar, and also splits the library provided `main` method into a separate library, so downstream project can control if they want to link it or not. If you check the NAS2D project for the unit test project file, you'll see it has `AdditionalDependencies` set, which includes `gtest.lib` and `gtest_main.lib`. For that project, both libraries appear to be excluded from automatic linking.

That matches with the description in the article you linked.

DanRStevens: To add to the previous bit, check the OPHD project file, and you'll see that it manually adds a dependency for `SDL2main` (but not for `SDL2`).
```xml
<AdditionalDependencies>NAS2D.lib;opengl32.lib;sdl2main.lib;%(AdditionalDependencies)</AdditionalDependencies>
```

DanRStevens:
Closing this as the `vcpkg` upgrades used in #1109 handles many of the concerns.

PR #955, opened 2021-07-25
Refactors fadeIn/Out argument type to std::chrono::milliseconds

### LINKED ISSUE #954: Inconsistency between Mixer and Render fade types
The `Mixer` fade speed is now specified with type `std::chrono::milliseconds`. However the `Renderer` fade methods were not also updated to use this type. This creates awkwardness in downstream OPHD, since it uses the same constant for both `Music` and `Renderer` fades.

We should update these types to match.

cugone:
Closes #954

### DIFF
--- a/NAS2D/Renderer/Renderer.cpp
+++ b/NAS2D/Renderer/Renderer.cpp
@@ -83,12 +83,12 @@ void Renderer::fadeColor(Color color)
 /**
  * Non-blocking screen fade.
  *
- * \param	delay	Time in miliseconds the fade should last. A value of 0
+ * \param	fadeInTime	Time in miliseconds the fade should last. A value of 0
  *					will instantly fade the screen in.
  */
-void Renderer::fadeIn(float delay)
+void Renderer::fadeIn(std::chrono::milliseconds fadeInTime)
 {
-	if (delay == 0)
+	if (fadeInTime.count() == 0LL)
 	{
 		mCurrentFade = 0.0f;
 		mCurrentFadeType = FadeType::None;
@@ -96,7 +96,7 @@ void Renderer::fadeIn(float delay)
 	}
 
 	mCurrentFadeType = FadeType::In;
-	mFadeStep = 255.0f / delay;
+	mFadeStep = 255.0f / static_cast<float>(fadeInTime.count());
 
 	fadeTimer.delta(); // clear timer
 }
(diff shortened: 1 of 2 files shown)


DanRStevens (approved):
Changes look good. Thank you for taking care of that.

----

On a related matter, there is also a `Fade` class, which attempted to extract the `fadeIn` and `fadeOut` functionality from the `Renderer` to a separate class. It should probably be updated too. We can handle that as a separate lower priority change though, since the new extracted class isn't really used outside of experimental branches with unfinished conversions.

Another possibly helpful change might be to set a default fade time of 0 ms. That would help avoid awkward `std::chrono::milliseconds` constructions in downstream projects when they want to jump instantly from one fade setting to another.

cugone:
> 
> Another possibly helpful change might be to set a default fade time of 0 ms. That would help avoid awkward `std::chrono::milliseconds` constructions in downstream projects when they want to jump instantly from one fade setting to another.

Instead of defaulting to zero, maybe [remove the default entirely](https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/)? Fades are generally situation-specific and having a default that is almost never the same between scenes forces the designer to specify one anyway.

DanRStevens:
The article on default arguments brings up some interesting points, many of which I've found to be true myself, though I disagree with his final conclusion about never using default arguments.

There are certainly cases when function overloads would work better than a default argument. I was even thinking of making such an update in `Filesystem`. Though I still feel default arguments have a valid place. In terms of overloads versus defaults, I'd simply suggest to avoid mixing them. In most cases you usually want one or the other. By choosing either one or the other, you avoid the more complicated cases of mixing overloads with defaults.

The boolean tarpit example has a whole other underlying problem, even before you consider default parameters. Even the author of the article admits it:
> Default function arguments do not directly cause boolean parameter tarpits, but they make their pernicious effects worse

There are very rare cases where boolean parameters are a good idea. The main example being a setter method for a boolean value. The name of the function should make the meaning of the parameter quite clear. Outside of that though, particularly when there are multiple parameters, you should avoid boolean arguments. They just aren't self documenting at the call site. It's pretty much impossible to get much context out of `true` or `false`, especially when there are many parameters. In that case, I would recommend either replacing the boolean arguments with enums, where the enum value names would provide proper context as to their meaning, or designing separate functions where the meaning of the would-be boolean parameters is encoded in the function name. Note that such functions are then not overloads, since they end up with different names.

For redeclaration rules, I would generally suggest not including default values in the implementation file (not even as a comment), when the default value is specified in a header file. Considering default values are evaluated at the call site, it makes sense to not put it with the implementation. And use header files to avoid redeclarations. Of course that's not to say you can never use default values in an implementation file. Default values might still make sense in implementation files for file local helper methods, such as those declared in unnamed namespaces, which might not have a forward declaration.

The virtual function example was interesting. I don't find I typically use default values with virtual methods though.

Anyway, most of the rest of the examples was about cases where you can't use default parameters. So yeah, obvious don't use them there. :wink:

----

As for the `Renderer` fade functions, non-zero values are quite arbitrary, so I would avoid specifying a non-zero default. Though zero has a special meaning of instantly jump from one state to another, without any visible fade effect. I think that makes good sense as a default if no fade length value is specified.

Currently the code for instant fade in OPHD is:
```cpp
renderer.fadeOut(std::chrono::milliseconds{0});
```

Anyway, I'm not overly attached to default parameters here or not.

PR #958, opened 2021-07-26
Documentation pass on Renderer folder

cugone:
See: #861 and #899

### DIFF
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -30,7 +30,6 @@ SignalSource<>& Fade::fadeComplete()
 }
 
 
-// Fade in from fadeColor
 void Fade::fadeIn(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
@@ -38,7 +37,6 @@ void Fade::fadeIn(unsigned int durationInMilliseconds)
 }
 
 
-// Fade out to fadeColor
 void Fade::fadeOut(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -30,6 +30,7 @@ SignalSource<>& Fade::fadeComplete()
 }
 
 
+// Fade in from fadeColor
 void Fade::fadeIn(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
@@ -37,6 +38,7 @@ void Fade::fadeIn(unsigned int durationInMilliseconds)
 }
 
 
+// Fade out to fadeColor
 void Fade::fadeOut(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
--- a/NAS2D/Renderer/Rectangle.h
+++ b/NAS2D/Renderer/Rectangle.h
@@ -135,8 +135,8 @@ namespace NAS2D
 			return startPoint() <= point && point < endPoint();
 		}
 
-		// Start point exclusive (x, y), endpoint exclusive (x + width, y + height)
-		// Area in interval notation: (x .. x + width), (y .. y + height)
+		// Start point inclusive (x, y), endpoint exclusive (x + width, y + height)
+		// Area in interval notation: [x .. x + width), [y .. y + height)
 		constexpr bool overlaps(const Rectangle& rect) const
 		{
 			return startPoint() < rect.endPoint() && rect.startPoint() < endPoint();
--- a/NAS2D/Renderer/Rectangle.h
+++ b/NAS2D/Renderer/Rectangle.h
@@ -24,13 +24,11 @@ namespace NAS2D
 		BaseType width = 0;
 		BaseType height = 0;
 
-		// Factory method
 		constexpr static Rectangle<BaseType> Create(Point<BaseType> startPoint, Vector<BaseType> size)
 		{
 			return {startPoint.x, startPoint.y, size.x, size.y};
 		}
 
-		// Factory method
 		constexpr static Rectangle<BaseType> Create(Point<BaseType> startPoint, Point<BaseType> endPoint)
 		{
 			return Create(startPoint, endPoint - startPoint);
@@ -137,8 +135,8 @@ namespace NAS2D
 			return startPoint() <= point && point < endPoint();
 		}
 
-		// Start point inclusive (x, y), endpoint exclusive (x + width, y + height)
-		// Area in interval notation: [x .. x + width), [y .. y + height)
+		// Start point exclusive (x, y), endpoint exclusive (x + width, y + height)
+		// Area in interval notation: (x .. x + width), (y .. y + height)
 		constexpr bool overlaps(const Rectangle& rect) const
 		{
 			return startPoint() < rect.endPoint() && rect.startPoint() < endPoint();
@@ -153,4 +151,4 @@ namespace NAS2D
 
 	template <typename BaseType>
 	Rectangle(BaseType, BaseType, BaseType, BaseType) -> Rectangle<BaseType>;
-} // namespace
+} // namespace NAS2D
--- a/NAS2D/Renderer/Renderer.cpp
+++ b/NAS2D/Renderer/Renderer.cpp
@@ -77,6 +77,12 @@ void Renderer::fadeIn(std::chrono::milliseconds fadeInTime)
 }
 
 
+/**
+ * Non-blocking screen fade.
+ *
+ * \param	fadeOutTime	Length of time the fade should last.
+ *			A value of 0 will instantly fade the screen out.
+*/
 void Renderer::fadeOut(std::chrono::milliseconds fadeOutTime)
 {
 	if (fadeOutTime.count() == 0LL)
--- a/NAS2D/Renderer/Renderer.cpp
+++ b/NAS2D/Renderer/Renderer.cpp
@@ -18,11 +18,6 @@
 using namespace NAS2D;
 
 
-/**
- * Internal constructor used by derived types to set the name of the Renderer.
- *
- * This c'tor is not public and can't be invoked externally.
- */
 Renderer::Renderer(const std::string& appTitle) : mTitle(appTitle)
 {}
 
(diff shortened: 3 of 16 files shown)


NAS2D/Renderer/Fade.cpp
```diff
@@ -30,15 +30,13 @@ SignalSource<>& Fade::fadeComplete()
...
 
-// Fade in from fadeColor
 void Fade::fadeIn(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
 	mDirection = FadeDirection::In;
 }
 
 
-// Fade out to fadeColor
```
DanRStevens, line 41: I guess technically the `fadeColor` is maybe not totally obvious from looking at just this one function. That might be documentation worth keeping.

NAS2D/Renderer/Rectangle.h
```diff
@@ -137,8 +135,8 @@ namespace NAS2D
 			return startPoint() <= point && point < endPoint();
 		}
 
-		// Start point inclusive (x, y), endpoint exclusive (x + width, y + height)
-		// Area in interval notation: [x .. x + width), [y .. y + height)
+		// Start point exclusive (x, y), endpoint exclusive (x + width, y + height)
+		// Area in interval notation: (x .. x + width), (y .. y + height)
```
DanRStevens, line 139: This change is wrong. The original documentation was using [interval notation](https://brilliant.org/wiki/interval-notation/), where square and round brackets have different meaning. Square brackets are inclusive, while round brackets are exclusive.

The choice of inclusive/exclusive endpoints was to make it easy to lay out covering rectangles with no overlap and no space in between. This typically matches how rectangles are used. Consider a rectangle representing screen coordinates for a 640x480 display. The point `{0, 0}` could be the top left pixel, while `{640, 480}` is off the screen.

NAS2D/Renderer/Renderer.cpp
```diff
@@ -69,23 +55,12 @@ void Renderer::drawTextShadow(const Font& font, std::string_view text, Point<flo
...
 {
 	mFadeColor = color;
 }
 
 
-/**
- * Non-blocking screen fade.
- *
- * \param	fadeInTime	Time in miliseconds the fade should last. A value of 0
- *					will instantly fade the screen in.
```
DanRStevens, line 87: I'm on the fence about the value of this comment.

NAS2D/Renderer/Renderer.cpp
```diff
@@ -102,12 +77,6 @@ void Renderer::fadeIn(std::chrono::milliseconds fadeInTime)
 }
 
 
-/**
- * Non-blocking screen fade.
- *
- * \param	fadeOutTime	Time in miliseconds the fade should last. A value of 0
- *					will instantly fade the screen in.
```
DanRStevens, line 109: Ditto.

NAS2D/Renderer/Renderer.cpp
```diff
@@ -124,48 +93,30 @@ void Renderer::fadeOut(std::chrono::milliseconds fadeOutTime)
...
 {
 	return Point{0, 0} + mResolution / 2;
 }
 
 
-/**
- * Updates the screen.
- *
- * \note	All derived Renderer objects must call Renderer::update()
- *			before performing screen refreshes.
```
DanRStevens, line 167: I'm a bit uncertain about the value of this comment. Or maybe it's because I'm uncertain about the design. Perhaps this relates to the work to extract fading into a separate control, which was never quite completed.

DanRStevens (changes requested):
Mostly good. A few small concerns, and one big issue.

NAS2D/Renderer/Rectangle.h
cugone: I changed the comments to match what the code was doing. `overlaps` was written to be exclusive: `startPoint() < endPoint()`. vs `contains` which was written as inclusive: `startPoint <= endPoint()`.

DanRStevens: Check the unit tests. The examples may help.

Both functions use the same criteria. The difference is that `contains` works with `Point` objects, which have no area, whereas `overlaps` works with `Rectangle` objects, which do have area. Comparisons with `endPoint()` use `<` for both of them.

DanRStevens: And for `overlaps`, pay attention to how the two different `Rectangle` objects are used. The `startPoint()` of each is compared against the `endPoint()` of the other.

NAS2D/Renderer/Renderer.cpp
cugone: If you aren't sure whether or not it's a bad comment, it probably is. :P

DanRStevens: From another perspective, it we're uncertain if something should be changed, it probably shouldn't be changed. :wink:

NAS2D/Renderer/Fade.cpp
cugone: It was not obvious that comment had anything to do with the code written. Nowhere in the function name, signature, or body is `color ` ever used. Hence, the comment got removed.

DanRStevens: Indeed, though the fade does involve a color. Such as fade to white or fade to black. It's because it's not anywhere in the function that it needs to be documented. It's the `fadeColor` method that sets that aspect of fading.

cugone: That's a different function. Long-form documentation can (should) be written to relate the two (docs.microsoft does this all the time). Trying to do so in the comments is just adding noise.

NAS2D/Renderer/Renderer.cpp
cugone: Asking "Do we need this?" or "Is this useful?" and if you can't come up with a good reason immediately, it likely can be cut.

In this case, the comment restated what already could be explicitly stated in the function and argument names. So I delete it.

DanRStevens: The "non-blocking" part might not be obvious. The treatment of `0` and whether or not it's a valid value might not be obvious.

cugone: Could you explain what "instantly fade" means? does that mean there is no fade at all or it cuts to the fade color for one frame?

NAS2D/Renderer/Renderer.cpp
cugone: The fade functions are most-definitely blocking. They just set up variables to be updated in `Renderer::Update` which runs on the main thread:

```cpp
void Renderer::update()
{
	if (mCurrentFadeType != FadeType::None)
	{
		float fade = (fadeTimer.delta() * mFadeStep) * static_cast<int>(mCurrentFadeType);

		mCurrentFade += fade;

		if (mCurrentFade < 0.0f || mCurrentFade > 255.0f)
		{
			mCurrentFade = std::clamp(mCurrentFade, 0.0f, 255.0f);
			mCurrentFadeType = FadeType::None;
			fadeCompleteSignal();
		}
	}

	if (mCurrentFade > 0.0f)
	{
		drawBoxFilled(Rectangle<float>::Create({0, 0}, size().to<float>()), mFadeColor.alphaFade(static_cast<uint8_t>(mCurrentFade)));
	}
}
```

As you can see, no where is a thread used or synchronized, nor is the value of the fade time passed to any thread. Therefore, the "Non-blocking screen fade" comments are at best out-of-date and at worst misleading. Together with the use of properly typed and named variables the entire block can be removed.

NAS2D/Renderer/Renderer.cpp
DanRStevens: This is part of why I want to defer on making edits to this comment. Maybe we should worry about the low hanging fruit first, where we're all in agreement that a comment has no real value and should be removed. That way we could get the PR merged more quickly, and don't have a large set of changes hanging around unmerged, and potentially accumulating merge conflicts.

For more contentious parts we can either discuss in the issue (and leave the issue open), create a new issue to deal with that (and close the old one after the bulk of changes are merged), or maybe even just ignore it if other future work is likely to make the point obsolete. For instance, there is some work to extract the fade ability to a separate control, which may or may not make some of these discussions obsolete.

NAS2D/Renderer/Renderer.cpp
DanRStevens: I think we have a disagreement over what "non-blocking" means.

Nowhere in that function is there any code that would prevent it from returning for an unspecified amount of time. It does it's thing, and then returns. There is no waiting.

cugone: I was conflating "non-blocking" with "asynchronous". Yes, these are "non-blocking", but so is any function by default. Specifically commenting this one as non-blocking is superfluous.

DanRStevens: Well, not exactly. Many function calls are blocking. A file read method is typically blocking, as disk IO may take arbitrarily long, particularly with disk read errors and recovery.

A fade method very well could be blocking, if it chose to do the entire fade before returning, along with timing delays. In this case, it merely sets up the parameters to do the fade, which then takes place during the `Renderer::update` method, and those updates are only one frame per call, with no built in timing delay.

We might debate the value of the comment, though I wouldn't exactly dismiss it as entirely superfluous.

NAS2D/Renderer/Fade.cpp
DanRStevens: I feel like this was never really addressed.

cugone: I was in the middle of a small refactor that combined `fadeColor` and `fadeIn/Out` into `fade`:

```cpp
void fadeIn(Color fadeColor, std::chrono::milliseconds fadeInTime) {
//...
}
void fadeOut(Color fadeColor, std::chrono::milliseconds fadeOutTime) {
//...
}
```

DanRStevens: That's not a documentation change though. If you're making such a change on another branch, then you should probably avoiding changing the documentation here preemptively.

NAS2D/Renderer/Fade.cpp
```diff
@@ -29,15 +29,13 @@ SignalSource<>& Fade::fadeComplete()
...
 // Fade in from fadeColor
 void Fade::fadeIn(unsigned int durationInMilliseconds)
 {
 	setDuration(durationInMilliseconds);
 	mDirection = FadeDirection::In;
 }
 
-
 // Fade out to fadeColor
 void Fade::fadeOut(unsigned int durationInMilliseconds)
```
DanRStevens, line 40: Inter-function spacing was changed here. That makes it inconsistent with the rest of the file.

cugone: Fixed

DanRStevens (approved):
Looks good

PR #960, opened 2021-07-30
Refactor Fade class mDuration to chrono::milliseconds

### LINKED ISSUE #957: Refactor Fade class duration type to std::chrono::milliseconds
See: [#955 comment](https://github.com/lairworks/nas2d-core/pull/955#pullrequestreview-714359646)

cugone:
Closes #957

### DIFF
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -31,7 +31,7 @@ SignalSource<>& Fade::fadeComplete()
 
 
 // Fade in from fadeColor
-void Fade::fadeIn(const std::chrono::milliseconds& fadeTime)
+void Fade::fadeIn(std::chrono::milliseconds fadeTime)
 {
 	setDuration(fadeTime);
 	mDirection = FadeDirection::In;
@@ -39,7 +39,7 @@ void Fade::fadeIn(const std::chrono::milliseconds& fadeTime)
 
 
 // Fade out to fadeColor
-void Fade::fadeOut(const std::chrono::milliseconds& fadeTime)
+void Fade::fadeOut(std::chrono::milliseconds fadeTime)
 {
 	setDuration(fadeTime);
 	mDirection = FadeDirection::Out;
@@ -86,7 +86,7 @@ void Fade::draw(Renderer& renderer) const
 }
 
 
-void Fade::setDuration(const std::chrono::milliseconds& newDuration)
+void Fade::setDuration(std::chrono::milliseconds newDuration)
 {
 	if (newDuration == newDuration.zero())
 	{
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -33,7 +33,7 @@ SignalSource<>& Fade::fadeComplete()
 // Fade in from fadeColor
 void Fade::fadeIn(const std::chrono::milliseconds& fadeTime)
 {
-	setDuration(durationInMilliseconds);
+	setDuration(fadeTime);
 	mDirection = FadeDirection::In;
 }
 
@@ -41,7 +41,7 @@ void Fade::fadeIn(const std::chrono::milliseconds& fadeTime)
 // Fade out to fadeColor
 void Fade::fadeOut(const std::chrono::milliseconds& fadeTime)
 {
-	setDuration(durationInMilliseconds);
+	setDuration(fadeTime);
 	mDirection = FadeDirection::Out;
 }
 
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -88,7 +88,7 @@ void Fade::draw(Renderer& renderer) const
 
 void Fade::setDuration(const std::chrono::milliseconds& newDuration)
 {
-	if (newDuration.count() <= 0LL)
+	if (newDuration == newDuration.zero())
 	{
 		throw std::runtime_error("Fade duration must be positive");
 	}
--- a/NAS2D/Renderer/Fade.cpp
+++ b/NAS2D/Renderer/Fade.cpp
@@ -31,7 +31,7 @@ SignalSource<>& Fade::fadeComplete()
 
 
 // Fade in from fadeColor
-void Fade::fadeIn(unsigned int durationInMilliseconds)
+void Fade::fadeIn(const std::chrono::milliseconds& fadeTime)
 {
 	setDuration(durationInMilliseconds);
 	mDirection = FadeDirection::In;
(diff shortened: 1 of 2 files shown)


NAS2D/Renderer/Fade.h
```diff
@@ -28,8 +29,8 @@ namespace NAS2D
 
 		SignalSource<>& fadeComplete();
 
-		void fadeIn(unsigned int durationInMilliseconds);
-		void fadeOut(unsigned int durationInMilliseconds);
+		void fadeIn(const std::chrono::milliseconds& fadeTime);
+		void fadeOut(const std::chrono::milliseconds& fadeTime);
```
DanRStevens, line 33: I believe it would be more consistent with existing code to remove the `const&`. Besides, this is basically a wrapped primitive type, so passing by `const&` would likely be less efficient than passing by value.

NAS2D/Renderer/Fade.h
```diff
@@ -38,7 +39,7 @@ namespace NAS2D
 		void draw(Renderer& renderer) const;
 
 	private:
-		void setDuration(unsigned int durationInMilliseconds);
+		void setDuration(const std::chrono::milliseconds& newDuration);
```
DanRStevens, line 42: Ditto.

NAS2D/Renderer/Fade.cpp
```diff
@@ -86,13 +86,13 @@ void Fade::draw(Renderer& renderer) const
 }
 
 
-void Fade::setDuration(unsigned int durationInMilliseconds)
+void Fade::setDuration(const std::chrono::milliseconds& newDuration)
 {
-	if (durationInMilliseconds == 0)
+	if (newDuration.count() <= 0LL)
```
DanRStevens, line 91: You can potentially use [`std::chrono::milliseconds::zero()`](https://en.cppreference.com/w/cpp/chrono/duration/zero) here. That would avoid having to convert to the underlying ticks to do the check.
```cpp
if (newDuration <= std::chrono::milliseconds::zero())
```

NAS2D/Renderer/Fade.cpp
```diff
@@ -65,7 +65,7 @@ void Fade::update()
 		return;
 	}
 
-	const auto step = static_cast<uint8_t>(std::clamp(mFadeTimer.accumulator() * 255u / mDuration, 0u, 255u));
+	const auto step = static_cast<uint8_t>(std::clamp(mFadeTimer.accumulator() * 255u / static_cast<unsigned int>(mDuration.count()), 0u, 255u));
```
DanRStevens, line 68: I feel like there should be a way to avoid this cast here. Particularly since the returned `mFadeTimer.accumulator()` value should be effectively the same type as `mDuration.count()`. Effectively we want to use the natural type of the expression.

NAS2D/Renderer/Fade.h
cugone: I'm going to argue against this. You should always use `const` whenever and wherever possible, particularly for non-built-in-types (even if you "know" the underlying type is "just" a primitive).

NAS2D/Renderer/Fade.cpp
cugone: The default type of `milliseconds` is `long long`, the cast was just me being paranoid with a potentially unsigned/signed mismatch.

cugone: Nope. Cast is required. Two errors pop up:

`A sub-expression may overflow before being assigned to a wider type` for the division and `no matching overload found for std::clamp arguments are long long, unsigned int, unsigned int`

NAS2D/Renderer/Fade.h
DanRStevens: This isn't so much about the `const` and it is about the `&`. Passing by reference adds overhead for small types. If you eliminate the `&`, there's also little point in maintaining the `const`. It's often inappropriate to have `const` on it's own in API signatures when passing or returning types by-value.

The whole point of passing by reference is efficiency. It makes no sense in cases that result in an efficiency loss.

----

You may have noticed in a link you provided just the other day that some compiler checks are disabled for primitives and small types that aren't at least twice the native word size.

https://docs.microsoft.com/en-us/cpp/code-quality/c26820?view=msvc-160

> This warning isn't raised for scalars, smart pointers, or views. It's also not raised for types whose size isn't more than twice the platform-dependent pointer size.

There's good reason why they don't warn in that case.

----

Rules shouldn't be blindly followed without understanding why they're there. Doing so leads to [Cargo cult programming](https://en.wikipedia.org/wiki/Cargo_cult_programming).

NAS2D/Renderer/Fade.cpp
DanRStevens: There may be other ways to address and avoid these problems. A `static_cast` might not be the only solution. Another consideration might be to change the `u` type suffix on the constants. That may not be so appropriate now if the other associated types have changed by the transition to `std::chrono::millisecond`.

It may also be possible to use an explicit type parameter when calling `std::clamp`.

NAS2D/Renderer/Fade.cpp
```diff
@@ -88,7 +88,7 @@ void Fade::draw(Renderer& renderer) const
 
 void Fade::setDuration(const std::chrono::milliseconds& newDuration)
 {
-	if (newDuration.count() <= 0LL)
+	if (newDuration == newDuration.zero())
```
DanRStevens: Hmm, I find this style a bit odd. The [`std::chrono::milliseconds::zero()`](https://en.cppreference.com/w/cpp/chrono/duration/zero) method is `static`, but here it's being called through an instance variable.

cugone: How would you feel about `decltype(newDuration)::zero()`? It allows for flexibility later if, for whatever reason, the duration type changes from milliseconds such that the code doesn't have to change.

DanRStevens: That had actually come to mind when reviewing the code. That would be acceptable. It's a little bit longer, but clear, and flexible.

NAS2D/Renderer/Fade.cpp
DanRStevens: After looking more closely, `mFadeTimer.accumulator()` returns an `unsigned int`, and is based on [`SDL_GetTicks`](https://wiki.libsdl.org/SDL_GetTicks), which return `Uint32`. Meanwhile `mDuration.count()` is calling [`std::chrono::millisecond::count()`](https://en.cppreference.com/w/cpp/chrono/duration/count), where for type [`std::chrono::millisecond`](https://en.cppreference.com/w/cpp/chrono/duration) the return type `Rep` is `/*signed integer type of at least 45 bits*/`.

Since we are using two separate types, there will be a `static_cast` here.

----

This does make me wonder if we should really be using a consistent type, all relating to a single source of ticks. Though that's perhaps a separate issue.

DanRStevens (approved):

NAS2D/Renderer/Fade.cpp
cugone: > 
> 
> That had actually come to mind when reviewing the code. That would be acceptable. It's a little bit longer, but clear, and flexible.

It's actually shorter by 5 characters. :P

PR #961, opened 2021-07-30
Fixes Sprite C26820 warning

cugone:
Fixes `Sprite.cpp(224): warning C26820: Assigning by value when a const-reference would suffice, use const auto& instead (p.9).`

See: https://docs.microsoft.com/en-us/cpp/code-quality/c26820?view=msvc-160


DanRStevens (approved):
Ahh, thank you. That was an oversight.

PR #964, opened 2021-07-31
Refactor AnimationSet loading

### LINKED ISSUE #962: AnimationSet C26495 warning
AnimationSet.cpp(58): warning C26495: Variable 'NAS2D::AnimationSet::mFileName' is uninitialized. Always initialize a member variable (type.6).

See: https://docs.microsoft.com/en-us/cpp/code-quality/c26495?view=msvc-160

cugone:
Closes #962

Delegating to a copy constructor is potentially an unsafe operation as well as MSVC complaining that member variables were not initialized. All member variables are now initialized in the header by default.

I took the opportunity to refactor the loading code because of the MSVC warning. The various `process` functions were private to the implementation file via an unnamed namespace as well as only used in one place. Therefore, they made great candidates for Immediately Invoked Initializing Lambdas which aren't accessible outside the function they are declared, reducing their scope even further.

### DIFF
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -50,7 +50,7 @@ bool AnimationSet::Frame::isStopFrame() const
 }
 
 
-AnimationSet::AnimationSet(std::string fileName) :
+AnimationSet::AnimationSet(const std::string& fileName) :
 	mFileName{fileName}
 {
 	try {
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -37,11 +37,6 @@ namespace
 	{
 		return " (Row: " + std::to_string(row) + ")";
 	}
-
-	AnimationSet processXml(std::string filePath, ImageCache& imageCache);
-	std::map<std::string, std::string> processImageSheets(const std::string& basePath, const Xml::XmlElement* element, ImageCache& imageCache);
-	std::map<std::string, std::vector<AnimationSet::Frame>> processActions(const std::map<std::string, std::string>& imageSheetMap, const Xml::XmlElement* element, ImageCache& imageCache);
-	std::vector<AnimationSet::Frame> processFrames(const std::map<std::string, std::string>& imageSheetMap, const Xml::XmlElement* element, ImageCache& imageCache);
 }
 
 
@@ -55,10 +50,36 @@ bool AnimationSet::Frame::isStopFrame() const
 }
 
 
-AnimationSet::AnimationSet(std::string fileName) : AnimationSet{processXml(std::move(fileName), animationImageCache)}
+AnimationSet::AnimationSet(std::string fileName) :
+	mFileName{fileName}
 {
+	try {
+		const auto& filesystem = Utility<Filesystem>::get();
+		const auto basePath = filesystem.parentPath(mFileName);
+
+		Xml::XmlDocument xmlDoc{};
+		xmlDoc.parse(filesystem.read(mFileName).c_str());
+
+		if(xmlDoc.error())
+		{
+			throw std::runtime_error("Sprite file has malformed XML: Row: " + std::to_string(xmlDoc.errorRow()) + " Column: " + std::to_string(xmlDoc.errorCol()) + " : " + xmlDoc.errorDesc());
+		}
+		loadFromXml(*xmlDoc.rootElement());
+	} catch(...) {
+		throw;
+	}
 }
 
+AnimationSet::AnimationSet(const Xml::XmlElement& element)
+{
+	try {
+		loadFromXml(element);
+	}
+	catch(...)
+	{
+		throw;
+	}
+}
 
 AnimationSet::AnimationSet(std::string fileName, std::map<std::string, std::string> imageSheetMap, std::map<std::string, std::vector<Frame>> actions) :
 	mFileName{std::move(fileName)},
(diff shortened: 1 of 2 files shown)


cugone:
@ldicker83 or @DanRStevens The issues Codacy complains about can be safely ignored. I do not have the proper permissions to ignore them:

---
### [Codacy](https://app.codacy.com/manual/lairworks/nas2d-core/commit?cid=570488173) detected an issue:
#### Message: `Function parameter 'fileName' should be passed by const reference.`

    
#### Occurred on:
+ **Commit**:  0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c
+ **File**:    [NAS2D/Resource/AnimationSet.cpp](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp)
+ **LineNum**: [53](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp#L53)
+ **Code**:    `AnimationSet::AnimationSet(std::string fileName) :`

    
#### Currently on:
+ **Commit**:  0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c
+ **File**:    [NAS2D/Resource/AnimationSet.cpp](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp)
+ **LineNum**: [53](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp#L53)

---
### [Codacy](https://app.codacy.com/manual/lairworks/nas2d-core/commit?cid=570488173) detected an issue:
#### Message: `Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20).  `

    
#### Occurred on:
+ **Commit**:  0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c
+ **File**:    [NAS2D/Resource/AnimationSet.cpp](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp)
+ **LineNum**: [61](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp#L61)
+ **Code**:    `xmlDoc.parse(filesystem.read(mFileName).c_str());`

    
#### Currently on:
+ **Commit**:  0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c
+ **File**:    [NAS2D/Resource/AnimationSet.cpp](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp)
+ **LineNum**: [61](https://github.com/lairworks/nas2d-core/blob/0c2ef49c17ec744209c1bd338a2d4b9f6d72c60c/NAS2D/Resource/AnimationSet.cpp#L61)

NAS2D/Resource/AnimationSet.cpp
```diff
@@ -50,7 +50,7 @@ bool AnimationSet::Frame::isStopFrame() const
 }
 
 
-AnimationSet::AnimationSet(std::string fileName) :
+AnimationSet::AnimationSet(const std::string& fileName) :
 	mFileName{fileName}
 {
```
DanRStevens: I think the more proper solution here is to use `std::move`.

```cpp
mFileName{std::move(fileName)}
```

You don't want to be passing by `const&` when using `std::move`. That would end up silently forcing a copy instead.

DanRStevens (changes requested):
Could you provide details or a reference to explain the following:
> Delegating to a copy constructor is potentially an unsafe operation

It might be the MSVC warning is simply a bad warning. It wouldn't be the only case where MSVC warns about things it probably shouldn't. In particular, the use of regular `enum` value names from C imports, such as SDL2, results in a warning that really shouldn't be generated.

----

This change kind of goes against refactoring to remove direct dependence on XML. Here that's been added back by putting an XML include in the `AnimationSet` header file, and incorporating XML types into member methods. Using methods from an unnamed namespace avoids this direct dependence.

----

I find named methods in an unnamed namespace to be more clear than unnamed immediately invoked function expressions. If you were to take scope limitation to the extreme, you might find inlining the code to be a natural conclusion, though that wouldn't necessarily be a great organization that would aid readability.

----

I would like to solve the MSVC warning mentioned above, though I feel this change takes us farther from where we want to be. I would like to back out of this change set.

cugone:
> 
> Could you provide details or a reference to explain the following:
> 
> > Delegating to a copy constructor is potentially an unsafe operation

Any and all documentation, websites, tutorials, etc, that Google-Fu brings up about delegating to the copy-constructor *never* gives an example of using the copy-constructor as the delegating constructor. Nor is it mentioned as a valid or even invalid use.

cugone:
I've created a separate PR just addressing the uninitialized variable issue. See #965

DanRStevens:
> Any and all documentation, websites, tutorials, etc, that Google-Fu brings up about delegating to the copy-constructor never gives an example of using the copy-constructor as the delegating constructor. Nor is it mentioned as a valid or even invalid use.

... so basically what you're saying is there's no documentation to state that it's invalid. :wink:

I'm pretty sure the warning is wrong. Nevertheless, it would be nice to have code that doesn't generate warnings.

PR #965, opened 2021-08-01
Fixes AnimationSet(std::string filename) uninitialized mFileName warning

### LINKED ISSUE #962: AnimationSet C26495 warning

cugone:
Closes #962.

### DIFF
--- a/NAS2D/Resource/AnimationSet.h
+++ b/NAS2D/Resource/AnimationSet.h
@@ -42,9 +42,9 @@ namespace NAS2D
 		const std::vector<Frame>& frames(const std::string& actionName) const;
 
 	private:
-		std::string mFileName{};
-		std::map<std::string, std::string> mImageSheetMap{};
-		std::map<std::string, std::vector<Frame>> mActions{};
+		std::string mFileName{"UNNAMED_ANIMATIONSET"};
+		std::map<std::string, std::string> mImageSheetMap;
+		std::map<std::string, std::vector<Frame>> mActions;
 	};
 
 } // namespace
--- a/NAS2D/Resource/AnimationSet.h
+++ b/NAS2D/Resource/AnimationSet.h
@@ -42,9 +42,9 @@ namespace NAS2D
 		const std::vector<Frame>& frames(const std::string& actionName) const;
 
 	private:
-		std::string mFileName;
-		std::map<std::string, std::string> mImageSheetMap;
-		std::map<std::string, std::vector<Frame>> mActions;
+		std::string mFileName{};
+		std::map<std::string, std::string> mImageSheetMap{};
+		std::map<std::string, std::vector<Frame>> mActions{};
 	};
 
 } // namespace
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -55,8 +55,51 @@ bool AnimationSet::Frame::isStopFrame() const
 }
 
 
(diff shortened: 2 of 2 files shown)


NAS2D/Resource/AnimationSet.h
```diff
@@ -42,9 +42,9 @@ namespace NAS2D
 		const std::vector<Frame>& frames(const std::string& actionName) const;
 
 	private:
-		std::string mFileName;
-		std::map<std::string, std::string> mImageSheetMap;
-		std::map<std::string, std::vector<Frame>> mActions;
+		std::string mFileName{};
+		std::map<std::string, std::string> mImageSheetMap{};
+		std::map<std::string, std::vector<Frame>> mActions{};
```
DanRStevens, line 47: I'm hesitant to set default initial values here. They have a tendency to silence warnings for uninitialized values when you may actually want those warnings. In this particular class, you might reasonably expect the constructor to explicitly set all fields, and should get a warning if a field is not explicitly set.

Somewhat less relevant, though for certain classes "empty" values might not even be allowed. That's not a problem here for the top level collections, though the nested frame list is not allowed to be empty.

DanRStevens (commented):
So this isn't too bad. It avoids changes to the public API would that introduce XML dependencies, so that's a huge plus.

On the downside, it is adding XML code back into the class, but only in a private manner, and only affecting the implementation file.

I feel like I might want to play with this a bit and see if we can get something better.

----

Related, the resource cache code relies on constructors. I had considered looking for ways to extend the resource cache code so it could alternatively use factory functions. Perhaps extend the template and give it a default factory function that delegates to the class constructor. I never put in the time to figure out the syntax though.

If we could update the resource cache code, there would be little reason to maintain a constructor that reads from a file. Particularly not if we want to remove dependencies on file parsing libraries. Instead, that can be moved to a factory function, such as already exists, and have the resource cache code use the factory function. That would enable us to lift the factory function out of the library into the executable project, along with dependencies on specific file processing libraries.

NAS2D/Resource/AnimationSet.h
cugone: The constructors do initialize the members. If not explicitly modified via the initializer list, the default value is used. In the specific case of the offending constructor, the `mImageSheetMap` and `mActions` are initialized in the body of the constructor at the bottom so they are "initialed to defaults" only for a brief period prior to the body being entered.

DanRStevens: I feel like you're missing the point here.

If you comment out one of those constructor lines that properly initializes one of those variables, you will no longer get a warning, since it's now default initialized.

Adding default initializers inhibits the compiler's ability to check for errors. It shouldn't be done unless the defaults are real meaningful defaults, and there are code paths that would make use of them.

DanRStevens:
I don't know if I'm quite ready to review this one yet. This one ties in with issue #972.

DanRStevens:
Closing this to go with a hybrid solution in #1003.

Both solutions were similar, in that initialization was moved to the constructor body, rather than attempting to use the constructor initialization list. That was the key factoring in removing the warning. The other solution keeps most processing in functions in unnamed namespaces, whereas the solution here moved some XML code into the constructor, while leaving the corresponding function from the unnamed namespace intact but uncalled.

PR #971, opened 2021-08-02
Documentation pass on Resource folder

cugone:
See #861 and #899

### DIFF
--- a/NAS2D/Resource/AnimationSet.cpp
+++ b/NAS2D/Resource/AnimationSet.cpp
@@ -32,7 +32,6 @@ namespace
 	ImageCache animationImageCache;
 
 
-	// Adds a row tag to the end of messages.
 	std::string endTag(int row)
 	{
 		return " (Row: " + std::to_string(row) + ")";
@@ -88,11 +87,7 @@ const std::vector<AnimationSet::Frame>& AnimationSet::frames(const std::string& 
 namespace
 {
 
-	/**
-	 * Parses a Sprite XML Definition File.
-	 *
-	 * \param filePath	File path of the sprite XML definition file.
-	 */
+
 	AnimationSet processXml(std::string filePath, ImageCache& imageCache)
 	{
 		try
@@ -108,14 +103,12 @@ namespace
 				throw std::runtime_error("Sprite file has malformed XML: Row: " + std::to_string(xmlDoc.errorRow()) + " Column: " + std::to_string(xmlDoc.errorCol()) + " : " + xmlDoc.errorDesc());
 			}
 
-			// Find the Sprite node.
 			const auto* xmlRootElement = xmlDoc.firstChildElement("sprite");
 			if (!xmlRootElement)
 			{
 				throw std::runtime_error("Sprite file does not contain required <sprite> tag");
 			}
 
-			// Get the Sprite version.
 			const auto version = xmlRootElement->attribute("version");
 			if (version.empty())
 			{
@@ -126,10 +119,6 @@ namespace
 				throw std::runtime_error("Sprite version mismatch. Expected: " + std::string{SPRITE_VERSION} + " Actual: " + versionString());
 			}
 
-			// Note:
-			// Here instead of going through each element and calling a processing function to handle
-			// it, we just iterate through all nodes to find sprite sheets. This allows us to define
-			// image sheets anywhere in the sprite file.
 			auto imageSheetMap = processImageSheets(basePath, xmlRootElement, imageCache);
 			auto actions = processActions(imageSheetMap, xmlRootElement, imageCache);
 			return {std::move(filePath), std::move(imageSheetMap), std::move(actions)};
@@ -141,14 +130,6 @@ namespace
 	}
 
 
-	/**
-	 * Iterates through all elements of a Sprite XML definition looking
-	 * for 'imagesheet' elements and processes them.
-	 *
-	 * \note	Since 'imagesheet' elements are processed before any other
-	 *			element in a sprite definition, these elements can appear
-	 *			anywhere in a Sprite XML definition.
-	 */
 	std::map<std::string, std::string> processImageSheets(const std::string& basePath, const Xml::XmlElement* element, ImageCache& imageCache)
 	{
 		std::map<std::string, std::string> imageSheetMap;
@@ -183,10 +164,6 @@ namespace
 	}
 
 
-	/**
-	 * Iterates through all elements of a Sprite XML definition looking
-	 * for 'action' elements and processes them.
-	 */
 	std::map<std::string, std::vector<AnimationSet::Frame>> processActions(const std::map<std::string, std::string>& imageSheetMap, const Xml::XmlElement* element, ImageCache& imageCache)
 	{
 		std::map<std::string, std::vector<AnimationSet::Frame>> actions;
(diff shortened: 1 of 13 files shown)


NAS2D/Resource/Font.cpp
```diff
@@ -66,7 +66,7 @@ namespace
  * Instantiate a Font using a TrueType or OpenType font.
  *
  * \param	filePath	Path to a font file.
- * \param	ptSize		Point size of the font. Defaults to 12pt.
+ * \param	ptSize		Point size of the font.
```
DanRStevens, line 69: For example, this looks to be removing an obsolete and now incorrect comment. This should ideally stand out from the other changes and be explained with a commit message.

NAS2D/Resource/Font.cpp
```diff
@@ -268,11 +254,6 @@ namespace
 	}
 
 
-	/**
-	 * Generates a glyph map of all ASCII standard characters from 0 - 255.
-	 *
-	 * Internal function used to generate a glyph texture map from an TTF_Font struct.
-	 */
```
DanRStevens, line 275: The ASCII limitation may not be obvious. We should probably keep that comment.

NAS2D/Resource/AnimationSet.cpp
```diff
@@ -126,10 +119,6 @@ namespace
 				throw std::runtime_error("Sprite version mismatch. Expected: " + std::string{SPRITE_VERSION} + " Actual: " + versionString());
 			}
 
-			// Note:
-			// Here instead of going through each element and calling a processing function to handle
-			// it, we just iterate through all nodes to find sprite sheets. This allows us to define
-			// image sheets anywhere in the sprite file.
```
DanRStevens, line 132: I'm a little on the fence about this comment. I'm not certain file processing code in this project is necessarily always generic enough to handle tags in any order. It may also take a more detailed analysis of the code to verify this is true here.

DanRStevens (changes requested):
You should split this into different commits for the different types of changes. I see changes to variable names, removing low value comments, and removing incorrect/obsolete comments, all in the same commit. These should be done separately so it's easier to sort out what is what.

I didn't look too carefully at all the comments being removed. I'm hesitant to spend much time on that before the commit can be split.

NAS2D/Resource/Font.cpp
cugone: The datatype should probably be changed to `uint8_t` or `unsigned char`.

cugone:
How do I split an already submitted and pushed commit without losing the work already done?

NAS2D/Resource/AnimationSet.cpp
cugone: It seems trying to re-write TinyXML/TinyXML2 instead of just using it as-is is causing more problems than necessary. :/ This is trivial with the original library.

cugone:
The Codacy "errors" can be ignored.

NAS2D/Resource/Font.cpp
DanRStevens: What data type? The limitation comes from the fixed size used for the `std::vector`.

DanRStevens:
Looks like you figured out how to split up the commit.

Can you perhaps rebase to remove the first two commits that cancel each other out?

----

That particular Codacy warning keeps popping up. It's getting annoying. It appears to be a false positive. I suspect it's taking the name of the function and assuming it's a different standard library function. I wonder if renaming that function would get rid of the reported warnings.

I'm hesitant to rename just to work around a false warning. Though in that case of that function, perhaps we could have named it a little better? Maybe `readAll`, as it reads the entire file at once.

NAS2D/Resource/Font.cpp
```diff
@@ -69,7 +69,7 @@ namespace
  * \param	ptSize		Point size of the font. Defaults to 12pt.
  */
 Font::Font(const std::string& filePath, unsigned int ptSize) :
-	mResourceName{filePath + "_" + std::to_string(ptSize) + "pt"},
+	mFilePath{filePath + "_" + std::to_string(ptSize) + "pt"},
 	mFontInfo{load(filePath, ptSize)}
```
DanRStevens, line 73: I believe it was called `mResourceName` because presumably not all resources would be loaded from files. In particular, take a look at the `Image` resource, which can be created from raw pixel data.

All resources use the same naming convention. That stems partly from them once having a shared based class with that field in it. Though it may make sense to keep the same naming convention between resource classes.

Honestly though, I'm tempted to get rid of the field, so maybe it doesn't matter. I guess it's somewhat useful for error messages, or debugging, but not much outside of that.

NAS2D/Resource/Image.cpp
```diff
@@ -39,13 +39,13 @@ namespace
 }
 
 
-Image::Image(const std::string& filePath) :
-	mResourceName{filePath}
+Image::Image(const std::string& filePathOrName) :
+	mFilePathOrName{filePathOrName}
```
DanRStevens, line 43: I'm very hesitant to start putting "Or" in variable names. In this case, I much prefer the more generic term `mResourceName` over `mFilePathOrName`. The fact that name was once a file path actually has very little relevance once the data is loaded. At that point, it's mostly just an identifier to help with debugging.

NAS2D/Resource/Font.cpp
```diff
@@ -99,21 +99,21 @@ Vector<int> Font::glyphCellSize() const
 }
 
 
-Vector<int> Font::size(std::string_view string) const
+Vector<int> Font::size(std::string_view text) const
```
DanRStevens, line 102: Mildly curious why you find `text` is less ambiguous than `string`. I find them equivalent. I suppose this is perhaps due to assumptions about embedded newlines?

The closest I can think of is "Text Controls", for multiline text, though I'm also used to hearing "Text Box" for single line of text. There's not really a clear distinction in my mind here.

NAS2D/Resource/AnimationSet.cpp
DanRStevens: It's not the TinyXML code that's the problem. It's how it was used. Some of the code is written with positional assumptions about the XML tag layouts. Such as assuming the version string is always the first attribute. Or making assumptions about the first or only sub element of an element.

NAS2D/Resource/Font.cpp
cugone: It's a semantics thing: Fonts are applied to text (that doesn't change); strings can be manipulated. Additionally, yes, text implies multi-line more strongly than string.

NAS2D/Resource/Image.cpp
cugone: This is partially a consequence of the Texture/Image problem. Images have paths, textures have names/IDs.

NAS2D/Resource/Font.cpp
ldicker83: It was originally for debugging purposes and I had this idea of being able to 'find' resources/assets from a 'resource manager' type of deal by name. That was .... wow that was a long time ago (approaching 20 years, damn). Anyway, it can be removed especially with the `FontCache` object introduced in OPHD (don't remember if that's built into NAS2D or an OPHD exclusive).

DanRStevens:
I was just looking back at this. Not sure what to do with it. There seems to be some valid changes here. It also seems kind of abandoned, with a number of merge conflicts, and discussions that were just sort of left open.

I created a couple of Issues to track a couple of ideas we may want to pull from this.

At this point it may be easier to rework with a narrower focus. I'm thinking maybe I'll close this.

PR #977, opened 2021-08-30
Adds Vector3 user-defined type

### LINKED ISSUE #974: Add Vector3<T> type
See https://github.com/OutpostUniverse/OPHD/issues/1039

cugone:
Closes #974
Closes https://github.com/OutpostUniverse/OPHD/issues/1039

I was unsure how to add a constructor that takes a Vector2, what with the template stuff going on.

### DIFF
--- a/NAS2D/Renderer/Vector3.h
+++ b/NAS2D/Renderer/Vector3.h
@@ -1,0 +1,207 @@
+// ==================================================================================
+// = NAS2D
+// = Copyright © 2008 - 2020 New Age Software
+// ==================================================================================
+// = NAS2D is distributed under the terms of the zlib license. You are free to copy,
+// = modify and distribute the software under the terms of the zlib license.
+// =
+// = Acknowledgment of your use of NAS2D is appreciated but is not required.
+// ==================================================================================
+
+#pragma once
+
+#include <stdexcept>
+
+
+namespace NAS2D
+{
+	template <typename BaseType>
+	struct Vector3
+	{
+		static inline Vector3 Zero = Vector3(BaseType{}, BaseType{}, BaseType{});
+		static inline Vector3 X_Axis = Vector3(BaseType{1}, BaseType{}, BaseType{});
+		static inline Vector3 Y_Axis = Vector3(BaseType{}, BaseType{1}, BaseType{});
+		static inline Vector3 Z_Axis = Vector3(BaseType{}, BaseType{}, BaseType{1});
+		static inline Vector3 XY_Axis = Vector3(BaseType{1}, BaseType{1}, BaseType{});
+		static inline Vector3 XZ_Axis = Vector3(BaseType{1}, BaseType{}, BaseType{1});
+		static inline Vector3 YZ_Axis = Vector3(BaseType{}, BaseType{1}, BaseType{1});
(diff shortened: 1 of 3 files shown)


cugone:
@DanRStevens  Review request.

DanRStevens:
Could you possibly give more details on the reason for the last commit? ("Add explicit BaseType universal initialization for zero and one"). I haven't been able to reproduce any warnings with floating point `Vector`, which I assume that change was there to fix.

NAS2D/Renderer/Vector3.h
```diff
@@ -204,7 +204,7 @@ namespace NAS2D
 	template <typename BaseType>
 	bool operator>=(Vector3<BaseType> v1, Vector3<BaseType> v2)
 	{
-		return v2 <= v1;
+		return !(v1 < v2);
```
DanRStevens: I don't believe these two checks are equivalent. That may hold for the scalar case, but not when there are multiple component values. With multiple component values, you can get incomparable values. This is a result of the multidimensional comparison forming only a partial order.

Scalar:
`1 <= 2`, true
`!(2 < 1)`, true

Multidimensional:
`{1, 0} <= {0, 1}`, false
`!({0, 1} < {1, 0})`, true

NAS2D/Renderer/Vector3.h
```diff
@@ -0,0 +1,268 @@
...
+	bool operator>(Vector3<BaseType> v1, Vector3<BaseType> v2)
+	{
+		return v2 < v1;
+	}
+
+	template <typename BaseType>
+	bool operator<=(Vector3<BaseType> v1, Vector3<BaseType> v2)
+	{
+		return !(v2 < v1);
+	}
```
DanRStevens, line 249: This comparison also has the same flaw for incomparable values.

DanRStevens (changes requested):
Btw, sorry I never gave a detailed review on this one. I remember a made a few general comments on the associated issue though.

I have a few concerns with these changes.

One concern is that I'm not sure NAS2D is the right place for these changes. If there was an easy way to update `Point` and `Vector` to templatize the dimension, then it would make sense to do it here. Otherwise, it's a little weird to put a 3D version into a project that is 2D focused and doesn't make any direct use of 3D versions of these structs.

Another is the `length` function, as written, is rather specific to floating point, and these structs are quite often going to be used with integer values. Not to mention it forces importing a floating point header even when integer values are being used.

I have mixed feelings about supporting `normal`, `dotProduct` and `crossProduct`, particularly given the concerns above. These do make sense to support in a generic 3D library, though NAS2D isn't really that.

The new operations introduce asymmetry in the API. This would seem to make it impossible to resolve the first concern by templating on the dimension. A 2D Vector would not have certain 3D operations. The above operations aside, there are also simple cases such as `reflectZ`.

It seems a little inappropriate to use `std::floor` in the type conversion function. If you want a specific type of rounding, that should probably be it's own function, rather than overloading the meaning of an existing function (and breaking backward compatibility, not that it's a huge issue here).

The comparison operators have subtle bugs.

----

I think maybe we should stick to a simpler change that's kept in OPHD, and more focused on map coordinates.

Issue #704, opened 2020-07-20
Rectangle packed internal format


DanRStevens:
Currently the `Rectangle` class has 4 fields:
- `BaseType x`
- `BaseType y`
- `BaseType width`
- `BaseType height`

We should consider using the alternative and largely equivalent 2 field packed form:
- `Point<BaseType> position`
- `Vector<BaseType> size`

One of the benefits of the packed form, is it becomes much easier to create a `Rectangle` from `Point` and `Vector` objects. Currently, we offer a factory `Create` method to do this, but the syntax isn't nearly as nice. In particular, the `BaseType` must be explicitly specified to access the namespace where the factory method is defined.

```cpp
const auto rect1 = NAS2D::Rectangle{position, size};
const auto rect2 = NAS2D::Rectangle<int>::Create(position, size);
```

One downside is that it makes accessing `Rectangle` position slightly more verbose. It would end up being `rect.position.x` instead of `rect.x`. Of course this assumes you only want one of the position coordinate pair, as otherwise the difference is `rect.position` versus `rect.startPoint()`. Considering a lot of code has been updated to use vectorized operations using the full pair, accessing an individual element becomes less common. That trend may further continue if layout managers are implemented.

PR #996, opened 2021-11-08
Add Rectangle::offset

DanRStevens:
Add method to offset a `Rectangle` by some `Vector`. This affects the `startPoint()`, but not the `size()`.


DanRStevens:
Tagging issue #704 as somewhat related. I think one of the motivating reasons for that issue was to make it easier to offset/translate rectangles, or to change their sizes.

PR #997, opened 2021-11-08
Renames offset to translate

cugone:
The verb translate better communicates what the function does.

PR #998, opened 2021-11-08
Add Rectangle translation operator overloads

cugone:
Adds `operator+`, `operator-`, `operator+=`, and `operator-=` with Vector arguments to affect the translation of a Rectangle.

Related: #997

### DIFF
--- a/NAS2D/Math/Rectangle.h
+++ b/NAS2D/Math/Rectangle.h
@@ -96,9 +96,9 @@ namespace NAS2D
 			return Create(startPoint() + translation, size());
 		}
 
-		constexpr Rectangle operator-(Vector<BaseType> antiTranslation) const
+		constexpr Rectangle operator-(Vector<BaseType> translation) const
 		{
-			return Create(startPoint() - antiTranslation, size());
+			return Create(startPoint() - translation, size());
 		}
 
 		constexpr Rectangle& operator+=(Vector<BaseType> translation)
@@ -108,10 +108,10 @@ namespace NAS2D
 			return *this;
 		}
 
-		constexpr Rectangle& operator-=(Vector<BaseType> antiTranslation)
+		constexpr Rectangle& operator-=(Vector<BaseType> translation)
 		{
-			x -= antiTranslation.x;
-			y -= antiTranslation.y;
+			x -= translation.x;
+			y -= translation.y;
 			return *this;
 		}
 
--- a/NAS2D/Math/Rectangle.h
+++ b/NAS2D/Math/Rectangle.h
@@ -91,6 +91,30 @@ namespace NAS2D
 			return Create(startPoint() + offsetAmount, size());
 		}
 
+		constexpr Rectangle operator+(Vector<BaseType> translation) const
+		{
+			return Create(startPoint() + translation, size());
+		}
+
+		constexpr Rectangle operator-(Vector<BaseType> antiTranslation) const
+		{
+			return Create(startPoint() - antiTranslation, size());
+		}
+
+		constexpr Rectangle& operator+=(Vector<BaseType> translation)
+		{
+			x += translation.x;
+			y += translation.y;
+			return *this;
+		}
+
+		constexpr Rectangle& operator-=(Vector<BaseType> antiTranslation)
+		{
+			x -= antiTranslation.x;
+			y -= antiTranslation.y;
+			return *this;
+		}
+
 		constexpr Rectangle inset(BaseType amount) const
 		{
 			return {x + amount, y + amount, width - 2 * amount, height - 2 * amount};
--- a/test/Math/Rectangle.test.cpp
+++ b/test/Math/Rectangle.test.cpp
@@ -61,6 +61,34 @@ TEST(Rectangle, translate) {
 	EXPECT_EQ((NAS2D::Rectangle{4, 6, 5, 6}), (NAS2D::Rectangle{3, 4, 5, 6}).translate({1, 2}));
 }
 
+TEST(Rectangle, operatorAddSubtract) {
+	//translate(...) and operator overloads are equivalent.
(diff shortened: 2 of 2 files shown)


NAS2D/Math/Rectangle.h
```diff
@@ -91,6 +91,30 @@ namespace NAS2D
...
 		}
 
+		constexpr Rectangle operator+(Vector<BaseType> translation) const
+		{
+			return Create(startPoint() + translation, size());
+		}
+
+		constexpr Rectangle operator-(Vector<BaseType> antiTranslation) const
+		{
+			return Create(startPoint() - antiTranslation, size());
```
ldicker83, line 101: 'antiTranslation' is somewhat unexpected naming here but I'm not going split hairs over it. Just figured I'd note it.

ldicker83:
So I think once some tests are put in place to verify correct behavior of the operator overloads I'm game to merge changes.

NAS2D/Math/Rectangle.h
DanRStevens: Yeah, that stuck out as odd to me too. It's still a translation, just in the opposite direction. The direction is implied by the sign. Putting it into the name too sounds a bit like a double negative.

DanRStevens:
Tagging #996 as related. Issue #704 is also perhaps a bit related.

NAS2D/Math/Rectangle.h
cugone: It was an attempt to prevent confusion. By naming it `antiTranslation` it makes it clear that the offset will go in the opposite direction, i.e. `A + -B`, a positive number will be negative and a negative number will be positive.

DanRStevens: Naming it `antiTranslation` make it sound like it's already `-B`, so it sounds like you're doing `A - (-B)`, which would be `A + B`.

DanRStevens: I would recommend keeping the same term `translation` across all methods.

test/Math/Rectangle.test.cpp
```diff
@@ -61,6 +61,34 @@ TEST(Rectangle, translate) {
 	EXPECT_EQ((NAS2D::Rectangle{4, 6, 5, 6}), (NAS2D::Rectangle{3, 4, 5, 6}).translate({1, 2}));
 }
 
+TEST(Rectangle, operatorAddSubtract) {
+	//translate(...) and operator overloads are equivalent.
```
DanRStevens, line 65: Comment typically have a space after the `//`:
```cpp
// Translate and operator overloads are equivalent
```

Also, I tend to avoid periods at the end of comments, unless maybe it's a block paragraph.

test/Math/Rectangle.test.cpp
```diff
@@ -61,6 +61,34 @@ TEST(Rectangle, translate) {
...
+	EXPECT_EQ((NAS2D::Rectangle{0, 0, 1, 1}).translate({1, 1}), (NAS2D::Rectangle{0, 0, 1, 1} + Vector<int>{1, 1}));
+	EXPECT_EQ((NAS2D::Rectangle{1, 1, 1, 1}).translate({1, 1}), (NAS2D::Rectangle{1, 1, 1, 1} + Vector<int>{1, 1}));
+	EXPECT_EQ((NAS2D::Rectangle{3, 4, 5, 6}).translate({1, 2}), (NAS2D::Rectangle{3, 4, 5, 6} + Vector<int>{1, 2}));
```
DanRStevens, line 68: Unit tests typically make use of CTAD (Class Template Argument Deduction), so the `<int>` is unnecessary in most cases. It should be removed for consistency.

test/Math/Rectangle.test.cpp
```diff
@@ -61,6 +61,34 @@ TEST(Rectangle, translate) {
...
+
+	//Translate by a negative vector is the same as adding a negative or subtracting a positive.
+	EXPECT_EQ((NAS2D::Rectangle{1, 1, 1, 1}).translate({-1, -1}), (NAS2D::Rectangle{1, 1, 1, 1} + Vector<int>{-1, -1}));
+	EXPECT_EQ((NAS2D::Rectangle{1, 1, 1, 1}).translate({-1, -1}), (NAS2D::Rectangle{1, 1, 1, 1} - Vector<int>{1, 1}));
+	EXPECT_EQ((NAS2D::Rectangle{3, 4, 5, 6}).translate({-1, 2}), (NAS2D::Rectangle{3, 4, 5, 6} + Vector<int>{1, 2}));
+
+	//Compound operators are the same as translating a rectangle then assigning it to itself...for now
```
DanRStevens, line 75: I assume this is a reference to issue #1001, which I'm not in favor of.

test/Math/Rectangle.test.cpp
```diff
@@ -61,6 +61,34 @@ TEST(Rectangle, translate) {
...
+	{
+		auto r1 = NAS2D::Rectangle{1, 1, 1, 1};
+		r1 = r1.translate({1, 1});
+		auto r2 = NAS2D::Rectangle{1, 1, 1, 1};
+		r2 += Vector<int>{1, 1};
+		EXPECT_EQ(r1, r2);
```
DanRStevens, line 81: I would suggest doing the assignment to `r1` as a single step here. I suppose maybe you're illustrating the similarity between the two operations here, though I don't feel the intermediate steps really adds much. Mostly it leaves me wondering why it was written that way, and if I need to watch out for something.

I suppose part of my concern is the unit tests are sometimes used as example code to show how a feature is used, and the assignment to `r1` is rather odd here.

DanRStevens:
This PR was linked to Issue #704, which was just closed. I believe the change from PR #1124, along with the older changes from PRs #996 and #997 pretty well cover the cases that are handled here.

----

A non-mutating translate can be done using:
```cpp
const auto translatedRect = originalRect.translate(vector);
```

A mutating translate can be done using:
```cpp
rectToTranslate.position += vector;
```

----

I'm going to go ahead and close this.

Issue #1015, opened 2021-12-06
Update `Delegate` and `Signal` for lambdas


DanRStevens:
Our `Delegate` code is super old. We should consider updating it.

----

The `Delegate` code has a lot of macro `#if` to support really old compilers. Our code relies on C++17, and perhaps soon, C++20. That means we need a reasonably modern compiler. We could gut many of the `#if` checks for old compiler support from the `Delegate` code.

----

Our custom `Delegate` implementation is from about 2005. Source:
https://www.codeproject.com/Articles/11015/The-Impossibly-Fast-C-Delegates

The current implementation far predates the C++11 standard, which introduced [`std::function`](https://en.cppreference.com/w/cpp/utility/functional/function). It seems `std::function` replaces the need to write a custom `Delegate` class. Maybe we should update to using `std::function` and get rid of the custom `Delegate` class.

----

Our current implementation of `Delegate` and `Signal` only really supports member function pointers. We should add lambda support when calling `Signal::connect` and `Signal::disconnect`.

```cpp
NAS2D::Signal<> signal;
const auto delegate = [&object]() { object.method(); };

signal.connect(delegate);
signal.emit();
signal.disconnect(delegate);
```

It may be possible to inline the lambda. Here's how lambdas would compare to the existing way:
```cpp
signal.connect([&object]() { object.method(); });
signal.connect(&object, &object::method);
```

If we move to `std::function`, it may make sense to use lambdas as the primary means of representing `Delegate`. That might mean using it as the primary storage type internally, or perhaps updating calls to prefer passing lambdas.

DanRStevens:
One area of concern I wanted to mention regarding the use of `std::function`, is there are some reports of performance issues with `std::function`, as compared to using lambas with templated methods. In particular, it adds a virtual function call, and may result in allocation of external memory when storing certain large functions, such as lambdas with a large capture group. These are not likely to be of any concern to our particular use case. In particular, for user input handlers, the tiny amount of overhead is never going to matter. Nevertheless, it might be good to better understand the situation by learning how `std::function` works.

First off, the performance comparison was relating `std::function` to template methods taking a lambda, which could inline the lambda call when instantiating the template method. The current implementation of `Delegate` doesn't do that. It uses pointers internally, and so it's already very similar to the virtual function call dispatch offered by `std::function`. The main concern then is the potential allocation of external memory.

[...]

PR #1016, opened 2021-12-07
Use `std::vector` in `Signal`

DanRStevens:
Reference: #1015

There are two main reasons for this change:
1. To make it easier to use `std::function` internally rather than the custom `Delegate` class
2. Speed (for the typical use case)

Using a `std::set` requires the stored object to be "less than" comparable. The custom `Delegate` class has an `operator<`. That being a bit odd aside, the `std::function` class does not have `operator<`. That means switching to `std::function` would either require a custom `operator<` function, or we could change to a different collection class that doesn't require `operator<` for the stored elements.

In terms of speed, a `std::vector` is about the fastest collection class you can likely get for small amounts of data. It's simple, and caches well, so pretty much all operations will be fast for small collections. Things like `std::set` may have `insert` and `delete` speed advantages for larger collections, due to better asymptotic complexity, but the more complex algorithms and lower cache locality mean it'll be slower for small collections. Most `Signal` objects only have a single listener. That far favours using a simpler collection, such as `std::vector`. Further, the main operation will generally be `emit`, rather than `connect` or `disconnect`. That means collection traversal speed matters more than `insert` and `delete` speed. In that case, the `std::vector` will definitely perform better, due to simpler iteration and better cache locality.

### DIFF
--- a/NAS2D/Signal/SignalSource.h
+++ b/NAS2D/Signal/SignalSource.h
@@ -11,7 +11,7 @@
 #pragma once
 
 #include "Delegate.h"
-#include <set>
+#include <vector>
 #include <algorithm>
 
 
@@ -34,7 +34,7 @@ namespace NAS2D
 			const auto iterator = std::find(delegateList.begin(), delegateList.end(), delegate);
 			if (iterator == delegateList.end())
 			{
-				delegateList.insert(delegate);
+				delegateList.push_back(delegate);
 			}
 		}
 
@@ -49,7 +49,7 @@ namespace NAS2D
 			const auto iterator = std::find(delegateList.begin(), delegateList.end(), delegate);
 			if (iterator != delegateList.end())
 			{
-				delegateList.erase(delegate);
+				delegateList.erase(iterator);
 			}
 		}
 
@@ -60,6 +60,6 @@ namespace NAS2D
 		void disconnect(Y* obj, void (X::*func)(Params...) const) { disconnect(MakeDelegate(obj, func)); }
 
 	protected:
-		std::set<DelegateType> delegateList{};
+		std::vector<DelegateType> delegateList{};
 	};
 } // namespace
--- a/NAS2D/Signal/SignalSource.h
+++ b/NAS2D/Signal/SignalSource.h
@@ -31,18 +31,18 @@ namespace NAS2D
 		void connect(DelegateType delegate) { delegateList.insert(delegate); }
 
 		template <typename X, typename Y>
-		void connect(Y* obj, void (X::*func)(Params...)) { delegateList.insert(MakeDelegate(obj, func)); }
+		void connect(Y* obj, void (X::*func)(Params...)) { connect(MakeDelegate(obj, func)); }
 
 		template <typename X, typename Y>
-		void connect(Y* obj, void (X::*func)(Params...) const) { delegateList.insert(MakeDelegate(obj, func)); }
+		void connect(Y* obj, void (X::*func)(Params...) const) { connect(MakeDelegate(obj, func)); }
 
 		void disconnect(DelegateType delegate) { delegateList.erase(delegate); }
 
 		template <typename X, typename Y>
-		void disconnect(Y* obj, void (X::*func)(Params...)) { delegateList.erase(MakeDelegate(obj, func)); }
+		void disconnect(Y* obj, void (X::*func)(Params...)) { disconnect(MakeDelegate(obj, func)); }
 
 		template <typename X, typename Y>
-		void disconnect(Y* obj, void (X::*func)(Params...) const) { delegateList.erase(MakeDelegate(obj, func)); }
+		void disconnect(Y* obj, void (X::*func)(Params...) const) { disconnect(MakeDelegate(obj, func)); }
 
 	protected:
 		std::set<DelegateType> delegateList{};
--- a/NAS2D/Signal/SignalSource.h
+++ b/NAS2D/Signal/SignalSource.h
@@ -28,7 +28,10 @@ namespace NAS2D
 
 		void clear() { delegateList.clear(); }
 
-		void connect(DelegateType delegate) { delegateList.insert(delegate); }
+		void connect(DelegateType delegate)
+		{
+			delegateList.insert(delegate);
+		}
 
 		template <typename X, typename Y>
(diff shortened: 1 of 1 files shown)


DanRStevens:
As a bit of an aside, I did try some speed testing, just out of curiosity. I played with a few parameters in a very ad hoc manner. Here's some sample test code to compare speed:
```cpp
namespace {
	class MockHandler {
	public:
		MOCK_CONST_METHOD0(MockMethod, void());
		MOCK_CONST_METHOD0(MockMethod2, void());
		MOCK_CONST_METHOD0(MockMethod3, void());
		MOCK_CONST_METHOD0(MockMethod4, void());
		MOCK_CONST_METHOD0(MockMethod5, void());
		MOCK_CONST_METHOD0(MockMethod6, void());
		MOCK_CONST_METHOD0(MockMethod7, void());
		MOCK_CONST_METHOD0(MockMethod8, void());
		MOCK_CONST_METHOD0(MockMethod9, void());
		MOCK_CONST_METHOD0(MockMethod10, void());
	};
}

// ...

TEST(Signal, Timing) {
	MockHandler handler{};
	auto delegate = NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod);

	for (int i = 0; i < 100000; ++i) {
		NAS2D::Signal<> signal;
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod2));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod3));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod4));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod5));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod6));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod7));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod8));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod9));
		// signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod10));
		for (int j = 0; j < 100; ++j) {
			signal.connect(delegate);
			// signal.disconnect(delegate);
		}
	}
}
```

Without extra connected methods (that are unique, since duplicates are ignored), the `std::vector` performs as well as or better than `std::set`. Sometimes `std::set` took about 50% more time. This was true even when iterating the `connect`/`disconnect` calls many times (tested ~ `1` .. `1000`). The main factor in speed differences was the presence of other listeners. Once you get to about 10 connected listeners, it starts to become a little bit faster (~10%) to call `connect`/`disconnect` with an underlying `std::set` collection type.

Doing a similar test with `emit()` shows `std::vector` performs the same as or better than `std::vector`. The `std::set` code would sometimes take about an extra 25% more time to run. I used a plain class for that, since using Google Test mock methods had significant overhead, which would drown out the measurements for the dispatch timing.

```cpp
TEST(Signal, Timing) {
	struct MockHandler {
		void MockMethod() {}
		void MockMethod2() {}
		void MockMethod3() {}
		void MockMethod4() {}
		void MockMethod5() {}
		void MockMethod6() {}
		void MockMethod7() {}
		void MockMethod8() {}
		void MockMethod9() {}
		void MockMethod10() {}
	};
	MockHandler handler{};
	auto delegate = NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod);

	for (int i = 0; i < 100000; ++i) {
		NAS2D::Signal<> signal;
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod2));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod3));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod4));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod5));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod6));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod7));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod8));
		signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod9));
		// signal.connect(NAS2D::MakeDelegate(&handler, &MockHandler::MockMethod10));
		signal.connect(delegate);
		for (int j = 0; j < 100; ++j) {
			signal.emit();
		}
		signal.disconnect(delegate);
	}
}
```

cugone:
Additionally, deletion from, and insertion to, the end of a vector is very fast via the `.pop_back()` and `.push_back()` member functions. Leveraging this fact, if order doesn't matter in the collection, `delete`tion of an element can be very fast via the `swap-and-pop` idiom: 

1. Swap the element to be deleted with the last element in the vector.
2. Call `.pop_back()`.

This makes insertion, traversal, *and* deletion using a `vector` very fast, sometimes faster than both the ordered and unordered associative containers (`map`, `unordered_map`, `set`, `unordered_set`) due to cache locality and the above.

DanRStevens:
That's a valid point. I don't believe insertion order was preserved by `std::set`, so technically we've never made any guarantees about the order of `connect` calls matching the order of callbacks during an `emit`.

We could avoid re-seating costs by overwriting the deleted element with the last entry in the `std::vector`, and then popping to reduce size. We wouldn't necessarily need to `swap` per se, as there's no need to copy/move the item being deleted.

----

Of course, the change here wasn't originally to optimize the `Signal` code, but rather to help replace `Delegate` with `std::function`. That effort has kind of failed since `std::function` is not equality comparable (and neither are lambdas), so can't replace `Delegate` with the current design of `Signal`. In particular, the `std::find` calls in `connect` and `disconnect` require equality comparable.

Of course, if we don't check for duplicates in `connect`, and don't offer `disconnect` other than through destruction of the `Signal` object, that problem kind of goes away. Most connect/disconnect calls are related to the way UI event handling is currently done. If we implement the event bubbling idea, we could potentially remove the vast bulk of `connect` and `disconnect` calls (and maybe even most uses of `Signal`). If we get to that point, it may be worth revisiting, and seeing if we still need `Signal`, or if the `disconnect` function can be removed and rely solely on the destructor to disconnect listeners. Might not even need a `connect` method either, if all connections are set in a constructor call. But then, that's getting to be more like `SignalConnection`.

PR #1021, opened 2022-02-07
Ignore zero width glyph

ldicker83:
Fixes an issue on Arch linux when rendering the only other zero-width glyph, ASCII character 173 in the extended set. Not thrilled with special case code but there are only two characters in the table that behave like this. 160 is another one that doesn't have any appearance but it's not zero width so... eh.

### DIFF
--- a/NAS2D/NAS2D.vcxproj
+++ b/NAS2D/NAS2D.vcxproj
@@ -208,6 +208,12 @@
     <ClCompile Include="Mixer\Mixer.cpp" />
     <ClCompile Include="Mixer\MixerSDL.cpp" />
     <ClCompile Include="Mixer\MixerNull.cpp" />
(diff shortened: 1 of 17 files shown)


DanRStevens:
Maybe we should put code to detect the character width, rather than special casing based on the code index.

Looks like you accidentally based the change on a branch I was working on. Normally I would fix that with a `git rebase`. Though if it's just one commit you can use `git cherry-pick` instead. Or if it needs rework, just create a new branch off of master, or use `git reset` to undo the commits for the current branch. For example, with the branch currently checked out, you can run `git reset --hard master` to undo all the commits on that branch, resetting it to look like `master`.

ldicker83:
I think I'll just create a new branch since the change would be super simple versus using special case code. Granted it's really only for the ASCII code page and characters 0 and 173, but still. Who knows how it'll behave on other configurations, etc.

Issue #973, opened 2021-08-04
Use `std::filesystem::path` instead of `std::string` for filepath arguments


cugone:
Various Resource classes and other files use `std::string` to represent file paths. These should be changed to `std::filesystem::path` types which can accept plain names as well, particularly if stored in a `std::map` that accepts `std::string` as the key.

Path corrections like canonicalizing or converting to absolute versions, and converting to the preferred slash type to make the keys consistent for searching may be required before storing.

ldicker83:
Generally speaking I agree with this but I want to be sure this will work on macOS builds without breaking. Apple has an annoying habit of lagging behind adopting updated standards.

DanRStevens:
Related, we'll probably want to use [`std::filesystem::path::extension`](https://en.cppreference.com/w/cpp/filesystem/path/extension) and such internally to break apart components of paths (think `extension` and `parentPath`). Or perhaps just always return strings as `path` objects, and let the caller call standard library methods to extract those values.

DanRStevens:
I was looking back at this issue. I noted one small compatibility issue, in that appending path components will need to use `operator /` rather than `operator +`. Example:
```cpp
filesystem.mountSoftFail(filesystem.basePath() / "data");
```

That compatibility issue is only likely to affect maybe one or so lines of projects that use the library.

----

The argument type change would require moving the `<filesystem>` include from `Filesystem.cpp` file to the header file. That means extra transitive includes. It might be wise to gain some timings for compile times before and after the change. Some standard library headers can result in a lot of (transitively) included code, which can negatively impact compile times. Based on `-ftime-report`, we spend almost 80% of the compile time on source code parsing, and a large part of that is transitive header includes. I'm guessing the effect of this particular library will be fairly minor though. Besides, it doesn't look like `Filesystem` is widely included, and it's always included from cpp files.

I suppose the other side of moving the include to the header is we effectively leak more details about the implementation. I think that's really what's on my mind, though is probably a pretty minor point.

----

We may want to define which types should use `std::filesystem::path`. For raw filesystem paths, we probably want to use `std::filesystem::path`, such as for the `mount` calls. For virtual filesystem paths, such as `readFile` and `writeFile`, maybe, maybe not. We are not strictly required to use paths in the same format within the virtual filesystem, though the way we implement things, they're just appended to real filesystem base paths, so will have the same format. This is maybe getting back to that point about leaking implementation details.

[...]

Issue #967, opened 2021-08-02
Refactor FileSystem internals to use std::filesystem


cugone:
As a stepping stone to moving away from complex string manipulations and towards type-safe and proper error checking with the std::filesystem library, we should refactor the internal and private implementations of `FileSystem` to use `std::filesystem::path` and the manipulation methods therein.

For now, leaving the public API as `std::string` will work as they can just be converted to paths immediately after being passed in and converted back via `std::filesystem::path::string()`.

Downstream code can prepare for the eventual API change by using `std::filesystem::path` variables and passing in a copy of `std::filesystem::path::string()`.

DanRStevens:
I approve of the idea to take a two step approach to the changes.

I've seen some code mix strings and paths, though I can't remember if it was for auto converting or simply for comparison. If `std::string` auto converts to `std::filesystem::path`, then it should be fairly safe to update the API without breaking source compatibility. Binary compatibility is a different matter, though we use static linking with this library, so I'm less concerned about that.

Be aware that functions as currently defined may not perfectly match functionality provided by `std::filesystem::path`. In particular, empty directory components may be inconsistent. Current code might refer to the current working directory as `""`, while `std::filesystem::path` might use `"./"`. This discrepancy may be caught by unit tests.

ldicker83:
Still need to ensure that building on macOS doesn't break using std::filesystem. I'm entirely in favor of this, just super annoyed that Apple lags way behind in adopting new standards even though this standard is over 4 years old now.

PR #1022, opened 2022-07-01
Use SDL for `prefPath`

### LINKED ISSUE #212: Determine special folders without using PhysFS

DanRStevens:
Use SDL for `basePath` and `prefPath`.

This changes the directory layout slightly on Linux. PhysFS ignored the `organizationName` parameter when forming the `prefPath`, while the SDL functions include it. People may need to move config files and saved games for them to be found at the updated path.

Closes #212
Related #967, #973

### DIFF
--- a/NAS2D/Filesystem.cpp
+++ b/NAS2D/Filesystem.cpp
@@ -12,6 +12,7 @@
 
 #include <physfs.h>
 
+#include <filesystem>
 #include <limits>
 #include <stdexcept>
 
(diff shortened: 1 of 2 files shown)


DanRStevens:
I'm going to merge this, though we'll probably want to make a release note about the changed paths for config settings and saved games when OPHD is updated.

Issue #999, opened 2021-11-09
Update Visual Studio solution and projects to Visual Studio 2022


cugone:
This will also require updating the Windows SDK targets. I deliberately did not install the WIndows 11 SDK (10.0.22000.0) because my current computer does not support it (no TPM 2.0, yet) and only installed the latest Windows 10 SDK (10.0.20348.0) along-side version 10.0.19041.0 (this lower version is required for some of the modules I use.)

cugone:
Turns out both AppVeyor CI and Unreal Engine 5.0 (Preview) don't yet support Visual Studio 2022 (I expect UE5 to by its full release) so I'm in the process of *re*-installing 2019 then 2022 (I'm paranoid that something will break, so side-by-side installs must be done in order.)

DanRStevens:
Reference: https://github.com/OutpostUniverse/OPHD/pull/1184

Looks like GitHub Actions can also do Windows builds, and there is a environment available for a preview of VS2022. This may be an alternative to using AppVeyor. Might even be better too, since it offers parallel builds on the free plan.

DanRStevens:
I tried rebasing cugone's work into a `updateToVS2022` branch. It produced peculiar build errors that seemed to indicate a problem within the standard library while building the `test-graphics` project. I'm not really sure what to do about it. The error message doesn't give a source line in the project file.

A preview of the changes and link to the build error can be found at:
https://github.com/lairworks/nas2d-core/compare/main...updateToVS2022

Example error message:
https://github.com/lairworks/nas2d-core/runs/7714848913?check_suite_focus=true
> C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.32.31326\include\xmemory(56,20): error C2220: the following warning is treated as an error (compiling source file TestGraphics.cpp) [D:\a\nas2d-core\nas2d-core\test-graphics\test-graphics.vcxproj]
         C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.32.31[326](https://github.com/lairworks/nas2d-core/runs/7714848913?check_suite_focus=true#step:6:327)\include\xmemory(56,20): error C4189: '_Overflow_is_possible': local variable is initialized but not referenced (compiling source file TestGraphics.cpp) [D:\a\nas2d-core\nas2d-core\test-graphics\test-graphics.vcxproj]

cugone:
[...]

PR #1025, opened 2022-08-06
Update AppVeyor image to VS2022

DanRStevens:
Related: Issue #999

The GitHub Actions workflow for the Windows build has been failing recently due to a `vcpkg` issue. Removing the `dynamic-load` feature for the `SDL2-mixer` package fixes the build with the newer version of `vcpkg` on GitHub Actions, but breaks the build for the older version used on AppVeyor. By updating both build systems to `Visual Studio 2022`, we can remove the `dynamic-load` flag for `SDL2-mixer` and have both build systems work.

Example of problem:
https://github.com/lairworks/nas2d-core/runs/7703423429?check_suite_focus=true
> Package sdl2-mixer does not have a dynamic-load feature
note: updating vcpkg by rerunning bootstrap-vcpkg may resolve this failure.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-image: Visual Studio 2019
+image: Visual Studio 2022
 configuration:
   - Release
 platform:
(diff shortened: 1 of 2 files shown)


DanRStevens:
I'm merging this since it fixes the recent CI build failures. It should probably be checked by Windows developers though. In particular, Windows developers may need to update `vcpkg` after this change.

PR #1026, opened 2022-08-06
Rename `Filesystem` methods to `readFile` and `writeFile`

DanRStevens:
Partly this is to emphasize these methods do full file read and writes. Partly this is to avoid false errors with Codacy, which seems to be confusing the custom `read` method with a standard library method that is subject to unsafe usage.

Example of Codacy errors:
https://app.codacy.com/gh/lairworks/nas2d-core/issues?categoryType=Security&bid=29267938
> Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20).

I suspect the error might actually apply to a standard library [`read`](https://en.cppreference.com/w/cpp/io/basic_istream/read) method.

### DIFF
--- a/NAS2D/Configuration.cpp
+++ b/NAS2D/Configuration.cpp
@@ -87,7 +87,7 @@ void Configuration::save(const std::string& filePath) const
 {
 	try
 	{
-		Utility<Filesystem>::get().write(filePath, saveData());
+		Utility<Filesystem>::get().writeFile(filePath, saveData());
 	}
 	catch (const std::runtime_error& e)
 	{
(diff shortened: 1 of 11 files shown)


DanRStevens:
As silly as that change was, Codacy is now showing 10 issues fixed:
https://app.codacy.com/gh/lairworks/nas2d-core/pullRequest?prid=9933749

ldicker83:
Was a little unsure at first because I tend to like terse naming schemes but 1) it fixes issues and really it's more clear what readFile and writeFile do vs simply read/write. So yay. :D

PR #1027, opened 2022-08-06
Re-implement `Filesystem` to not use PhysFS

DanRStevens:
Related issues: #967, #973
Related PRs: #1024

Re-implement `Filesystem` so it no longer depends on PhysFS. Instead, it uses `std::filesystem` and standard library streams internally.

### DIFF
--- a/NAS2D/Filesystem.cpp
+++ b/NAS2D/Filesystem.cpp
@@ -242,8 +242,8 @@ std::vector<std::string> Filesystem::directoryList(const std::string& dir, const
  */
 bool Filesystem::isDirectory(const std::string& path) const
 {
-	PHYSFS_Stat stat;
-	return (PHYSFS_stat(path.c_str(), &stat) != 0) && (stat.filetype == PHYSFS_FILETYPE_DIRECTORY);
+	const auto& filePath = findFirstPath(path, mSearchPaths);
+	return std::filesystem::is_directory(filePath);
 }
 
 
--- a/NAS2D/Filesystem.cpp
+++ b/NAS2D/Filesystem.cpp
@@ -281,9 +281,10 @@ bool Filesystem::exists(const std::string& path) const
  */
 void Filesystem::del(const std::string& filename)
 {
-	if (PHYSFS_delete(filename.c_str()) == 0)
+	const auto& filePath = std::filesystem::path{mWritePath} / filename;
+	if (!std::filesystem::remove(filePath))
 	{
-		throw std::runtime_error("Error deleting file: " + filename + " : " + getLastPhysfsError());
+		throw std::runtime_error("Error deleting file: " + filename + " : " + errorDescription());
 	}
 }
 
--- a/NAS2D/Filesystem.cpp
+++ b/NAS2D/Filesystem.cpp
@@ -41,6 +41,20 @@ namespace {
 	}
 
 
+	std::string findFirstPath(const std::string& path, const std::vector<std::string>& searchPaths)
+	{
+		for (const auto& searchPath : searchPaths)
+		{
+			const auto& filePath = std::filesystem::path{searchPath} / path;
+			if (std::filesystem::exists(filePath))
+			{
+				return filePath.string();
+			}
+		}
+		return {};
+	}
+
+
 	std::string getLastPhysfsError()
 	{
 		return PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode());
--- a/NAS2D/Filesystem.cpp
+++ b/NAS2D/Filesystem.cpp
@@ -67,16 +67,6 @@ namespace {
 	{
 		return PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode());
 	}
-
-
-	bool closeFile(void* file)
-	{
-		if (!file) { return false; }
-
-		if (PHYSFS_close(static_cast<PHYSFS_File*>(file)) != 0) { return true; }
-
-		throw std::runtime_error("Unable to close file handle: " + getLastPhysfsError());
-	}
 }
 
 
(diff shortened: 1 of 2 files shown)


NAS2D/Filesystem.cpp
```diff
@@ -270,7 +270,8 @@ void Filesystem::makeDirectory(const std::string& path)
  */
 bool Filesystem::exists(const std::string& path) const
 {
-	return PHYSFS_exists(path.c_str()) != 0;
+	const auto& filePath = findFirstPath(path, mSearchPaths);
```
cugone, line 225: Use `std::filesystem::exists` here.

```
const auto filePath = std::filesystem::path{path}; //This variable is for readability only and can be omitted.
return std::filesystem::exists(filePath);
```

NAS2D/Filesystem.cpp
```diff
@@ -291,34 +292,35 @@ void Filesystem::del(const std::string& filename)
...
+	std::ifstream file{filePath, std::ios::in | std::ios::binary};
+	if (!file)
 	{
-		closeFile(myFile);
-		throw std::runtime_error("Error determining length of file or file too large: " + filename + " : Length = " + std::to_string(fileLength));
+		throw std::runtime_error("Error opening file: " + filename + " : " + errorDescription());
+	}
+
+	const auto fileSize = std::filesystem::file_size(filePath);
+	if constexpr (std::numeric_limits<decltype(fileSize)>::max() > std::numeric_limits<std::string::size_type>::max())
```
cugone, line 260: Small nitpick, but it's important for cross-platform compatibility. `<windows.h>` defines its own `max` and `min` macros. In order to get around this you must surround all uses of the `numeric_limits` versions with parenthesis in order for the parser to not freak out:

`if constexpr ((std::numeric_limits<decltype(fileSize)>::max)() > (std::numeric_limits<std::string::size_time>::max)())`

NAS2D/Filesystem.cpp
```diff
@@ -291,34 +292,35 @@ void Filesystem::del(const std::string& filename)
...
+	{
+		if (fileSize > std::numeric_limits<std::string::size_type>::max())
```
cugone, line 262: Same as previous:

`if (fileSize > (std::numeric_limits<std::string::size_type>::max)())`

NAS2D/Filesystem.cpp
```diff
@@ -291,34 +292,35 @@ void Filesystem::del(const std::string& filename)
 
 std::string Filesystem::readFile(const std::string& filename) const
 {
-	PHYSFS_file* myFile = PHYSFS_openRead(filename.c_str());
-	if (!myFile)
+	const auto& filePath = findFirstPath(filename, mSearchPaths);
+	if (filePath.empty())
```
cugone, line 248: Use `std::filesystem::exists` here.

```
const auto filePath = std::filesystem::path{findFirstPath(filename, mSearchPaths)}; 
if(!std::filesystem::exists(filePath)) //File doesn't exist
{
//...throw
}
//Guaranteed to exist.

File: NAS2D/Filesystem.cpp (hunk 1 above)
Comment (cugone) [reply NAS2D/Filesystem.cpp:225]: Since `fileFirstPath` returns a string it should probably be converted to a path first:

```
const auto filePath = std::filesystem::path{findFirstPath(path, mSearchPaths)}; //This variable is for readability only and can be omitted.
return std::filesystem::exists(filePath);
```

File: NAS2D/Filesystem.cpp [hunk 5]
```diff
@@ -77,25 +70,15 @@ enum MountPosition
 };
 
 
-Filesystem::Filesystem(const std::string& argv_0, const std::string& appName, const std::string& organizationName) :
+Filesystem::Filesystem(const std::string& /*argv_0*/, const std::string& appName, const std::string& organizationName) :
```
cugone, line 66: You now have a useless argument *at the front* of the argument list. Recommend flat out removing it.

`Filesystem(const std::string& appName, const std::string& organizationName)`

or, if you are afraid it will break something (why? You are specifically refactoring out old/broken code), use a delegating constructor:

```
Filesystem::Filesystem(const std::string& appName, const std::string& organizationName) : 
mBasePathSdlString{sdlString{SDL_GetBasePath()}.get()},
mPrefPath{SdlString{SDL_GetPrefPath(organizationName.c_str(), appName.c_str())}.get()}
{
}

Filesystem::Filesystem(const std::& /*argv_0*/, const std::string& appName, const std::string& organizationName) : 
Filesystem(appName, organizationName)
{
}
```

NAS2D/Filesystem.cpp
DanRStevens: The `findFirstPath` function already uses `std::filesystem::exists` internally. No need to call the same function twice. And the `findFirstPath` function is needed, since you need to check a specified search path, which you don't get when calling `std::filesystem::exists` directly.

NAS2D/Filesystem.cpp
DanRStevens: I'm not sure what you mean. The Windows CI builds didn't show any warnings or errors. When exactly is this a problem?

Also, Filesystem.cpp doesn't include `<windows.h>`.

NAS2D/Filesystem.cpp
DanRStevens: Same as previous. That check is already performed internally as part of `findFirstPath`.

NAS2D/Filesystem.cpp
DanRStevens: Well that's exactly why I left it in. It would break source code compatibility if I removed that argument. I'm sure it will be removed eventually, just as part of a future change set. I wanted to do the initial refactor without changing the external API.

The delegating constructor idea may be useful for a transition period.

NAS2D/Filesystem.cpp
cugone: >
> Edit: I assume you mean, had we included `<windows.h>`, it would have been a problem.

Yes, if `<windows.h>` is ever included *anywhere* (third-party libraries as well). Even with `NO_MINMAX` defined, the problem presents itself.

DanRStevens: I don't think that's an overly large concern in this case. The code is in the implementation file, not the header file, so the include environment is fairly well controlled, and there is really no reason to include `<windows.h>` in `Filesystem.cpp`.

PR #1036, opened 2022-08-13
Skip caching of problem glyphs

DanRStevens:
Glyphs of size 0 fail to create a character surface when pre-rendering characters to a bitmap image cache. We don't actually care about the reason why the pre-rendering fails. We just want to skip over glyphs that can be printed. Effectively we will print an empty glyph in their place, which will likely also have a size of 0.

This is an alternate solution to PR #1021.


ldicker83:
I like this solution a lot better.

PR #1047, opened 2022-08-22
Remove global constructors

DanRStevens:
Reference: #528

Noticed a few easy improvements after running Clang with `-Wglobal-constructors`. There are more detected with that flag, but these were the easy ones.

### DIFF
--- a/NAS2D/Xml/XmlParser.cpp
+++ b/NAS2D/Xml/XmlParser.cpp
@@ -50,23 +50,26 @@
 #	endif
 #endif
 
-const std::vector<std::string> XML_ERROR_TABLE = {
-	"No error",
-	"Unspecified Error",
-	"Error parsing Element.",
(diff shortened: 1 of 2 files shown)


NAS2D/Xml/XmlParser.cpp
```diff
@@ -50,23 +51,26 @@
...
-	"Error parsing Comment.",
-	"Error parsing Declaration.",
-	"Error: Document empty.",
-	"Error: Unexpected EOF found in input stream.",
-	"Error parsing CDATA.",
-	"Error adding XmlDocument to document: XmlDocument can only be at the root.",
-};
+namespace
+{
+	const auto XML_ERROR_TABLE = std::array{
```
cugone, line 56: You can declare this as `constexpr` since it contains only hard-coded strings.

PR #1050, opened 2022-08-23
Add `Sprite` constructor overload taking `animationCache`

DanRStevens:
Add `Sprite` constructor overload taking an explicit `animationCache`. This allows projects to define and use their own `AnimationSet` cache, which they can control the lifetime and maintenance of. This eliminates much of the need for the library to define it's own `defaultAnimationCache`. This change also makes the API a bit more consistent, in that `AnimationSet` has a constructor overload that can accept a custom `imageCache`.

----

The `defaultAnimationCache` is flagged when compiling with the Clang warning flag `-Wglobal-constructors` and `-Wexit-time-destructors`.

Reference: #528

### DIFF
--- a/NAS2D/Resource/AnimationSet.h
+++ b/NAS2D/Resource/AnimationSet.h
@@ -41,7 +41,7 @@ namespace NAS2D
 		using ActionsMap = std::map<std::string, std::vector<Frame>>;
 
 
-		AnimationSet(std::string fileName);
+		explicit AnimationSet(std::string fileName);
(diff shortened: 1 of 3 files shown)


DanRStevens:
Hmm, this idea may be a bit flawed, in that if we want a custom cache for `Sprite` to use to load `AnimationSet` objects, we'll probably also want the `AnimationSet` objects to get a custom cache to load `Image` objects. It would seem the two levels of caching don't compose particularly well.

PR #1056, opened 2022-08-28
Remove unpacked `connect`/`disconnect` overloads from `SignalSource`

DanRStevens:
These overloads are currently unused.

A `DelegateX` object can be preconstructed at the point of call by enclosing the two parameters with braces `{}`, which can then be passed to the remaining overload. That perhaps adds clarity as well, since the `this` pointer and the member function pointer are closely related, so it makes sense to pass them as a compound object.

Reducing the number of overloads makes it easier to add debug code to just a single overload, which should help with occasional event handling issues.

Reference: https://github.com/lairworks/nas2d-core/pull/1055, https://github.com/OutpostUniverse/OPHD/pull/1271, https://github.com/OutpostUniverse/OPHD/issues/1269

### DIFF
--- a/NAS2D/Signal/SignalSource.h
+++ b/NAS2D/Signal/SignalSource.h
@@ -38,12 +38,6 @@ namespace NAS2D
 			}
 		}
 
-		template <typename X, typename Y>
-		void connect(Y* obj, void (X::*func)(Params...)) { connect(MakeDelegate(obj, func)); }
-
-		template <typename X, typename Y>
-		void connect(Y* obj, void (X::*func)(Params...) const) { connect(MakeDelegate(obj, func)); }
-
 		void disconnect(DelegateType delegate)
 		{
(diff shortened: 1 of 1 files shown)


DanRStevens:
Example of possible debug code:
```cpp
void connect(DelegateType delegate, const char* caller = __builtin_FUNCTION())
{
	std::cout << "Call to connect from: " << caller << " to " << this << std::endl;
	// ...
```

Output (from OPHD):
> Call to connect from: StateManager to 0x7e0a80
Call to connect from: initialize to 0x7e09c0
Call to connect from: initialize to 0x7e0a08
Call to connect from: UIContainer to 0x7e0a08
Call to connect from: Window to 0x7e0a20
Call to connect from: Window to 0x7e0a50
Call to connect from: Button to 0x7e0a08
...

Could probably get a bit more detail about the caller with some additional work.

PR #1064, opened 2022-09-01
Retarget to VS2022 and C++20

ldicker83:


DanRStevens:
I just noticed only the `x64` configs were updated to use `stdcpp20`. The `Win32` configs were still targetting `stdcpp17`. And only the main project was updated, not the unit test project, nor the test-graphics project.

No worries. Now that I see the impact of the setting change on the project file, I believe I can manually apply updates.

Issue #682, opened 2020-05-30
Linux Docker environment updates


DanRStevens:
Linux has a number of different distributions, such as **Ubuntu**, **CentOS**, and **Arch Linux**. Each one has it's own package manager for installing dependencies, and it's own set of package names for needed build tools. Over time, those distributions have come out with newer versions, leaving old sets of packages and one-time setup steps to become out of date. These need to be updated, or perhaps removed.

One way to keep things more relevant, and documented, would be to add Docker images to build and test on these other distributions. In particular, the Dockerfile used to setup the environment would provide documentation on how to get setup and started on those distributions. Currently we only have Dockerfiles covering **Ubuntu**. If we are to maintain the current level of support for environment setup, we should add Dockerfiles for the following distributions:
- **CentOS**
- **Arch Linux**

In particular, the CentOS rules currently available work for CentOS 7, but not for CentOS 8. With CentOS 7, the needed SDL2 packages could be found in the EPEL (Extra Packaged for Enterprise Linux) repository. In CentOS 8, some of these have been moved to PowerTools, while others don't seem to be obviously available.

To enable the PowerTools repository on CentOS 8:
```
dnf --assumeyes install 'dnf-command(config-manager)'
yum config-manager --set-enabled PowerTools
```

For packages not found in repositories, a source install can be performed.

Additionally, the version of PhysFS available on CentOS 7 is too old to be usable, so a source install is needed there.

----

In addition to library dependencies, the Dockerfiles would also serve to document the needed build tools, such as `make` and `g++` or `clang++`. As part of that, it would include steps to enable additional repositories for newer compiler versions, or any setup needed.

In particular, to make use of C++17 features, which this project does, the stock compiler for CentOS 7 is not sufficient. Instead, a newer GCC compiler can be installed with:
```
yum install centos-release-scl
yum install devtoolset-8-gcc-c++
scl enable devtoolset-8 -- bash
```

DanRStevens:
It seems the existing Linux Dockerfiles don't work with GitHub Actions. The use of `USER` in the `Dockerfile` creates permission problems as GitHub Actions expects commands within containers to run as `root`.

They also discourage the use of `WORKDIR`.

Documentation:
https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions

DanRStevens:
It seems the build is able to run on GitHub Actions using the updated Docker images, though with the caveat of there being `git` warning messages in the output.

[...]

PR #1071, opened 2022-09-18
Add Arch Linux build using GitHub Actions

DanRStevens:
Reference: #682


DanRStevens:
Once caveat mentioned in the associated issue, is there are `git` warning messages printed to the console when building from within a Docker container on GitHub Actions. These don't cause build failures though, so aren't really noticeable unless someone is looking at the build log.

PR #1087, opened 2023-04-12
Fix box alignment

DanRStevens:
Fix alignment so `drawBox` and `drawBoxFilled` line up as expected.

Avoid drawing zero and negative area boxes.

Related: #1086

### DIFF
--- a/NAS2D/Math/Rectangle.h
+++ b/NAS2D/Math/Rectangle.h
@@ -74,6 +74,11 @@ namespace NAS2D
 			return (width == 0) || (height == 0);
 		}
 
+		constexpr bool empty() const
+		{
+			return (width <= 0) || (height <= 0);
+		}
+
 		void size(NAS2D::Vector<BaseType> newSize)
 		{
 			width = newSize.x;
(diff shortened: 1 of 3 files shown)


cugone:
Small nitpick. A rectangle with a width or height of zero is a line, a rectangle with a width and height of zero is a point; both should still be drawable. A negative width or height can be handled by swapping the offending points which decays to a normal rectangle again.

DanRStevens:
In the current code base, rectangles are generally treated as endpoint exclusive. I wanted the graphics code to match the existing convention. That makes it easier to place rectangles next to each other with no gaps and no overlap. In such a scheme, a rectangle with a dimension of 1 would be a line. A dimension of 0 would effectively have no interior, and so would not have any area to draw. It also means that when a rectangle is drawn, it's size in pixels exactly matches the `size` of the rectangle. There is no plus 1 to the size to draw the bottom and right edges.

I suppose we are being a bit loose with terminology though. We have functions to draw points and lines, which technically, in the true mathematical sense, are not directly drawable as they have no area. Pixels have area, not points.

The negative dimension case may or may not make sense depending on context. I figure if drawing code has a calculation that results in negative dimensions, it's probably because something was out of bounds and needs to be clipped, hence the guard to not draw such rectangles.

PR #1088, opened 2023-04-15
Fix case of `#include <windows.h>`

DanRStevens:
This potentially matters for a Linux build using MingW, since the Linux filesystem is case sensitive. However, currently the MingW build does not trigger the Windows compile path here. Perhaps a sign the defines being checked here are a bit imprecise.

Reference:
https://github.com/OutpostUniverse/OPHD/pull/1305#issuecomment-1509181851

### DIFF
--- a/NAS2D/Renderer/Window.cpp
+++ b/NAS2D/Renderer/Window.cpp
@@ -15,7 +15,7 @@
 #endif
 
 #if defined(WINDOWS) || defined(WIN32)
-	#include <Windows.h>
+	#include <windows.h>
 	#include <SDL2/SDL_syswm.h>
 #endif
 


ldicker83 (commented):
So I did a little reading and we should be using `_WIN32` for our checks. Mingw checks for this flag and it's defined internally by the MSVC compiler suite.

It's probably why it works the way it did when I updated the code for the shell execute in OPHD.

DanRStevens:
That matches what I recently read.

It seems the underscore is for the implementation reserved namespace, which means compilers can define it by themselves, whereas the non-prefixed version would need to be set in source code or project compiler config settings. That seems to be how MSVC handles the two. The `_WIN32` value is defined by the compiler, whereas the `WIN32` value is defined in `<windows.h>`. Obviously this makes it silly to use `WIN32` as a check for including `<windows.h>`.

https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c

Also, as a side note, the last answer mentions that MingW defines `WIN32` automatically, unless you use `-std=c++??`, like we do (`-std=c++20`). I was able to confirm that behaviour by dumping compiler defines.

https://stackoverflow.com/questions/2224334/gcc-dump-preprocessor-defines

- `echo | x86_64-w64-mingw32-g++ -dM -E -x c++ -`
- `echo | x86_64-w64-mingw32-g++ -dM -E -x c++ -std=c++20 -`

PR #1187, opened 2025-01-19
Grammar and const

Brett208:
Correct some minor spelling / grammar mistakes. Also marked a couple of functions as const within the filesystem class.

@DanRStevens in regards to the comment you made on the OPHD PR, VS has a built in spellchecker that makes finding these sort of errors fairly easy.

### DIFF
--- a/NAS2D/EventHandler.cpp
+++ b/NAS2D/EventHandler.cpp
@@ -243,7 +243,7 @@ EventHandler::WindowMouseLeaveEventSource& EventHandler::windowMouseLeave()
 
 
 /**
- * Joystic axis events generate three values:
+ * Joystick axis events generate three values:
  * Device ID, Axis ID and Axis Position.
  *
  * To connect an event handler to this event, call the 'connect()'
@@ -296,7 +296,7 @@ EventHandler::JoystickBallMotionEventSource& EventHandler::joystickBallMotion()
 /**
  * Joystick button events generate two values:
  * Device ID, and Button ID
- * Button events occure when a joystick button is pressed or released.
+ * Button events occur when a joystick button is pressed or released.
  *
  * To connect an event handler to this event, call the 'connect()'
  * function as follows:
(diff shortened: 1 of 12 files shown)


NAS2D/Filesystem.h
```diff
@@ -51,13 +51,13 @@ namespace NAS2D
...
-		void makeDirectory(const std::filesystem::path& path);
+		void makeDirectory(const std::filesystem::path& path) const;
 
 		bool exists(const std::filesystem::path& path) const;
-		void del(const std::filesystem::path& path);
+		void del(const std::filesystem::path& path) const;
 
 		std::string readFile(const std::filesystem::path& filename) const;
-		void writeFile(const std::filesystem::path& filename, const std::string& data, WriteFlags flags = WriteFlags::Overwrite);
+		void writeFile(const std::filesystem::path& filename, const std::string& data, WriteFlags flags = WriteFlags::Overwrite) const;
```
DanRStevens, line 60: I'm a little hesitant to mark these functions as `const`, since they are write operations.

I know they can be marked `const`. Though the const-ness of these functions depends on implementation details. It's a bit like marking something as `const` because of data indirection, where all the writing is through a pointer, so no direct member fields are modified.

If we changed the implementation to a virtual in-memory only filesystem, these functions could not be `const`.

I guess this raises a bit of a question of what we mean by a `const` `Filesystem` object. If it represents a view of a filesystem, as in having a fixed search path, and designated write folder, then yes, none of that is changing. Though it the idea is extended to include the file data too, then it may mean that having a `const` object should prevent writing to the filesystem.

DanRStevens (commented):
The grammar updates look good. I have some concern about the `const` updates.

NAS2D/Filesystem.h
Brett208: yeah, I approached as not changing the direct members of the class Filesystem and did not extend the meaning of the filesystem class to include constness of the physical filesystem being operated on. Or another way to put it, things like the base path and search paths are not changing. I skipped the const commit and did a force push to make more palatable. ~Brett

DanRStevens (approved):

PR #1191, opened 2025-01-31
Refactor `Point` and `Vector` tests

DanRStevens:
Noticed this while working on:
- PR #1189

### DIFF
--- a/NAS2D/Math/Point.h
+++ b/NAS2D/Math/Point.h
@@ -79,13 +79,13 @@ namespace NAS2D
 		template <typename NewBaseType>
 		constexpr operator Point<NewBaseType>() const
 		{
-			return {static_cast<NewBaseType>(x), static_cast<NewBaseType>(y)};
+			return {NewBaseType(x), NewBaseType(y)};
 		}
 
 		template <typename NewBaseType>
 		constexpr Point<NewBaseType> to() const
 		{
-			return static_cast<Point<NewBaseType>>(*this);
+			return Point<NewBaseType>(*this);
 		}
 
 		explicit operator std::string() const
--- a/NAS2D/Math/Rectangle.h
+++ b/NAS2D/Math/Rectangle.h
@@ -108,15 +108,15 @@ namespace NAS2D
 		constexpr operator Rectangle<NewBaseType>() const
 		{
 			return {
-				static_cast<Point<NewBaseType>>(position),
-				static_cast<Vector<NewBaseType>>(size),
+				Point<NewBaseType>(position),
+				Vector<NewBaseType>(size),
 			};
 		}
 
 		template <typename NewBaseType>
 		constexpr Rectangle<NewBaseType> to() const
 		{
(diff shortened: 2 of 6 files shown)


DanRStevens:
In hindsight, removing `static_cast` in a few places where explicit type conversion is being done was possibly a mistake. It turns out that explicit type conversion using function notation, or function-style casts, is equivalent to C-style casts, in some contexts. Notably, when rounded parenthesis `()` are used rather than braces `{}`.
https://en.cppreference.com/w/cpp/language/explicit_cast
> 2-7) Explicit type conversion (functional notation), also called function-style cast.
> ...
> If the function-style cast is of syntax (2), and there is exactly one expression in parentheses, this cast is equivalent to the corresponding C-style cast.

DanRStevens:
So after further reading, I don't think there's much to worry about here.

There are a number of sources online that suggest the function-style cast should be avoided, and as part of that they often mention that equivalence to C-style casts. However, they all fail to mention the restricted nature of that equivalence. They also all fail to provide any concrete examples of how function-style casts can be bad. When pressed, the discussion always turns to C-style casts, and how C-style casts are bad, and there is never any mention of function-style casts. The examples of bad C-style casts never apply to function-style casts.

I think one of the key points is the difference between a _type-id_ and a _simple-type-specifier_. As per the source on [Explicit Type Conversion](https://en.cppreference.com/w/cpp/language/explicit_cast), the first form of conversions is the C-style cast, which takes a _type-id_, and allows for far greater conversions than a _simple-type-specifier_. In particular, using a _type-id_ allows for pointer and reference conversions, as well as stripping `const` from existing objects. The later forms of conversions are the function-style casts, which use a _simple-type-specifier_. It does not appear to be possible to use pointer or reference syntax with a _simple-type-specifier_.

Further, with function-style casts, the intention is to create a new object of some type. Hence there is no ability to create a modifier reference to an existing object. In particular, there is no stripping of `const` for an existing object, nor is there any `reinterpret_cast` manipulations of a pointer or reference to an existing object. In all cases, a new object is being created. That new object may have `const` specified or not, but that's for a new object, not a modification to an existing object. The restricted nature of function-style casts makes them roughly equivalent to `static_cast` being used for type conversions. Those are generally the safe well defined casts.

The one case where `static_cast` can cause issues is when downcasting to a derived type from a base class pointer which doesn't point to a derived class instance. This is not possible with function-style casts, since _simple-type-specifier_ doesn't appear to allow pointer or reference syntax, and because it's used to create new objects, not modified pointers or references to existing objects.

In summary, the few comments I've seen about avoiding function-style casts appear to be misunderstandings about just how equivalent they actually are to C-style casts, and fear of C-style casts. Though if someone can come up with a concrete example of how function-style casts can be bad, please let me know.

----

On a related note, there is a small difference between function-style casts using `()` vs `{}`. The form using `{}` does not allow narrowing conversions. For instance, converting a `double` to ` float`, or an `int` to a `char`.

PR #1227, opened 2025-02-17
Fill texture coordinate array when initializing OpenGL

ldicker83:
Fixes an issue where calling Renderer::drawBox() before any other call would segfault with a null pointer dereference. This was caused by the texture coordinate buffer not being set by the function. Since the function doesn't ever use texture coordinates, adding a call in Renderer::initGL() to set a texture array at startup resolves this without having to modify the drawBox() function.

### DIFF
--- a/NAS2D/Renderer/RendererOpenGL.cpp
+++ b/NAS2D/Renderer/RendererOpenGL.cpp
@@ -445,15 +445,14 @@ void RendererOpenGL::drawBox(const Rectangle<float>& rect, Color color)
 		return;
 	}
 
+	setColor(color);
 	glDisable(GL_TEXTURE_2D);
 
-	setColor(color);
-
-	const auto p1 = rect.position +  Vector{0.5, 0.5}; // OpenGL centers pixels between integer values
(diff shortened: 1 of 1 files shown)


NAS2D/Renderer/RendererOpenGL.cpp
```diff
@@ -445,15 +445,14 @@ void RendererOpenGL::drawBox(const Rectangle<float>& rect, Color color)
 		return;
 	}
 
-	glDisable(GL_TEXTURE_2D);
-
 	setColor(color);
+	glDisable(GL_TEXTURE_2D);
```
DanRStevens, line 449: Not sure why this is being moved.

DanRStevens (approved):

NAS2D/Renderer/RendererOpenGL.cpp
ldicker83: Normalize it with the drawBoxFilled function, purely a cosmetic change though I didn't mean to push it in this PR :D

PR #1329, opened 2025-09-20
Add `Duration` operations

DanRStevens:
Add basic operator support to `Duration`, covering comparison, addition, subtraction, scalar multiplication, division, and modulo.

### DIFF
--- a/NAS2D/Duration.h
+++ b/NAS2D/Duration.h
@@ -21,4 +21,8 @@ namespace NAS2D
 
 	inline Duration operator+(Duration duration1, Duration duration2) { return Duration{duration1.milliseconds + duration2.milliseconds}; }
 	inline Duration operator-(Duration duration1, Duration duration2) { return Duration{duration1.milliseconds - duration2.milliseconds}; }
+
+	inline Duration operator*(Duration duration, uint32_t scalar) { return Duration{duration.milliseconds * scalar}; }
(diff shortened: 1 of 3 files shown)


cugone:
I don't understand why you are re-implementing the wheel. `std::chrono` exists.

DanRStevens:
To add some numbers to the above:
- Parsing `Duration.h`: 0.678 ms
- Parsing `Duration.h` with `<chrono>` include: 1,174.904 ms

Using `<chrono>` takes over a second to include and parse, and that cost is paid for every translation unit that sees it. Currently `Duration.h` impacts 32 files in OPHD, so in the single threaded case that potentially adds 30 seconds to the compile time. Granted, we have multi-thread compilation, so it wouldn't be quite that bad in practice, though it's still a fairly noticeable chunk of time.

DanRStevens:
Here's a graphical depiction, showing the include and parse time for `<chrono>` (highlighted):
image

As can be seen, `<chrono>` takes the vast majority of the compile time for that file. When `<chrono>` is included, it includes many other headers from the standard library (some of which may be private details), such as `type_traits`, `limits`, `compare`, `sstream`, `vector`, `stl_algo`, `shared_ptr`, and `chrono_io`. The largest sub-include was for `sstream`. I guess maybe some of the standard types in `<chrono>` have special support for serializing them to C++ streams.

PR #1498, opened 2026-06-15
Default screen size to 16x9

cugone:
Changes the default screen size to 1600x900 instead of 1000x700.

The de-facto standard screen resolutions in 2026 are 16:9 aspect ratios. As such the minimum supported screen size of a monitor should be 1600x900.

### DIFF
--- a/NAS2D/Application.cpp
+++ b/NAS2D/Application.cpp
@@ -37,8 +37,8 @@ namespace
 	NAS2D::Dictionary defaultConfigGraphics()
 	{
 		return NAS2D::Dictionary{{
-			{"screenwidth", 1000},
-			{"screenheight", 700},
+			{"screenwidth", 1600},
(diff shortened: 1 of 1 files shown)


DanRStevens (approved):
I'm going to approve since this update to the default setting in the `Dictionary` instance in `Application.cpp` matches the default field value for `mResolution` in `Window.h`. If we're going to set a default, we should be consistent about it.

There was some discussion about using 1920x1080 as a more commonly available minimum resolution today. Maybe that's still worth considering.

Alternatively, maybe we shouldn't be setting a default resolution in the library. This seems like something for the final application to decide on. It's tempting to remove the default values here, and make resolution a required constructor parameter for the `Window` class. Though that's a whole separate thing I don't want to get into here. That would be a breaking change with downstream impacts.

Issue #1452, opened 2026-04-24
Reduce repetition in MSVC project settings


DanRStevens:
There's a lot of duplication in the MSVC project settings, both within projects for the various configurations, and between projects. Perhaps we could make use of Property Sheets to reduce that repetition:

[Share or reuse Visual Studio project settings](https://learn.microsoft.com/en-us/cpp/build/create-reusable-property-configurations?view=msvc-170)

----

Certain values don't vary between configurations, yet are included only in configuration specific sections. In particular `ConfigurationType`, `PlatformToolset`, and `CharacterSet`:
```xml
<ConfigurationType>Application</ConfigurationType>
<PlatformToolset>v143</PlatformToolset>
<CharacterSet>MultiByte</CharacterSet>
```

Some settings vary only by `Debug` and `Release` configurations, and not by `Platform`, such as `UseDebugLibraries`, and `LinkIncremental`:
```xml
<UseDebugLibraries>true</UseDebugLibraries>
```

Some downstream projects (OPHD) have values that vary only only by variable substitutions which are part of the `Condition` check:
```xml
<ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
  <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
</ImportGroup>
```

The `ClCompile` and `Link` sections contain a lot of repeated entries, and could probably make good use of a global section, or only varying values by `Debug` and `Release` configurations.

----

MSBuild should be able to work just fine with the settings grouped differently.

One concern is how certain changes might impact the Visual Studio IDE. What may need testing is how Visual Studio responds if configuration specific values end up in sections where the `Condition` only checks for `Debug` or `Release` and not `Platform` values. Also it would be good to know what Visual Studio will do to the project file if one of the moved settings is edited, either globally in all configurations, or specifically in just one configuration.

----

Edit: The following link appears to be C# specific. It does not work the same for C++ projects.

There may be better support for such editing in newer versions of Visual Studio, based on:
[Create and edit configurations](https://learn.microsoft.com/en-us/visualstudio/ide/how-to-create-and-edit-configurations?view=visualstudio)
[...]

PR #1501, opened 2026-06-16
Use `Directory.Build.props` to reduce `.vcxproj` repetition

DanRStevens:
CI results look good, though this may require more extensive testing with the Visual Studio IDE for odd effects, particularly with the properties editor.

We may also want to check there are no odd effects from moving the configurations into a common file.

Either way, I kind of want this merged so we have a record of working CI changes that help reduce repetition. If it creates IDE issues that aren't easily fixable we'd probably want to revert it. Or maybe if we just don't like the new organization with the various sections split out into the new `MsvcDefaults/` folder.

Related:
- Issue #1452
  - Comment: https://github.com/lairworks/nas2d-core/issues/1452#issuecomment-4408618071
  - Comment: https://github.com/lairworks/nas2d-core/issues/1452#issuecomment-4566867808

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,5 +2,6 @@
   <PropertyGroup>
     <ForceImportAfterCppDefaultProps>$(MSBuildThisFileDirectory)/MsvcDefaults/PropertyDefaults.props</ForceImportAfterCppDefaultProps>
     <ForceImportAfterCppProps>$(MSBuildThisFileDirectory)/MsvcDefaults/PropertyDerived.props</ForceImportAfterCppProps>
+    <ForceImportAfterCppTargets>$(MSBuildThisFileDirectory)/MsvcDefaults/TargetOverrides.targets</ForceImportAfterCppTargets>
   </PropertyGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,5 @@
 <Project>
+  <Import Project="MsvcDefaults/CommonConfigurations.props" />
   <PropertyGroup>
     <ForceImportAfterCppDefaultProps>$(MSBuildThisFileDirectory)/MsvcDefaults/PropertyDefaults.props</ForceImportAfterCppDefaultProps>
     <ForceImportAfterCppProps>$(MSBuildThisFileDirectory)/MsvcDefaults/PropertyDerived.props</ForceImportAfterCppProps>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
     <ForceImportAfterCppDefaultProps>$(MSBuildThisFileDirectory)/MsvcDefaults/PropertyDefaults.props</ForceImportAfterCppDefaultProps>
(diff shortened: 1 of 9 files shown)


DanRStevens:
A remote desktop session was used to check how Visual Studio property editing behaves with the above project file updates.

----

Modifying any property in the "General Properties" section seems to cause `WindowsTargetPlatformVersion` to be written back to the project file with it's current value of `10.0` from `PropertyDefaults.props`. The value can be removed from the project file by selecting it in the General Properties section and setting it to inherit it's value. The properties editor pops up a brief error box, though the project file is updated as expected, with `WindowsTargetPlatformVersion` removed.

General Properties screen:
image

Error box when removing (setting to "inherit") `WindowsTargetPlatformVersion`:
image

----

Modifying other General Properties, such as `ConfigurationType` work as expected for that particular setting. Just with the added oddity of writing `WindowsTargetPlatformVersion` back to the project file as well as the modification.

----

When modifying `WholeProgramOptimization` to `true` from the `Configuration` section with all platforms and configurations selected it added the `WholeProgramOptimization` setting to the `Debug` configurations where the new setting was a non-default value. The `Release` configurations which already have a value of `true` were not modified.

When starting from a clean slate and modifying `WholeProgramOptimization` to `false`, the setting was set for all `Release` configurations where the new setting was a non-default value. The `Debug` configurations were not modified in this case.

When removing the `WholeProgramOptimization` customization and saving, the now empty configuration section was shortened to an automatically closed tag, rather than an open/close tag pair:
```xml
<PropertyGroup Condition="..." Label="Configuration" />
```

----

For the `ClCompile` properties section, the `MultiProcessorCompilation` setting was set to `false`, which updated all configuration blocks. When removing the customization by setting the property to inherit, it truncated the added `ClCompile` section to an automatically closed tag:
```xml
<ClCompile />
```

Notably, it did not remove the empty `ClCompile` tag, and it did not shorten the `ItemDefinitionGroup` tag to an automatically closed tag.

----

When modifying `EnableMicrosoftCodeAnalysis` for a non-existent section, a block of sections was added for every configuration immediately after the `Import` for `Microsoft.Cpp.props`. This matches the location where the `IntDir` and `OutDir` settings used to be.

When modifying `IntDir` and `OutDir` from a clean slate, for the same non-existent section, a block of sections was added for every configuration immediately after the `Import` for `Microsoft.Cpp.props`. This was expected. Additionally, the `WindowsTargetPlatformVersion` value was added back to the `Globals` section of the project file, which matches the description above about modifying General Properties.

----

When modifying `Link` settings such as `OptimizeReferences`, it is written to the configuration blocks where the new value is non-default.

When removing the customization by setting it to inherit, the setting is removed from the project file, as expected.

----

I tried looking for `CAExcludePath` in the properties editor, though it doesn't seem to be editable from there. It seems it may be an earlier undocumented way of handling Code Analysis path exclusions. This might warrant further investigation.

----

Given the above findings, it might make sense to move `WindowsTargetPlatformVersion` back to the individual project files. It might also make sense to use automatically closed tags for the `PropertyGroup` tags with `Label` of `Configuration`.

